PR #1, opened 2023-04-19
Add prettier, husky, format everything

humphd:
This adds prettier and husky to automatically format the code.

I would have added VSCode integration too, but you seem to have that disabled in gititnore.  Let me know if you want it, it's easy to do so this works on save.


humphd:
Done

tarasglek:
Thank you!

PR #4, opened 2023-04-20
Add Chakra-UI 

humphd:
This adds [Chakra-UI](https://chakra-ui.com/) for component and styling, to better support experiments with the UI, make things responsive, accessible, etc.

A quick tour of how you use Chakra-UI.  You get a set of [components](https://chakra-ui.com/docs/components), [UI hooks](https://chakra-ui.com/docs/hooks/use-boolean), and [style props](https://chakra-ui.com/docs/styled-system/style-props).

## Style Props

I'll start with the latter.  When you want to style things, you can add props directly to components, for example to increase the margin:

```tsx
<Box marginTop={2} mb={2} />
```

This uses the expanded form `marginTop` and also the shorter `mb`.  There are tons of styles you can apply this way, and you won't need to include `.css` files as a result.  If you need to style non-Chakra components, you can do it [like so](https://chakra-ui.com/docs/styled-system/chakra-factory):

```tsx
import { chakra } from '@chakra-ui/react'

<chakra.button
  px='3'
  py='2'
  bg='green.200'
  rounded='md'
  _hover={{ bg: 'green.300' }}
>
  Click me
</chakra.button>
```

Here a regular `<button>` uses Chakra style props by being created via the `chakra.button` factory method.  Every HTML element is available this way.

## Components

You can do pretty much anything you want using the built-in [components](https://chakra-ui.com/docs/components).  These components are both visual, but also functional.  For example, you can use [`<Flex />`](https://chakra-ui.com/docs/components/flex), [`<Grid />`](https://chakra-ui.com/docs/components/grid), [`<Stack />`](https://chakra-ui.com/docs/components/stack), etc.

Let's say I wanted to rebuild what you have for the input at the bottom.  I might do this:

```tsx
import { Flex, Input, Button } from '@chakra-ui/react';

<Flex>
  <Input placeholder="Type your question..." flex="1" />
  <Button colorScheme='blue'>Send</Button>
</Flex>
```

There are components for almost anything you need, including layout.  The site https://chakra-templates.dev/ has some nice examples of common layouts, components, etc.  If you like video, there's a lot of great short examples of how to build things on https://www.youtube.com/channel/UC4TmDovH46TB4S0SM0Y4CIg.

## Hooks

There are various hooks to help with common tasks in the UI.  Here are a few examples:

```tsx
import { useDisclosure } from '@chakra-ui/react'

function Dialog() {
  // Keep track of whether some modal is open or closed
  const { isOpen, onOpen, onClose } = useDisclosure()

  return (
    <>
      <Button onClick={onOpen}>Open</Button>
      ...
```

In the above, you have all the state handling you need for dealing with something be open or closed.

Another common example is needing to work with text on the clipboard:

```tsx
import { useClipboard } from '@chakra-ui/react'

function Example() {
  const { onCopy, value, setValue, hasCopied } = useClipboard("");

  return (
    <>
        <Input
          value={value}
          onChange={(e) => {
            setValue(e.target.value);
          }}
        />
        <Button onClick={onCopy}>{hasCopied ? "Copied!" : "Copy"}</Button>
```

I'll start reworking the UI to use Chakra in follow-ups, but wanted to land the deps to make it work, and show you briefly what it is.  Once you get used to using something like this, everything goes way faster (and looks much better!).

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,10 @@
     "prettier": "prettier --write ."
   },
   "dependencies": {
+    "@chakra-ui/react": "^2.5.5",
+    "@emotion/react": "^11.10.6",
+    "@emotion/styled": "^11.10.6",
+    "framer-motion": "^10.12.4",
     "langchain": "^0.0.59",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -1,10 +1,14 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
+import { ChakraProvider } from "@chakra-ui/react";
+
 import App from "./App";
 import "./index.css";
 
 ReactDOM.createRoot(document.getElementsByClassName("chat-container")[0] as HTMLElement).render(
   <React.StrictMode>
-    <App />
+    <ChakraProvider>
+      <App />
+    </ChakraProvider>
   </React.StrictMode>
 );
--- a/yarn.lock
+++ b/yarn.lock
@@ -18,7 +18,7 @@
     "@fortaine/fetch-event-source" "^3.0.6"
     cross-fetch "^3.1.5"
 
-"@babel/code-frame@^7.18.6", "@babel/code-frame@^7.21.4":
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.18.6", "@babel/code-frame@^7.21.4":
   version "7.21.4"
   resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.21.4.tgz"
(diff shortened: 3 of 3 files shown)


tarasglek:
feel free to land stuff directly. I think pull reqs are only needed for informational stuff like this.

My only concern re chakra is that gpt was able to understand write plain css/html for me. I'm less sure that it can help me layout with chakra

humphd:
> feel free to land stuff directly. I think pull reqs are only needed for informational stuff like this.

Sure.  I might do PRs and then land them, just so it's clear what I'm changing.

> My only concern re chakra is that gpt was able to understand write plain css/html for me. I'm less sure that it can help me layout with chakra

If you can't make stuff work, do it in HTML/CSS and we'll refactor it.  However, I think you'll be fine asking it for components instead, for example:

Screenshot 2023-04-20 at 11 35 08 AM

PR #5, opened 2023-04-20
Rework current layout to use Chakra-UI

humphd:
I've done a first pass at cleaning up the UI.  This change does the following:

- Breaks up `App.tsx` into smaller components
- Rebuilds the prompt form using Chakra-UI, making it responsive and expandable (this was my focus)
- Refactors the current message view into new components (I haven't touched the markup yet)

I think moving things around might have broken the Mermaid rendering, so I'm not going to merge this yet.  I'll rip out the `App.css` when I get this all converted, but for now, some of what's in there might be fighting with what I have.  Seems to be ok.

Here's how it looks currently:

### Wide

Screenshot 2023-04-20 at 6 20 00 PM

### Normal

Screenshot 2023-04-20 at 6 21 50 PM

### Expanded

Screenshot 2023-04-20 at 6 22 37 PM

## Mobile 

Screenshot 2023-04-20 at 6 20 20 PM


tarasglek:
can you check how the site renders in dark mode?

humphd:
I'll have to add theme support, I only did it in light mode so far.

I can't figure out why it won't render mermaid stuff with this PR.  It's like it never exists the mermaid fenced block, and continues to include all of the rest of the prompt inside it.

I also tried switching over to use pure-React components of what you're doing:

```tsx
// src/components/MarkdownWithMermaid.tsx
import ReactMarkdown from "react-markdown";
import remarkMermaid from "remark-mermaidjs";

interface MarkdownWithMermaidProps {
  children: string;
}

const MarkdownWithMermaid = ({ children }: MarkdownWithMermaidProps) => {
  console.log({ children });

  return <ReactMarkdown children={children} remarkPlugins={[remarkMermaid]} />;
};

export default MarkdownWithMermaid;
``` 

This should work, but it doesn't end the block and keeps going:

Screenshot 2023-04-20 at 9 49 27 PM

Not sure.  I'll play with it more over the next few days when I take breaks from marking.

tarasglek:
just land it, i'll fix mermaid stuff

humphd:
Added a default theme (using "blue" for the colour scheme for all components, but it's easy to change in `src/theme.ts`) as well as support for both dark and light mode.  I've included a new button in the top-right of the chat panel to allow switching modes (not sure if this is the right location, but it was easy to start with it here).  Changes to the mode are remembered in `localStorage`.

All Chakra-UI components behave in both modes by default, but I'll need to redo the messages to use Chakra for it to work globally.  Here is how they both look:

Screenshot 2023-04-21 at 4 29 14 PM
Screenshot 2023-04-21 at 4 29 06 PM

Regarding the Mermaid bug, I noticed something.  For some reason, my code is wrapping an extra set of 3-backticks around the mermaid fences, and yours didn't:

Screenshot 2023-04-21 at 3 36 50 PM
Screenshot 2023-04-21 at 3 36 24 PM

humphd:
OK, I had a few more minutes, and did styling on the messages too:

Screenshot 2023-04-21 at 8 29 33 PM

Screenshot 2023-04-21 at 8 29 39 PM

I've also added copy-to-clipboard.  I think this is probably done enough at this point to land and iterate.

humphd:
- Added a header, moved dark/light mode there.
- Added user preferences with model + hook
- Fixed a bunch of UI problems on mobile

Screenshot 2023-04-22 at 9 05 27 PM

Screenshot 2023-04-22 at 9 05 06 PM

tarasglek:
This looks amazing. Note if you do branches on this repo, we automatically get https://developers.cloudflare.com/pages/platform/preview-deployments/ which makes it easier to validate

tarasglek:
Couple thoughts:
* I like the dynamic "shift enter" explanation
* I don't like the manual expansion of textarea height over previous dynamic one
* overall UI looks much cleaner now.
* The animation for loading and button layout are very nice
* Now that ui is much cleaner, it feels like single-message mode should probably apply some css effect to fade messages that wont get sent to model
* The options are really hard to discover now. IMHO they should live in a sidebar on the left. Would be cool if they showed in a sidebar that would expand automatically when there is empty space. I don't know how to do that sort of dynamic css ui

humphd:
Thanks for the feedback.  Here's what I've done:

> * I like the dynamic "shift enter" explanation

Cool, I've improved it a bit

> * I don't like the manual expansion of textarea height over previous dynamic one
>   
>   * the expansion button on far right is tricky. I think all input controls should be aligned to be same width as chat history

I didn't realize your old one worked like that.  I've updated it so that it works in one of two ways:

- shift+enter will expand the prompt 1 line at a time until it's half-hight
- toggling the expand/minimize icon does it all in one step
- toggle is lined up with prompt vs. all the way to right

The reason I want it to open-on-expand is so that I can jump to a bigger space right away, but I get why you want it to grow on demand.  Hopefully this works for both cases.

> * overall UI looks much cleaner now.

:)

> * The animation for loading and button layout are very nice
> * Now that ui is much cleaner, it feels like single-message mode should probably apply some css effect to fade messages that wont get sent to model

Done.  Clicking on single-message-mode will toggle the messages above the last one to collapse or expand back into view.  I'm useless at CSS animations, but Chakra makes it easy if you want to play with others: https://chakra-ui.com/docs/components/transitions

> * The options are really hard to discover now. IMHO they should live in a sidebar on the left. Would be cool if they showed in a sidebar that would expand automatically when there is empty space. I don't know how to do that sort of dynamic css ui

What you want is a Drawer, see https://chakra-ui.com/docs/components/drawer

> * I see that mermaid works again!

If it does, that means it's due to some variability in what GPT sends, since I didn't fix anything.

I've also done some error handling, and moved the Message CSS to `src/components/Message.css`.  All GPT text lives in an `.message-text` element, so we can style beneath that.

If you're cool with this, let's merge and iterate on top.  I'll do PRs in this repo next time for the preview deployments (didn't know you had that setup, awesome!)

tarasglek:
Note this pull req broke `yarn build` pushed https://github.com/tarasglek/private.overthinker.dev/commit/b785274a62e04ff92cc30ed6d48b24780692f7a9

humphd:
I think I've filed or fixed everything mentioned above.

PR #7, opened 2023-04-24
API Key startup UX

humphd:
This alters the startup logic to check for an API Key in local storage (using `useLocalStorage()` from [react-use](https://github.com/streamich/react-use/blob/master/docs/useLocalStorage.md)), and if it's missing, we show a splash screen with form to enter.  The replaces the sync `prompt()` browser flow.

I'd like to include a way to delete the API key from storage via the UI, but getting it to work how I wanted was more work than I want to do this morning.

To test, remove your API key from local storage and refresh:

Screenshot 2023-04-24 at 9 32 51 AM

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,27 +11,11 @@ import { useSettings } from "./hooks/use-settings";
 import useMessages from "./hooks/use-messages";
 
 function App() {
-  const { isOpen: isExpanded, onToggle: toggleExpanded } = useDisclosure();
   const { messages, setMessages, removeMessage } = useMessages();
   const [singleMessageMode, setSingleMessageMode] = useState(false);
   const { settings } = useSettings();
+  const { isOpen: isExpanded, onToggle: toggleExpanded } = useDisclosure();
   const [loading, setLoading] = useState(false);
-  const [openai_api_key] = useState(() => {
-    // getting stored value
-    const saved = localStorage.getItem("openai_api_key");
-    if (!saved || saved.length === 0) {
-      // get it from user via input func
-      const key = prompt("Please enter your OpenAI API key");
-      // save it
-      if (key) {
-        localStorage.setItem("openai_api_key", JSON.stringify(key));
-        return key;
-      }
-    } else {
-      return JSON.parse(saved);
-    }
-    return "";
-  });
   const messageListRef = useRef<HTMLDivElement>(null);
   const toast = useToast();
 
@@ -71,7 +55,7 @@ function App() {
     try {
       // Send chat history to API
       const chat = new ChatOpenAI({
-        openAIApiKey: openai_api_key,
+        openAIApiKey: settings.apiKey,
         temperature: 0,
         streaming: true,
         modelName: settings.model,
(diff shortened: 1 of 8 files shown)


tarasglek:
i was thinking we could just ask for it via the chat ui :)

humphd:
Yeah, that would be good too.  I can try that.

tarasglek:
I would rather have less code. Why not just show a modal options dialog here? I would prefer to not maintain 2 dialogs. Lets just make the requirement for key scarier when it's missing.

humphd:
Reworked it to be inline, added the `apiKey` to the `Settings` (you'll need to put your API key in again after this is merged, since localStorage key has changed).

When the user starts the app for first time, chat form is replaced with API key form.  When they save a key, it gets replaced.  They can change/delete it in settings modal:

Screenshot 2023-04-25 at 1 09 09 PM

Screenshot 2023-04-25 at 1 09 31 PM

tarasglek:
can we prioritize this? I think UX for key is main blocker for sharing this project

humphd:
Sure, can you check what I have on this PR?  Are you ok with it, want changes?  We can land if you're satisfied.

humphd:
GitHub has been weird for me today.  I'm talking about https://github.com/tarasglek/chatcraft.org/pull/7#issuecomment-1522141561 in case the UI isn't updating.

tarasglek:
I think this would work. Couple thoughts:
* Not a fan of hiding the key in settings or this input box. I often use these to look up key for when i need to use it elsewhere.
* The text doesn't make it clear that one must enter the key. "Please enter your openai key:" would be more direct.
* Shouldn't display the helpful assistant message. Should instead include some docs on how to get the key there.
* probably should show the template of what the key looks like and validate it to not accept invalid keys

humphd:
Alright, I've reworked it:

> * Not a fan of hiding the key in settings or this input box. I often use these to look up key for when i need to use it elsewhere.

I've added a Show/Hide button inline in the password input, so people can see it.  Personally, I don't want my key being shown, but I get that you do in some cases.  This gives both options.  The preferences modal does the same.

> * The text doesn't make it clear that one must enter the key. "Please enter your openai key:" would be more direct.

Updated

> * Shouldn't display the helpful assistant message. Should instead include some docs on how to get the key there.

I've changed the `useMessages()` hook to dynamically alter the first message we give to either be instructional or the typical greeting, depending on whether there's an api key.

> * probably should show the template of what the key looks like and validate it to not accept invalid keys

I've included an example in the instruction message.  I don't think we should try to validate the key, though, since [OpenAI hasn't committed to a format](https://community.openai.com/t/regex-s-to-validate-api-key-and-org-id-format/44619).

Screenshot 2023-04-26 at 1 31 14 PM

Screenshot 2023-04-26 at 1 31 40 PM

humphd:
@tarasglek test this out a bunch of ways before you merge and set people loose on it.  I *think* I've got all cases covered with this, but there might be edge cases I haven't tried.

tarasglek:
Seems to work for all my testing.

PR #15, opened 2023-04-25
Add better code block rendering

humphd:
This is a WIP, but it's worth checking out for comment.

I'm playing with [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter) and some custom CSS to make code easier to read.

So far, things work, but the Mermaid diagram broke the app until I refreshed (might be smart to load it in an iframe too?)

Here's what it looks like:

Screenshot 2023-04-24 at 11 01 53 PM
Screenshot 2023-04-24 at 11 01 39 PM
Screenshot 2023-04-24 at 11 01 29 PM

I'm not sure which languages you want to include, whether you want syntax highlighting at all, etc.  See what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,13 +20,16 @@
     "react-dom": "^18.2.0",
     "react-icons": "^4.8.0",
     "react-markdown": "^8.0.6",
+    "react-syntax-highlighter": "^15.5.0",
     "react-textarea-autosize": "^8.4.1",
     "react-use": "^17.4.0",
+    "remark-gfm": "^3.0.1",
     "remark-mermaidjs": "^4.1.1"
   },
   "devDependencies": {
     "@types/react": "^18.0.28",
     "@types/react-dom": "^18.0.11",
+    "@types/react-syntax-highlighter": "^15.5.6",
     "@vitejs/plugin-react": "^3.1.0",
     "husky": "^8.0.3",
     "lint-staged": "^13.2.1",
--- a/src/components/MarkdownWithMermaid.tsx
+++ b/src/components/MarkdownWithMermaid.tsx
@@ -1,69 +1,66 @@
 // TODO: switch to using proper React components for this...
-// import ReactMarkdown from "react-markdown";
-// import remarkMermaid from "remark-mermaidjs";
-
-// interface MarkdownWithMermaidProps {
-//   children: string;
-// }
-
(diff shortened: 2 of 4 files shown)


tarasglek:
yeah this was on my mental todo list. Previous iteration of this used same highlighter. I would love to include all languages but I couldn't figure out how to make the library bundle all supported langs.

I'm flexible re mermaid and iframes.

humphd:
I was thinking about this, and the "support all languages" approach means we need to convert this to use dynamic imports when we see a language for the first time.  There's no way to bundle all the language modules statically, it would be too huge.  I'll see if I can make that work.

Another thing that occurs to me: we can pass a component for the parent of a `<code>` or `<pre>` that `<SyntaxHighlighter>` uses, and inside that, we can add various buttons.  For example: copy the code in this particular code block, but also, render this Mermaid/HTML.  Doing it via a button vs. as it streams would stop half-broken renders from crashing the app.  Also, trying to render a half-available Mermaid diagram, which constantly shifts and flickers, isn't great UI anyway.

humphd:
I had the app teach me how to get the code-as-text passed to a header component, so now copy/paste works:

Screenshot 2023-04-25 at 6 26 10 PM

We could easily put other buttons in the code-block headers too.

tarasglek:
wow, you found a working react mermaid renderer!
 I spent a few days searching for that and gave up. This branch is absolutely amazing.

if you ask gpt4 to do "sample mermaid diagram: ```mermaid", the ui will go blank :)

We need to disable rendering previews for wip messages

tarasglek:
I pushed a loading through to message rendering to stops the mermaid previews from crashing also reduces flicker.
I suspect it will also crash when fed invalid mermaid from model. Need a try/catch somewhere.

I think we can also use loading state to enable scrolling while message is coming in and only scrolling to the end when message is finished. This was another user complaint.

I noticed we now replace mermaid source with a diagram which isn't quite right.

tarasglek:
I'm good to land this, soon as you ready. I think current ui is freaking incredible for coding :)

humphd:
Looks good.  One thing I had thought about was adding a "render" button in the code header when the language can be rendered (i.e., make it opt-in), but this is nice too, since it only shows up when it's done.

I notice it is re-rendering on mouse move, which we should fix in a follow-up.

I've rebased this on the API Key stuff and fixed conflicts.  I'll land it.

PR #17, opened 2023-04-26
Allow up-arrow to copy previous message into prompt

humphd:
This adds support for pressing the Up Arrow in the prompt and getting the previous message copied into the textarea (kind of like up-arrow in a shell).  I tried doing multi-up/down to scroll through the other messages, but because we have line breaks in prompts, it becomes impossible to know if an up/down is meant to scroll through the message or scroll to next.

I think this is useful as is for being able to copy what AI wrote and alter it to build the next prompt.


tarasglek:
I love this so much. Good hack to get a lot of functionality unlocked

PR #19, opened 2023-04-27
Expanding prompt form switches to CMD+Enter mode

humphd:
When the user has Shift+Enter mode enabled, expanding the prompt overrides to give CMD+Enter mode.


tarasglek:
is there a point to keeping the other piece of state? Is there a time when you want a non-expanded form and enter-newline mode?

humphd:
I think having this choice is pretty standard, see prior art in Slack:

Screenshot 2023-04-27 at 10 28 13 AM

I'm basically copying that.

PR #21, opened 2023-04-27
Reduce re-renders of markdown in messages

humphd:
I noticed that the HTML/Mermaid live renderings were constantly being re-rendered when I was moving my mouse over/out-of a message.  This is due to the state change on the message with the mouse handler.

Using ChatCraft (ha!) I discovered that I can memoize the whole markdown component, and only re-render it when its direct props change vs. parent props.  This works via [`React.memo()`](https://react.dev/reference/react/memo), which I'd never heard of (I've only ever known about `useMemo()` as a hook).

I've also renamed the component, since it's now doing more than Markdown + Mermaid (i.e., it's a general purpose `Markdown` renderer now).

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -1,4 +1,4 @@
-import type { ReactNode } from "react";
+import { memo, type ReactNode } from "react";
 import {
   Flex,
   ButtonGroup,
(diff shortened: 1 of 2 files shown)


tarasglek:
Man, i wanna have sharing functionality just to see this convo :)

I find the singlefile extension works great for saving these convos in meantime.

PR #22, opened 2023-04-27
PWA support via vite-plugin-pwa

humphd:
Make the app installable as a PWA using [vite-plugin-pwa](https://vite-pwa-org.netlify.app/).

My reading of the docs says that the only decision we really need to make here is whether to have the service worker do updates automatically or require user prompt:

- [Automatic](https://vite-pwa-org.netlify.app/guide/auto-update.html)
- [Prompt](https://vite-pwa-org.netlify.app/guide/prompt-for-update.html)

It's a bit tricky to change from one to the other later, apparently.  I've used automatic updates here, since we already cache values in storage (i.e., we won't lose data).  Read the above and see if you disagree.  We could also add some buttons in header to indicate when things need to be updated.

Using this, you should be able to load the app, set Network to Offline mode, and reload.  Everything should be cached, so future runs will be fast.


humphd:
It's interesting how this works by default, see ...:

Screenshot 2023-04-27 at 12 48 03 PM

It's pre-caching all our chunks, which includes 150+ language JS files for markdown styling.  I think that's the right way to do it, though we could also have it do these on demand by changing the workbox strategy.  see what you think.

Here is is installed as a PWA on desktop, running in its own window!

Screenshot 2023-04-27 at 12 49 46 PM

humphd:
Even if you never do anything with a PWA (don't install) the service worker will cache all the assets, so it helps with speedup of loading.

tarasglek:
I'll ok with this feature if you if it works for you

PR #25, opened 2023-04-28
Refactor to fix re-renders, split messages from streaming message

humphd:
I've done a bunch of clean-up and refactoring to get rid of various unnecessary re-renders.  Specifically, this change does the following:

1. Rename `MessageView` to `MessagesView` and pass in both `messages` as well as the  new streaming message that is being loaded
2. Connected to 1., split apart the `messages` from the current response that is being streamed in
3. Move all the OpenAI code and management of this streaming message to its own hook, which cleans up `App.tsx`
4. Fixed a bunch of bugs where effects, callbacks, etc. were missing dependencies
5. Fixed a bug in the Markdown code blocks with copy/paste, weird height issues on wrapping, etc
6. Added logic to allow for short-form language names in code fences (e.g., `ts` vs. only `typescript`)

Everything seems to work well for me, but see if you spot any problems.  This reduces jank quite a bit for me.


tarasglek:
my approach is to land these and back em out if you see problems

PR #28, opened 2023-04-28
Don't precache all Prism languages in service worker

### LINKED ISSUE #27: All the languages in the world are loaded eagerly
image

humphd:
Fixes #27.

Previously, told the service worker (via Workbox) to precache everything under `build/**/*.js`.  This included all 180 Prism language files (used for syntax highlighting).

This PR changes our Vite PWA build such that we first build include and ignore patterns for all the languages, and explicitly exclude them all, then add back in only the ones defined in  `includedLanguages`.  Currently that means: CSS, JS, TS.  Adding more languages is as simple as adding more to that array.

The entire set of language files is still included in the static build (see `build/assets/*`), so they can be loaded on-demand at runtime.

With this change, the number of files being precached in the service worker's cache goes from 318 to 19.  It's possible we'll want to add more as we use this.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -3,6 +3,38 @@ import react from "@vitejs/plugin-react";
 import wasm from "vite-plugin-wasm";
 import { VitePWA } from "vite-plugin-pwa";
 
+import { readdirSync } from "node:fs";
+import { join } from "node:path";
+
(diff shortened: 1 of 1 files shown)


humphd:
Tested this on the preview URL, and it's working: TS is precached, stuff like Docker and C++ get loaded on demand.

tarasglek:
This is the most magical of all your contributions. I wouldn't have figured this out. Tested it out, works great!

PR #30, opened 2023-04-29
Add ability to count and display token count, with pref

humphd:
Adds the ability to see how many tokens are being sent in a chat (i.e., across all messages).  It's off by default (adds extra download size for token parser), but for people who care about this, it can be enabled via the preferences.

Screenshot 2023-04-29 at 12 28 07 PM

Uses the token parse for the model currently being used, which is nice (thank you langchainjs!)


tarasglek:
That's cool. Should do the math for $$$ :)

humphd:
Done!  It only displays if it's $0.01 or more:

Screenshot 2023-05-01 at 4 23 07 PM

PR #33, opened 2023-05-02
Don't auto focus API Key input element

### LINKED ISSUE #31: Autofocus asking for openai key causes issues with lastpass autofill on android
per https://news.ycombinator.com/item?id=35785533

humphd:
Fixes #31

I don't have Android or LastPass so I can't test this; however, I think this is probably enough.

Is there more you think this needs, or is this good?

### DIFF
--- a/src/components/PromptForm.tsx
+++ b/src/components/PromptForm.tsx
@@ -244,7 +244,7 @@ function PromptForm({
           </Flex>
         </chakra.form>
       ) : (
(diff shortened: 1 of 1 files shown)


tarasglek:
I think this needs autocomplete=off per https://stackoverflow.com/questions/20954944/stop-lastpass-filling-out-a-form

humphd:
Added `autocomplete=off`

PR #38, opened 2023-05-05
Add word wrap to code blocks

### LINKED ISSUE #37: add word-wrap for code
otherwise it's really hard to read:
image

humphd:
Fixes #37 

There are a bunch of bugs in the way `react-syntax-highlighter` does their word wrapping.  I've used what I could, and overridden the rest with manual styles.  I think this is "good enough."  See if you agree:

Screenshot 2023-05-05 at 1 11 07 PM

Screenshot 2023-05-05 at 1 13 00 PM
 
I've tried half-a-dozen different approaches to this, but without controlling more of how the HTML gets generated for each line, it's hard to make it perfect.

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -94,7 +94,6 @@ function Markdown({ includePlugins, previewCode, children }: MarkdownProps) {
                   borderColor: "gray.600",
                 }}
                 pb={1}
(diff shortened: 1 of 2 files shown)


tarasglek:
This seems like a hard problem, this is an improvement. Thank you.

PR #39, opened 2023-05-06
Improve HTML previews, encourage Mermaid diagrams

humphd:
This improves a few things about previews:

1. Adds a note to the system prompt to indicate that we'd like to get Mermaid diagrams when discussing anything visual.  My testing has shown that this causes it to use them more often.
2. Make it possible to open an HTML preview in its own tab, and dress up the UI around it a bit. I've also moved the preview below the code, which makes more sense to me as a reference:

Screenshot 2023-05-06 at 1 17 35 PM


tarasglek:
Wow. That's an incredible improvement. I actually needed to draw a diagram, so used this to test it out.

PR #43, opened 2023-05-08
Support cancel, pause, resume on responses

### LINKED ISSUE #35: Pausing streaming while mouse down
I often want to copy data out as it's being generated.

### LINKED ISSUE #41: option to "cancel" a generation
would be great if there is a way to "cancel" a generation next to the "loading ..." button when i don't need more generations for the current answer

humphd:
Fixes #41
Fixes #35 

This adds support for cancelling, pausing, and resuming streaming responses.  I've included the following:

- escape key cancels + UI button to click and discover
- mouse down while loading pauses, mouse up resumes
- pause/resume toggle button in UI to click and discover

Screenshot 2023-05-07 at 9 22 59 PM

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
     "react/no-children-prop": "off",
     "jsx-a11y/no-autofocus": "off",
     "jsx-a11y/iframe-has-title": "off",
-    "@typescript-eslint/no-explicit-any": "off"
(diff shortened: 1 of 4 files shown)


tarasglek:
This is nice. I think i would prefer loading button to become cancel button instead of adding another element.

PR #45, opened 2023-05-08
Just show me the code

tarasglek:
This changes the system and default prompts to make gpt a lot less verbose when coding. Thanks Willem for the prompt!

image

image


tarasglek:
There is some serious screwage with switching messages.

tarasglek:
I changed code to compare values of messages instead of Object equality as that probably didn't work with serialization

PR #46, opened 2023-05-08
Clean up prompt buttons, add loading/pause hint

humphd:
Fixes for comments in https://github.com/tarasglek/chatcraft.org/pull/43#issuecomment-1537733989.

Reduces the number of buttons we have when loading, and adds some hints about what's going on:

Screenshot 2023-05-08 at 1 36 38 PM
Screenshot 2023-05-08 at 1 36 32 PM

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -122,6 +122,12 @@ function App() {
     }
   }
 
+  // Restart auto-scrolling and resume a paused response when Follow Chat is clicked
+  function handleFollowChatClick() {
+    setShouldAutoScroll(true);
+    resume();
+  }
+
   return (
     <Box w="100%" h="100%">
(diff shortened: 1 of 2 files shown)


humphd:
I've updated this to leave the 'Clear Chat' and 'Send' buttons alone, and moved the `Cancel`, `Pause`/`Resume` buttons up to the hint area:

Screenshot 2023-05-09 at 3 11 47 PM

So you're saying you like the `Send` button NOT being disabled with the form is empty?  I've left it like that, but if I've misunderstood, I can fix.

tarasglek:
I like the idea of moving those buttons, but I think the buttons should live on/near current message being written. 

We have a general ui problem where on mobile devices with kb open we can't see content at all 

image

can address that in followup

PR #48, opened 2023-05-09
Use our own Mermaid rendering component vs. remark plugin

### LINKED ISSUE #44: Show code for mermaid diagrams
Currently we hide the source code for mermaid once we render responses. Should also render in user requests before the response starts coming in. At the moment we only render after response completed

humphd:
Fixes #44

This change updates how we use Mermaid.  Previously I was using a Remark plugin, but doing so meant we couldn't control how code vs. diagrams were rendered.  Also, we couldn't easily control the theme that was used.

I'm now wrapping the `mermaid` library directly in a `MermaidPreview` component, similar to how HTML works.  The advantage of this is that I can get light/dark versions of the diagrams:

Screenshot 2023-05-09 at 2 03 02 PM
Screenshot 2023-05-09 at 2 02 54 PM

The Mermaid library isn't meant to be used in React, and I had to add a `setTimeout()` to get it to render perfectly.  However, I think it's OK.

Using the library directly, we can now update to the latest version more easily.  All of the Remark plugin and React versions are using old versions.

### DIFF
--- a/src/components/MermaidPreview.tsx
+++ b/src/components/MermaidPreview.tsx
@@ -1,0 +1,79 @@
+import { useEffect, useRef } from "react";
+import { Card, CardBody, IconButton, useClipboard, useToast } from "@chakra-ui/react";
+import mermaid from "mermaid";
+import { TbCopy } from "react-icons/tb";
+
+import { unique } from "../utils";
+
+type MermaidPreviewProps = {
+  children: React.ReactNode & React.ReactNode[];
+};
+
+const MermaidPreview = ({ children }: MermaidPreviewProps) => {
+  const { onCopy, value, setValue } = useClipboard("");
+  const toast = useToast();
+  const diagramRef = useRef<HTMLDivElement | null>(null);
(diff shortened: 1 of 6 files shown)


tarasglek:
It's odd that we have titles and aria labels(i think aria labels are redundant when you have a title), but no tooltips. 
```
        aria-label="Copy Diagram to Clipboard"
        title="Copy Diagram to Clipboard"
```

src/components/MermaidPreview.tsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  // Render the diagram as an SVG into our card's body
+  useEffect(() => {
+    const diagramDiv = diagramRef.current;
+    if (!diagramDiv) {
+      return;
+    }
+
+    const renderDiagram = () => {
```
tarasglek, line 37: Feels like this code should only do svg insertion once. Yet when using branch it seems to do it every time a message is wip and all components get disabled then re-rendered

PR #56, opened 2023-05-11
Streaming buttons

### LINKED ISSUE #51: Move pause/resume/cancel buttons into streaming message
Follow up from https://github.com/tarasglek/chatcraft.org/pull/46#issuecomment-1541592561:

>  I think the buttons should live on/near current message being written.

Right now the buttons to control streaming (pause/resume, cancel) are placed in the keyboard hint area.  Let's try moving them into the current message.  Perhaps we can do smaller buttons in the bottom-right corner or bottom-middle. If we put them there, making them less distracting might be good (e.g., icons vs. words).

### LINKED ISSUE #52: Don't hide previews (HTML, Mermaid) in old messages while replying
Follow-up from https://github.com/tarasglek/chatcraft.org/pull/48#issuecomment-1541603409:

>  it hides all mermaids during reply. I think it should only hide the in-progress one, my original code did that.

The `Markdown` component is only showing previews when the `previewCode` prop is passed in.  We did this to reduce flicker/crashing when previews might crash (they won't now) and to deal with re-renders of the messages.  We should make the previews stick around once rendered, and not re-render old messages.

humphd:
Fixes #52 
Fixes #51 

This fixes and improves a bunch of things.

To start, I've moved the Cancel and Pause/Resume buttons out of the prompt form and below the streaming message.  In order to make this work, I had to fix a bunch of other things:

- fix the way that the pause, resume, etc. hook functions and state work
- move the "mouse down pauses" functionality, and remove "mouse up resumes."  I think this makes way more sense, and "feels right" to me when I use it.  See if you agree.
- reduce re-renders in a bunch of places by memoizing things, which fixed the HTML/Mermaid previews
- refactored the streaming message and streaming buttons into a new component, `NewMessage.tsx`

I've also made it so that when a Mermaid diagram crashes, if you click "copy" it will copy the error vs. the diagram SVG, so you can prompt the AI to debug it.

I've also fixed a bug I hit during testing where the AI would return `text` vs. `mermaid` for mermaid diagrams, making it so they didn't render.  I've included more instructions in the system prompt.

I think these changes make things way smoother.  Test and let me know if you think we need to change anything.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,7 @@ import useChatOpenAI from "./hooks/use-chat-openai";
 
 function App() {
   // When chatting with OpenAI, a streaming message is returned during loading
-  const { streamingMessage, callChatApi, cancel, isPaused, pause, resume } = useChatOpenAI();
+  const { streamingMessage, callChatApi, cancel, paused, resume, togglePause } = useChatOpenAI();
   // Messages are all the static, previous messages in the chat
   const { messages, tokenInfo, setMessages, removeMessage } = useMessages();
   // Whether to include the whole message chat history or just the last response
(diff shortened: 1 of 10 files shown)


tarasglek:
This is good to land with prompt changes removed. Works much better now.

humphd:
Reverted system prompt change in https://github.com/tarasglek/chatcraft.org/commit/5f3fef1820623b81cbfd5330fd6ebeb262a1cfa0  and landed.

PR #63, opened 2023-05-12
Switch to a different clipboard hook for API key copy in prefs modal

### LINKED ISSUE #58: copy button doesn't actually copy

humphd:
Fixes #58

See if this works for you.  I switch the clipboard hook for this use case.


tarasglek:
Weird that there are 2 ways to do this and one of them doesn't work.

humphd:
I could make it work with the other Chakra hook, but to do it I have to always copy the key to the clipboard when the modal opens. This seems wrong to me.  Using this other one, I can do it only when needed.

PR #64, opened 2023-05-12
Anthropic Claude support wip

tarasglek:


humphd:
We'll have to think about how to deal with needing different API keys in the initial UI + prefs modal and `settings`, and to show only models you have API key for vs. all in drop-down.

humphd:
Here's a concept for the "Welcome" UI we show:

Screenshot 2023-05-12 at 12 55 38 PM

So if the user has no API keys, we hide the prompt, and show this new Welcome Message, which also includes the form to enter your key(s).

Is there other stuff we'd need on first-run?  I think we should only show the "Login with GitHub" stuff when you need it (e.g., when sharing) vs. doing it as soon as you land on the page.

If you like it, I can do a PR.

tarasglek:
Yeah that's how old branch worked I like how login is only done when required. 
For api key we can ask them to paste either key in...can validate the model key belongs to and pick cheapest model. This keeps setup simple.

tarasglek:
We might need to proxy calls to anthropic. I don't have admin rights to see if we can adjust CORS policy better. In meantime I filed a langchain issue to enable custom anthropic URLS, that's a blocker for now. https://github.com/hwchase17/langchainjs/issues/1255

humphd:
Can we close this in favour of #161?

PR #66, opened 2023-05-12
Draft UI experiment of new Welcome page

humphd:
Code from https://github.com/tarasglek/chatcraft.org/pull/64#issuecomment-1546029803. This isn't ready, but I'm dumping into a PR so it's easy to find later.


tarasglek:
Ooh the concept of rendering options inline in chat is so good. This would also work well for retry-with-different model workflow. I like this a lot better than modal dialogs

humphd:
Closing in favour of #122.

Issue #62, opened 2023-05-12
Tool Use and/or Allow tracing langchain with chatcraft


tarasglek:
* Add notion of tool messages that are shown differently
* WebRTC for connecting to chatcraft to local dev env per https://github.com/tarasglek/chatcraft.org/pull/68
* Tracing ui for langchain:
Chatcraft could be used to experiment with langchain.js. We might need threads for this...but current state of the art is not that hard to improve upon:
https://twitter.com/marktenenholtz/status/1656657288340590600/photo/1

Would give chatcraft a new use a live repl for langchain. Pretty sure this would be useful for developing chatcraft itself

humphd:
I'm not sure what to do with this.  @tarasglek can you refine this, based on recent changes, or close?

PR #68, opened 2023-05-16
run js code and send response back to model

tarasglek:
This change adds a run button which executes javascript and pastes return value into a reply to bot.
image
Here is a sample tool made this way:
image

This is a pretty sick feature when combined with websockets. Attached sample convo. 
[websocket-chatcraft.html.zip](https://github.com/tarasglek/chatcraft.org/files/11487819/websocket-chatcraft.html.zip)

Used websocat exposed via cloudflare to connect to my computer
`websocat  -E ws-listen:0.0.0.0:8810 sh-c:'TERM=v100 bash -i 2>&1' --binary`

`cloudflared tunnel --url localhost:8810`

Currently i force-feed js output back into model. 
Thoughts:
* I think output should probably go into input box instead, not sure. 
* Model could potentially write JS to go into infinite feedback loop and take over the world :)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -139,6 +139,7 @@ function App() {
             isPaused={paused}
             onTogglePause={togglePause}
             onCancel={cancel}
+            onPrompt={onPrompt}
           />
 
           {
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -9,11 +9,16 @@ import {
   Text,
   Box,
 } from "@chakra-ui/react";
-import { TbCopy, TbDownload } from "react-icons/tb";
+import { TbCopy, TbDownload, TbRun } from "react-icons/tb";
 
-type PreHeaderProps = { language: string; children: ReactNode; code: string };
+type PreHeaderProps = {
+  language: string;
+  children: ReactNode;
+  onPrompt?: (prompt: string) => void;
+  code: string;
+};
 
-function CodeHeader({ language, children, code }: PreHeaderProps) {
+function CodeHeader({ language, children, onPrompt, code }: PreHeaderProps) {
   const { onCopy } = useClipboard(code);
   const toast = useToast();
 
@@ -47,6 +52,27 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
     });
   };
 
+  const handleRun = async () => {
+    if (!onPrompt) {
+      return;
+    }
+    let ret = undefined;
+    try {
+      let evalResult = eval(code);
+      if (evalResult instanceof Promise) {
+        evalResult = await evalResult;
+      }
+      if (typeof evalResult !== "string") {
+        evalResult = JSON.stringify(evalResult);
+      }
+      // this probably needs smarter escaping
+      ret = "```\n" + evalResult + "\n```";
+    } catch (error: any) {
+      ret = error instanceof Error ? `${error.name}: ${error.message}\n${error.stack}` : `${error}`;
+    }
+    onPrompt(ret);
+  };
+
   return (
     <>
       <Flex
@@ -62,6 +88,18 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
           </Text>
         </Box>
         <ButtonGroup isAttached pr={2}>
+          {(language === "js" || language === "javascript") && onPrompt && (
+            <IconButton
+              size="sm"
+              aria-label="Run code"
+              title="Run code"
+              icon={<TbRun />}
+              color="gray.600"
+              _dark={{ color: "gray.300" }}
+              variant="ghost"
+              onClick={handleRun}
+            />
+          )}
           <IconButton
             size="sm"
             aria-label="Download code"
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -45,10 +45,11 @@ const fixLanguageName = (language: string | null) => {
 
 type MarkdownProps = {
   previewCode?: boolean;
+  onPrompt?: (prompt: string) => void;
   children: string;
 };
 
-function Markdown({ previewCode, children }: MarkdownProps) {
+function Markdown({ previewCode, onPrompt, children }: MarkdownProps) {
   const style = useColorModeValue(oneLight, oneDark);
 
   return (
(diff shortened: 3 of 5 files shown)


tarasglek:
you are my linux shell assistant. 
1. When i ask for a command you will output just the code
```javascript
// <reasoning behind command>
runCMDOverWebSocket("<command>")
```
and nothing else.
I will run command for you and tell you response. 
2. when you hear the response you will render it as a mermaid diagram with valid syntax

tarasglek:
1. write async javascript function(cmd) that connects to wss://folders-brochure-duty-teachers.trycloudflare.com or reuses pre-existing window.socket there and sends a message containing cmd+"\n", then buffers responses into a string for 2 seconds, returns that string. 
2. prior to re-using window.socket, ensure it's connected.
3. attach function to window.runCMDOverWebSocket
4. console log JSON.stringify buffer right after adding to it

tarasglek:
` websocat  -E ws-listen:0.0.0.0:8810 sh-c:'TERM=v100 bash -i' --text`

tarasglek:
you are my linux shell assistant. 
1. When i ask for a command you will output just the code
```javascript
// <reasoning behind command>
runCMDOverWebSocket("<command>")
```
and nothing else.
I will run command for you and tell you response. 
2. when you hear the response you will render it as pretty inline html, possibly with svg,css,etc
```html
....
```

src/App.tsx
```diff
@@ -109,6 +109,7 @@ function App() {
     },
     [messages, setMessages, singleMessageMode, setLoading, setShouldAutoScroll, callChatApi, toast]
   );
+  (window as any).onPrompt = onPrompt;
```
humphd, line 112: This isn't very "Reacty".  Can you pass this down on props through `MessageView` > `Message` > `Markdown` > `CodeHeader` instead?  Similar to what we do with prompt form below in same file:

```tsx
            <PromptForm
              onPrompt={onPrompt}
```

We can convert it to use a React Context later if we find we need to expose the `onPrompt` as a hook deep in the component tree.

src/components/CodeHeader.tsx
```diff
@@ -47,6 +47,24 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
     });
   };
 
+  const handleRun = async () => {
+    let ret = undefined;
+    try {
+      let eval_result = eval(code);
```
humphd, line 53: This seems fairly unsafe, and should probably get sandboxed later.  Figma has a [cool post](https://www.figma.com/blog/how-we-built-the-figma-plugin-system/) on how they built something similar, fyi.  We could run it in an iframe on a different origin, or in a worker out of main thread, or use something like https://github.com/endojs/endo.

Also prefer camelCase in JS.

src/components/CodeHeader.tsx
```diff
@@ -47,6 +47,24 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
...
+      if (eval_result instanceof Promise) {
+        eval_result = await eval_result;
+      }
+      if (typeof eval_result != "string") {
```
humphd, line 57: `!==`

src/components/CodeHeader.tsx
```diff
@@ -47,6 +47,24 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
...
+        eval_result = JSON.stringify(eval_result);
+      }
+      // this probably needs smarter escaping
+      ret = "```\n" + eval_result + "\n```";
+    } catch (error: any) {
+      ret = error instanceof Error ? `${error.name}: ${error.message}\n${error.stack}` : `${error}`;
+    }
+    (window as any).onPrompt(ret);
```
humphd, line 65: I wonder if you should decorate this with info that indicates how to interpret the result?  It could be as simple as:

```
Result:

...your ret value here
```

Later we could use a different Message type somehow, so the UI looks slightly different.

src/components/CodeHeader.tsx
```diff
@@ -62,6 +80,20 @@ function CodeHeader({ language, children, code }: PreHeaderProps) {
           </Text>
         </Box>
         <ButtonGroup isAttached pr={2}>
+          {language === "js" || language === "javascript" ? (
```
humphd, line 83: Drop the ternary and only render if the language conditions are `true`

```tsx
{ (language === "js" || language === "javascript") && <IconButton.../>) }
```

humphd (changes requested):
I left some feedback.  This is a pretty interesting idea.

Your websocket logic could live in a worker thread, along with the `eval()` execution environment.  If running this against your local machine in a shell is a goal, it might be cool to offer a "Download" link to get a client you can start in your shell.

I also wonder if you could use Tailscale in your browser to facilitate this, see https://twitter.com/bradfitz/status/1451423386777751561, https://tailscale.com/blog/ssh-console/, and https://www.npmjs.com/package/@tailscale/connect.

src/components/CodeHeader.tsx
tarasglek: I didn't know this pattern

src/components/CodeHeader.tsx
tarasglek: i'm not clear what you mean

src/components/CodeHeader.tsx
tarasglek: WebRTC is the way to do this properly. Eg setup signaling via https://github.com/gfodor/p2pcf . Though I'm not sure that will be as slick as websockets via cloudflared.

src/components/CodeHeader.tsx
tarasglek: I'll leave that for a follow up. Need to think about how to integrate tools into UI.

humphd: Sounds good.  Soon I want to create multiple `Message` UI types (e.g., "ChatGPT" vs. "GPT-4" vs. "Human" etc) and once I do, adding a "result" Message type would be easy.  Being able to differentiate who/what wrote each part of the chat will be nice when we start mixing AIs, tools in a single chat.

humphd (approved):
Looks good.  I don't know if we should pass `onPrompt` to `<NewMessage />`, since it's a streaming message that isn't yet complete, and will get converted to a `Message` once it is.  However, making it optional there would require changing your types on `onPrompt` everywhere to be optional.

tarasglek:
> Looks good. I don't know if we should pass `onPrompt` to `<NewMessage />`, since it's a streaming message that isn't yet complete, and will get converted to a `Message` once it is. However, making it optional there would require changing your types on `onPrompt` everywhere to be optional.

I can change it to optional, just say the word. It's awkward both ways :)

tarasglek:
Captured some of the followups in https://github.com/tarasglek/chatcraft.org/issues/62

humphd:
If you're OK doing it, I would recommend that `onPrompt` not get passed down to `NewMessage`, and that you make the type on all those props be `onPrompt?: ...` and deal with the function not existing when you call it:

```ts
const handleRun = async () => {
  if(typeof onPrompt !== 'function') {
    return;
  }
  ...
```

humphd:
Actually, add it to your conditional rendering too:

```tsx
{ (language == 'js' || language === 'javascript') && onPrompt && <IconButton .../> }

Issue #13, opened 2023-04-24
Better storage for messages


humphd:
Currently, the messages are stored as an `Array` in `localStorage`.  This is OK, but it has limits, since we can't store more than 5M.  If we want to store everything we ever do, search it, etc. we could improve things by moving to `indexeddb`.

We could consider using a lib like [dexie](https://dexie.org/docs/Tutorial/React), which has interesting hooks to do live queries like this:

```ts
const messages = useLiveQuery(
    async () => {
      //
      // Query Dexie's API
      //
      const mesages = await db.messages
        .where('some search criteria...')
        .toArray();

      // Return result
      return messages;
    },
  );
```

Using this, we could create separate threads vs. only have the current chat session, give the ability to search across those threads, pick up where we left off in a past conversation, etc.

There might be something better than `dexie`, it's just one option I know of that supports React well (I haven't used it before).

tarasglek:
Note this ties in with indexing via https://twitter.com/nfcampos/status/1651915458222718976

humphd:
Let's figure out what we actually want to store.  Things I'm considering:

1. Chats (i.e., lists of Messages like we have now), but augmented with date info (e.g., each chat has a date when it was created and each message has date info).  Having dates (or some other unique id) will make it easier in UI to use as a `key` or React
2. Documents.  Could be hunks of code you paste in, files you upload.  I'm imagining being able to drag one or more files into the app and have them get used as part of the chat somehow
3. Embeddings.  For both 1. and 2., it would be useful to create and store embeddings so we can do semantic search later on for building context
4. Settings.  I think what we do now with `localStorage` is fine for this.

It probably makes sense to use IndexedDB for this, since it gives us lots of room.  However, langchainjs also has this notion of a [Vector Store](https://js.langchain.com/docs/modules/indexes/vector_stores/), which combines document + embedding storage and retrieval/querying.  Most of the [ones they support](https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/) are cloud-based, but we could start with [MemoryVectorStore](https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/memory), and if it works, write an IndexedDBVectorStore based on [their code](https://github.com/hwchase17/langchainjs/blob/2886ad92d97c1f9efdfbc1906d61e57265b16b5a/langchain/src/vectorstores/memory.ts).

[...]

PR #69, opened 2023-05-17
Add ChatCraftMessage classes, support different message types

humphd:
In preparation for a bunch of feature work (backend db, supporting more models, file uploads, etc) I needed to refactor how the messages are stored, modelled, and displayed.

This change introduces different message types in the UI.  For example, we can now see which model made which response (green is ChatGPT, purple is GPT-4):

Screenshot 2023-05-17 at 10 46 07 AM

Internally, I've extended the langchain message types to add some metadata to each message (e.g., a unique `id` and which `model` was used, if any).  I've created our own classes (e.g., `ChatCraftMessage`, `ChatCraftAiMessage`, `ChatCraftHumanMessage`, `ChatCraftSystemMessage`) which extend all the langchain types.

I've also fixed a few things I hit along the way:

- disable the "Run" button when a message is already streaming in
- switch from `eval()` to `new Function()` to get around a rollup warning
- better display for error messages when "Run" fails

I've left the code to support the old serialization format, but upgrade messages to the new format.  Eventually we can remove this, but for now it will mean that people won't lose data as we upgrade the format on first use.


humphd:
https://15d7b36f.console-overthinker-dev.pages.dev/helloworld works, but the root `/` is now broken.  I must need to specify that I want to serve the static content there explicitly...

src/ChatCraftMessage.ts
```diff
@@ -0,0 +1,81 @@
+import { BaseChatMessage, type MessageType } from "langchain/schema";
+import { nanoid } from "nanoid";
+
+// We store ChatCraft messages as flat JSON
+export type SerializedChatCraftMessage = {
+  id: string;
+  type: MessageType;
+  model: GptModel | null;
```
tarasglek, line 8: i would serialize model params here..maybe in a deduplicated form

humphd:
Can you elaborate on these comments?  What you're saying is connected to what I was discussing in https://github.com/tarasglek/chatcraft.org/issues/13#issuecomment-1528852244.

Currently we don't expose any knobs for altering the [model params (temperature, etc.)](https://github.com/hwchase17/langchainjs/blob/3a6362f/langchain/src/chat_models/openai.ts#L103-L119), is that something you want to do?

What do you want in this JSON?  It would be good to pin it down now, since it will affect storage (sqlite, R2) so we don't have to support too many formats.

humphd:
I'm going to land this so that I can build the sharing stuff on top.  I'll refine the chat/message format stuff as part of that.

PR #70, opened 2023-05-17
Prototyping Pages Functions

humphd:
Trying to get Pages Functions wired in so we can do GitHub auth, sharing, etc.  The docs make it sound like this is all I need to do.  Let's find out!

### DIFF
--- a/functions/api/login.ts
+++ b/functions/api/login.ts
@@ -1,0 +1,68 @@
+// TODO: check token https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28#check-a-token
+
+interface Env {
+  CLIENT_ID: string;
+  CLIENT_SECRET: string;
+}
+
+// Handle CORS pre-flight request
+export const onRequestOptions: PagesFunction = async () => {
+  return new Response(null, {
+    headers: {
+      "Access-Control-Allow-Origin": "*",
+      "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+      "Access-Control-Allow-Headers": "Content-Type",
+    },
+  });
+};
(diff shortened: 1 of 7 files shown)


humphd:
I want to use this PR to add GitHub OAuth login support.  Here's a list of what I think we need to do in order to make that happen:

1. On GitHub, create a [new OAuth App](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app).  For the **Authorization callback URL** in production, I wonder if we should put all functions under `functions/api/*` so it would be https://chatcraft.org/api/login.  For scopes, I don't think we need any, since all we want to do is know that this user is who they say that they are, and to access public info (e.g., we don't want to create/manage any GitHub objects, right?)
2. Get the `CLIENT_ID` and `CLIENT_SECRET` for the OAuth app.
3. Add these to CloudFlare Secrets: **Account Home** > **Workers & Pages** > chatcraft.org project > **Settings** > **Environment variables** > **Add variables** under **Production** and select **Encrypt** then **Save** (see [docs](https://developers.cloudflare.com/pages/platform/functions/bindings/#secrets))
4. We can then use these values in our functions via `context.env.*`
5. I think I can adapt the code in https://github.com/gr2m/cloudflare-worker-github-oauth-login/blob/main/github-oauth-login.js to have it work as a function vs. worker for the actual login bit.
6. Later, when we add functions that require authorization, we can probably convert this to [middleware](https://developers.cloudflare.com/pages/platform/functions/middleware/), but I've never done this so I can't say more.

humphd:
I think this is ready, I'm going to merge so I can test on prod and then do follow-ups.

Issue #61, opened 2023-05-12
Make token-tracking work more like a receipt


tarasglek:
This is related to https://github.com/tarasglek/chatcraft.org/issues/59 should make token-tracking function more like a receipt that tells you how much it cost to generate the message near the model details.

This will be more useful once we add claude. 

Furthermore current token-tracking is misleading:
* it's undercounting as it simplify reflects cost if all currently visible messages were produced by gpt. Instead it should be updated every time we get a response back and stored in local storage.
* It's overcounting gpt-4 price by not differentiating different input/output costs

tarasglek:
Found this code to track token usage. This is also the key to tracing langchain agents. https://github.com/hwchase17/langchainjs/issues/1079#issuecomment-1532315417

humphd:
Started to look into this, but ran into https://github.com/hwchase17/langchainjs/issues/1682

PR #76, opened 2023-05-31
Add IndexedDB database backend, routing, search, sharing sidebar

### LINKED ISSUE #13: Better storage for messages

### LINKED ISSUE #34: TODO: Share/save/project functionality
Sometimes I can write whole files using chatcraft. Ideally i'd leave a link to convo that generated them at top of the file.

* Going to that link should restore conversation as it was. perhaps link is a hash of conversation stored in s3?
* convo should also have a human-written description of what the convo is about..gpt could prefill it.
* Restoring basically prefills current msg history
   - should prompt on whether to erase current history?
   - perhaps we should have explicit support for named conversations..where importing a convo would create a new convo "imported blabla" which could be renamed

Another interesting approach is to make a comment that summarizes the prompts needed to recreate the file.  This can be done by deleting all the responses, example:
image
This is cool as it doesn't depend on chatcraft, any gpt implementation could reimplement similar code from such prompts.

### LINKED ISSUE #42: ability to "fork" a chat thread
would be amazing if I could take an existing chat thread and then fork and duplicate it into a new thread so I can continue off coding a different section of an app

e.g. 

here's the codebase ...

fork1 -> code the database migrations for me
fork2 -> code the frontend for me
fork3 -> fix this bug for me

and have them all running at the same time

### LINKED ISSUE #72: UI for sharing, settings
In order to get the "sharing" stuff in, I need to rework the UI.  While I'm doing it, I'm also trying to think about other things that need to get added/moved.  Here's a concept:

Screenshot 2023-05-19 at 3 32 57 PM

In this updated UI, notice a few things:

1. I'm logged in (upper-right), showing my GitHub avatar vs. the stock one.  If I'm not logged in, it would show a generic avatar.
2. The icon to Expand the chat has been moved to the left-hand side.  On the right-hand side, 2 new buttons are added: 1) Settings; 2) Share.  Both can expose more UI to the user.  With Settings, we'd show various options to control the model, etc. and with Share we'd let you Copy, Download, or post and Share a chat.  These could either be a modal, or we could slide-up a panel that's attached to the prompt form, then slide it away again.

Another question: how important is the "Single Message Mode" checkbox?  Could it get moved into Settings?

Thoughts?

humphd:
Fixes #72
Fixes #42
Fixes #34
Fixes #13

Apologies for doing this as such a big PR, but this essentially changes the underpinnings of the whole app, which affected everything.  I'll try to discuss and demo everything, and you can ask questions on whatever isn't clear.

## Database

I'm using [Dexie](https://dexie.org/), which is built on top of IndexedDB, and provides some nice reactivity for queries.  Dexie allows for [import/export](https://dexie.org/docs/ExportImport/dexie-export-import) using a JSON format, which I've included via an `Export` button in the User Settings dialog.  This will let us move to another backend later if we decide to innovate on that part of things.  However, I don't think we're going to improve on what Dexie offers.

The database is split into two tables:

1. `chats` - this is a `ChatCraftChat` object, but the `messages` are only stored as `id`s vs. full messages
2. `messages` - this is a `ChatCraftMessage` object, but we also include the `chatid` of the chat which owns it.

I've split the `messages` out of the `chats` so that search is easier to do (i.e., I only do search on message contents).

Both `chats` and `messages` have an autogenerated [nanoid](https://www.npmjs.com/package/nanoid) `id`.  This is used in various ways, including routing.

## Routing

We used to only have a single route, `/`.  Now we have the following routes (defined in `src/router.ts`):

- `/` the default entry point, redirects to `/new` and creates a new chat by default
- `/new` creates a new chat and redirects to it
- `/c/:chatId` loads a chat by `id` from the database and displays it
- `/c/:chatId/fork/:messageId?` forks the chat at `chatId` (i.e., creates a new chat with same messages).  If a `messageId` is present, forks from that message vs. all messages (e.g., fork half-way through a chat)
- `/c/:user/:chatId` loads a shared chat over the network
- `/s?q=...` searches across all `messages` for `...` and shows results

## UI

The UI has been updated.  Here's what a chat looks like:

Screenshot 2023-05-31 at 3 36 54 PM

I can also open the sidebar to show my shared chats (I don't have any at the moment, and need to land this on prod to fully test that bit):

Screenshot 2023-05-31 at 3 37 29 PM

Notice the URL: http://localhost:5173/c/cOwwpnz-l-3TaTcoMxFDB. This shows that I'm in the chat with `id=Owwpnz-l-3TaTcoMxFDB`. If you open that same URL in another tab, the messages will be synced (i.e., they display the same live data).

Notice too that messages now have a date/time link.  The message in this view has the URL: http://localhost:5173/c/cOwwpnz-l-3TaTcoMxFDB#abUKHyapxo576etk4pSsn, which is chat `id=cOwwpnz-l-3TaTcoMxFDB` and `messageId=abUKHyapxo576etk4pSsn`. This `#messageId` hash lets me deep-link into a conversatation

As a chat grows, each `message` gets its own unique link, which I can open in a new tab, bookmark, etc:

Screenshot 2023-05-31 at 3 41 09 PM

At the bottom of the form, there are two links:

1. `New` - starts a new blank chat.  
2.  `Fork` - starts a new chat using this chat as the starting point

Both can be right-clicked to open in a new tab/window.

I can also Fork from a message in the middle of a chat using the menu:

Screenshot 2023-05-31 at 3 43 05 PM

## Search

In the Header there is now a Search form.  I can type some text and press Enter or click the Search button:

Screenshot 2023-05-31 at 3 44 25 PM

Searches look for your input text anywhere in the `message` text, across all messages in your history.  They are returned in reverse chronological order.

You can open the chat to which the message belongs by clicking on the Date/Time link.  Or you can fork a new chat from that message using the menu.

## Breaking Changes and Bugs

- This will lose the user's current chat in `localStorage`.  However, after the first load, the db will be stable.  Users can also request that the database be persisted (not deleted by browser without permission) in settings:

Screenshot 2023-05-31 at 3 48 27 PM

- The token counting is not connected to the new backend yet.  We can fix as part of #61.  We also need to remove `src/hooks/use-messages.ts`

- `App.tsx` has been renamed to `Chat.tsx`, and there is some code duplication between `Chat.tsx` and `Search.tsx` that we could clean up with React Router child routes and `<Outlet />`s.  It wasn't enough to worry about in this change.

- I need to track a setting in storage for whether or not the sidebar is default open/closed, but didn't bother yet.

- There's a slight difference between the types for what's in the database (`src/lib/db.ts`) and the `ChatCraftMessage` and `ChatCraftChat` classes.  I think I've dealt with all of it, but we could probably add some more helper methods to convert between types.  For now, this is working.

- Current shared chats are going to break, since the URL structure is different.  I think we should just empty the bucket, but if something there is important, we can manually fix.

## Other Changes

- I moved the main UI over to use CSS Grid instead of all Flexbox, and it's working a lot better.

- I've introduced a `readonly` mode to the `Chat` so that we can show a chat you don't own/have in the db (e.g., load a share). 

- I had to change a bunch of `props` to deal with passing both `messageId` and `chatId`
 
- I've made the "enter your api key" instructions less of a deal-breaker.  Right now you can't get passed them if you want to read.  Now you can.

- I added a simple summarization algorithm for the default summary we use when not invoking OpenAI to generate.  The only time we need this is for sharing.  See `src/lib/summarize.ts`.  There's probably something smarter we can do without needing an AI/model, but this works for now, and users can click `Generate` to have OpenAI write one instead.


tarasglek:
This is really good.

Thoughts:
* I would prefer to have 'new'/'fork' chat open a new window now by default. This helps reinforce the idea that old chats don't go away
* When you click new, it's not clear that chat history is retained...I would display an indicator saying "131323 saved chats" next to search box
* Not clear what share urls look like now.

Issues:
* Our message asking for key also shows the helpful assistant message above it
* search should be case-insensitive by default

PR #85, opened 2023-06-02
Going to / should redirect to most recent chat if possible

humphd:
I'm not happy with the way that the `/` route works.  Currently, it always redirects to `/new`, so you end up with lots of chats you might not want in the db.

I think this change is better: when you go to `/` it will get the most recent chat and open that.  If you want to make a new one, you opt into that manually.

See if you agree.


tarasglek:
yes, this makes more sense!

PR #90, opened 2023-06-03
Refactor functions to use JWT, cookie-based auth

humphd:
Based on our conversation about how to improve the auth situation, I've rewritten the back-end functions and front-end user/api code.  As with all auth related stuff, I need to test this fully on prod, but I'll wait to land this before getting another set of eyes on the approach. Here's what this does:

A user must authenticate with GitHub via OAuth, which is the same as before.  However, previously we would get back an `access_token` from GitHub, and we'd pass that along with the user profile info to the front-end via query params on a redirect URL.  This was problematic because the token expired in 8 hours, and we had no way to refresh it; thus the user looks logged in, but isn't.

With this change, a user authenticates with GitHub via OAuth, and we then use their profile info to create our own ChatCraft token: a JWT that we sign (I've added a new `JWT_SECRET` to the Cloudflare ENV).  We set this token in the browser's cookies instead of storing it in the client storage, and it rides along with every request to the back-end functions.

I've set the expiry for this token + cookie to 30 days (we can lower this easily, if you want).  Whenever the user interacts with the back-end API, we always refresh the token+cookie automatically, thus extending their login for another 30 days.  So if you share something once in 30 days, you will stay logged in.  Otherwise, the browser will log you out at the end of those 30 days.

On the client, I've redone the `useUser()` hook to use the cookies vs. localStorage as its backend.  Therefore, every load of the app is always going to reflect the actual login status of the user.  To accomplish this, I've decided against making the cookie HTTP-Only.  Originally I wrote it that way (and we might decide to go back to it), but because we're doing Secure + SameSite + Domain settings on the cookie as well, it's not going to get leaked.

If you want it to be HTTP-Only, I could add another endpoint to get the user's info using the secured token in their cookies.  It just requires additional network requests and re-renders to get the `user` all over the place, which kind of sucks.

Take a look through this and see if you're OK with the approach.  If so, I'll land and test/fix on prod.


tarasglek:
Your write-up makes sense. I looked at the code, but don't know enough to see anything wrong with it. Lets land it

humphd:
It looks like I need to switch to a different JWT module, since `jsonwebtoken` depends on too much stuff in node.js that isn't available in CloudFlare.

humphd:
[jose](https://github.com/panva/jose) is compatible with CloudFlare workers, so I've switched to that.  Build works again.

humphd:
The next time I have a few hours of free time, I'll land this and test/fix on prod.

humphd:
I had a thought: there's no reason to choose between allowing browser to access user info in a cookie AND securing the token in an HTTP-Only cookie, I can use 2 cookies: `id_token` (available to browser, has user info) and `access_token` (only available to server functions, can't be stolen).

This simplifies things greatly, since I can now rely on the user's `access_token` for using API endpoints, but give the React app access to the user info via the `id_token`.

Also, I can indefinitely extend the life of the cookie whenever the user hits an API endpoint without creating new tokens that can get stolen or re-used.  Good UX and better security!

I've updated the code.

tarasglek:
I'm a little confused. I see that you are returning accessToken as a cookie. Do you mean that this cookie is not available to react so can't be stolen in JS?

humphd:
OK, I broke down and wrote tests.  Apparently CloudFlare Pages Functions aren't really testable "yet...", but I was able to get coverage for a bunch of things via mocks.

humphd:
I'm going to land this and try fixing on prod.

humphd:
I had to fix a few things, but it works now!  Try logging in, look at your Cookies in Dev Tools, try Sharing, try Logging out, etc.

Issue #91, opened 2023-06-06
When sharing, login should return to the chat you were working on


tarasglek:
This one caught me by surprise. I had one old chat and a newer chat...eg if i open a new chatcraft it opens newer chat.

I then go to share older chat, it asks me to login and then returns to newer chat...i seemingly lose the older chat..luckily i could get back via search.

humphd:
Help me reproduce.

So if I go to `chatcraft.org/` right now, it takes me to the latest chat, which is https://chatcraft.org/c/zJAbsFS1ZTzY6tQI5Daki for me.

Then you are making a new chat by clicking ***New > Blank Chat...***?

Now I have https://chatcraft.org/c/d9mMqEvKTh-6eQwbAxGQY (new) and https://chatcraft.org/c/zJAbsFS1ZTzY6tQI5Daki (old).

When I open chatcraft/ again, it takes me to https://chatcraft.org/c/d9mMqEvKTh-6eQwbAxGQY (new).

That seems right to me, since https://chatcraft.org/c/d9mMqEvKTh-6eQwbAxGQY (new) is the most recent.

Maybe what we need here is to add a **Recent** list to the sidebar, which shows your last 5-10 chats above the Shared chats?

tarasglek:
made a recording..will share it once it finishes uploading

humphd:
See #93 too.

tarasglek:
see vid
https://github.com/tarasglek/chatcraft.org/assets/857083/09924153-6815-41cc-81fb-ddbd7f67ad5e

humphd:
Got, so this is about the fact that doing a login redirects back to `/` vs `/:currentId`.  That's a bug for sure, nice find.

PR #93, opened 2023-06-06
Show recent chats, allow deletes

humphd:
Part of #91

This adds **Recent Chats** to the sidebar (defaults to show previous 5, but you can click **Show More...** to load another 10, etc).

I've also added UI to let you **Delete** old chats.

The only risk I see with delete is that it means you could delete a chat you forked off, and currently I don't do a deep copy of a fork'd chat (i.e., I share the messages) so that we don't have a done of duplication in search results.

Screenshot 2023-06-06 at 12 21 17 PM

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -3,14 +3,16 @@ import {
   Box,
   Button,
   Flex,
+  Link,
   Text,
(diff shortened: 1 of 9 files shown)


tarasglek:
merged it, need to play with it to see how it feels

PR #94, opened 2023-06-06
Login and Logout should redirect back to same chat URL

### LINKED ISSUE #91: When sharing, login should return to the chat you were working on

humphd:
Fixes #91 

This updates the `/api/login` and `/api/logout` routes to allow passing a `?chat_id=...`.  Doing so will mean that the user gets redirected back to this chat vs. going to a different one.

Having the tests already in place for this was nice.

It's going to be impossible to fully test this until it lands on prod, so I might do that and then fix afterward.

### DIFF
--- a/functions/api/login.test.ts
+++ b/functions/api/login.test.ts
@@ -8,6 +8,7 @@ describe("/api/login", () => {
   test("/api/login without code should redirect to GitHub's OAuth login", async () => {
     const res = await handleLogin({
       code: null,
+      chatId: null,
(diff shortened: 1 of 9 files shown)


humphd:
Tested on prod.  @tarasglek try your use case again and make sure you agree, but this looks fixed to me.

PR #97, opened 2023-06-07
Fix sidebar layout shift, use pref for showing/hiding sidebar, cleanup

humphd:
The way I wrote the sidebar show/hide was causing layout shift on load, so I've reworked it.  While I was touching this code I also fixed:

- use proper visual header for Search results and Chats
- added a pref for showing/hiding the sidebar.  I store the last value in localStorage, and use it on reloads (e.g., if you hide it, it will be hidden next time, if you show it, shown next time).
- I've made the sidebar default to closed

I want to get feedback on the last point. I think this is the right default, since your sidebar will be mostly empty when you start using the app, and you'll only require it as you use things for a while.  But see if you agree.

Screenshot 2023-06-07 at 11 43 27 AM
Screenshot 2023-06-07 at 11 43 14 AM
Screenshot 2023-06-07 at 11 42 59 AM
Screenshot 2023-06-07 at 11 44 18 AM


tarasglek:
personally i would prefer to move search there too.

btw other ui nit is that messagebox expansion gets grayed out during gpt response...this is frustrating and usually you wanna see the response while model is responding...perhaps the proper fix would be to just shrink input box while it's disabled automatically.

Other thing that i noticed is that with recent changes we are often obscuring responses while they are streaming from the model and need to scroll up to follow latest text stream.

Issue #59, opened 2023-05-12
Add indicator of what model responded to message and menu to try response with a different model


tarasglek:
For example..at bottom of message add greyed out "response by chatgpt-3.5" with a little menu, that has buttons to "retry with gpt-4/claude/etc". This would also in the future support the full set of params eg temperature, prompt, etc. Maybe this would tie in with threads(https://github.com/tarasglek/chatcraft.org/issues/40), receipts(https://github.com/tarasglek/chatcraft.org/issues/61)

This would make it easy to compare models plus their params via "retry with gpt4", make for some good blog posts.

One potential UI like..is clicking on the "response by chatgpt-3.5" would open sidebar where we can tweak prompt, model, etc and click retry

humphd:
I think we have the pieces in place to explore this more easily now.  Consider:

<img width="923" alt="Screenshot 2023-06-01 at 9 31 37 PM" src="https://github.com/tarasglek/chatcraft.org/assets/427398/b16006ef-6a58-414f-afac-bd370022bb40">

We could put a "gear" icon or something near the date or name of model, and it could open an accordion showing a form with model params used to generate it.  It could also include a "Retry" button.

Or we could integrate this into the "dots" menu on the right somehow.

Let's say I have a chat like this:

- message 1
- message 2
- message 3
- message 4
- message 5

I am interested in having `message 3` get retried with a different model.  What is the UX?  I tweak the params and click "Retry" and then it sends `[message 1, message 2, message 3]` to the new model (i.e., ignores 4 and 5) and then replaces the text in `message 3` when it comes back?  Or does this delete message 3 and create a new message in its place?

tarasglek:
oooh..if we integrate it into the menu, then it will be really slick..just add a "retry-with" menu that opens up to submenu of all available models

humphd:
How about my other question though: what does retrying an arbitrary message in a chat mean in terms of the context you send?

tarasglek:
tabbed messages is an interesting concept

niallo:
Threads might be interesting too. I was thinking tabs because I like how OpenAI does it (although these aren't really functional after the first message - but maybe it would work?). I haven't fully considered all the implications tho.

<img width="645" alt="Screen Shot 2023-06-07 at 10 18 51 AM" src="https://github.com/tarasglek/chatcraft.org/assets/218162/4070fe28-95f5-401f-b425-1514f8675e67">

tarasglek:
[...]

PR #99, opened 2023-06-07
Refactor ChatCraftMessage to support AI Versions

humphd:
Part of #59

The following changes update the type system and db in order to support adding `versions` to `ChatCraftAiMessage`s.  It also does a migration to introduction a new optional `versions` property in the db.

I've removed/moved a bunch of utilities and gutted the `useMessages()` hook, since none of that code is right anymore.  The token counting needs to get fixed, but we can do that as part of #61.

I have only tested this lightly, so I'll look at it again before landing.

After this is in, the focus can switch to the UI and being able to retry an AI Message with a different model.


humphd:
This is functional now, and extends what I did previously on the types/db to add UI support for retrying AI messages with different models in order to produce new versions and then switch between those versions.

I've also included my work from #100 to allow editing any message's text, since it's connected with this work.  When editing an AI message, I add another version (i.e., so you can go back).

Here are some screenshots of what it looks like:

## Get back a response, use new "Retry..." menu options to try with another model
Screenshot 2023-06-08 at 4 04 06 PM

## New version is streamed into current message
Screenshot 2023-06-08 at 4 04 20 PM

## When a message has multiple versions, new UI lets you switch between them
Screenshot 2023-06-08 at 4 04 35 PM

## Clicking "Edit" allows editing in place
Screenshot 2023-06-08 at 4 04 49 PM

I think this opens up some interesting workflows:

- I am no longer constrained by aspects of a response I don't like.  Everything is text, and all text is editable.  Further, I am the author of all text (e.g., what the AI responds with is equally part of the "text I am working on").
- I am no longer constrained by the choice of model in a previous step.  Every response can be retried with a different model, but also with a different context (e.g., if you edit or retry previous messages, you build a new context for the response).

There are aspects of what I'm doing here that could be cleaned-up in the code, but I didn't bother to perfect things until it's clear on UX direction we want to take.  Having something to test and discuss will hopefully make that easier.

cc @niallo

niallo:
I love this approach! In particular making everything "text I am working on" and having multiple versions you can switch between. Nice job - I look forward to hopefully trying it once it is ready.

tarasglek:
> I love this approach! In particular making everything "text I am working on" and having multiple versions you can switch between. Nice job - I look forward to hopefully trying it once it is ready.

@niallo fyi we have previews for all pull requests. just click on 
image
then 
image

tarasglek:
image

This is too complex. We should combine gpt-4/versions. Eg if we already generated chatgpt version too then

|GPT-4 🔽                       | [Jun 9, 2023, 12:02 PM]
| ChatGPT-3.5 [Jun 9, 2023, 12:01 PM] |
-----------------------------
| Retry with ChatGPT-3.5  |
| Retry with GPT-4            |

This way can have multiple generations and can extend this in the future for passing other model params

tarasglek:
Other issue that if i started with chatgpt, then switched model to gpt-4, i would expect subsequent messages to use gpt-4...but this now changes your submission logic where you pick model from previous message and it gets weird with welcome message..so we can leave this for followup

tarasglek:
btw, i also think we shouldn't submit welcome message to gpt..it should just be a message from chatcraft.

humphd:
I think putting versions in the menu is not obvious, and will clutter menu.

Putting an indicator in the header of the message borrows from what GtiHub does with edited versions of comments in issues/PRs:

Screenshot 2023-06-09 at 8 05 37 AM

For example, you often edit stuff in GitHub, and I will go to an issue and see "edited" and know that it's been updated.  Similarly, if there are multiple versions of something, I'd like to have an indicator that they exist.

What does "this is too complex" mean exactly?  So, I'm hearing:

- add timestamp beside model name
- don't say "version" but somehow turn the name/timestamp into a drop-down instead?

If so, that's interesting.  The downside of this is that the timestamp is the link to this message, so you can't have it be a link and a drop-down.

Can you file unrelated stuff (e.g., "don't send welcome message...") in other issues so we don't lose?

tarasglek:
I wanna be clear, this functionality is fantastic and makes chatcraft even more of a joy to use, but I wanna simplify it more.

> What does "this is too complex" mean exactly? 

In the this pr: we to the side menu to ask another model for response..and then switch between model responses in 2nd menu..and also indicate chatgpt/gpt-4 in 3rd indicator.

That makes the titlebar cluttered with redundant info without much grouping of related functionality.

>So, I'm hearing:
> 
> * add timestamp beside model name
> * don't say "version" but somehow turn the name/timestamp into a drop-down instead?

Yes

Ok. will file separate issues for other stuff.

humphd:
Playing with this locally, here's another iteration:

Screenshot 2023-06-12 at 10 36 51 PM
Screenshot 2023-06-12 at 10 37 00 PM

I don't want to mix Actions and State in the 3-dots menu, which should all be Actions.  I think having this chevron and drop-down menu where the Avatar, Name, and Date sit makes sense.

humphd:
Pushed my code if you want to test.  It's not ready to land, but it's getting closer.

tarasglek:
Current version is very slick.

humphd:
@niallo this is live on chatcraft.org, if you want to play with it now.  Try it out and file feedback on ways to improve it.

PR #113, opened 2023-06-16
Create ChatCraftAppMessage and AppMessage for ChatCraft messages

### LINKED ISSUE #103: Do not send "I am a helpful assistant! How can I help?" message to server
It doesn't improve results to have that, costs us credits and token limit. I would replace it with a stand-in-message that disappears first time we click send.

humphd:
Fixes #103

This adds a new message type, `ChatCraftAppMessage`, and a new component to display them, `AppMessage`.

We now have two ways to work with the messages in a chat:

1. `.messages` includes all messages
2. `.nonAppMessages` all messages with app messages stripped out

The latter is suitable for sending to an LLM or sharing, the former for showing in the UI/db.

Screenshot 2023-06-16 at 1 47 48 PM

Later, we can build on this to put widgets into the UI for things like entering API key.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -138,10 +138,10 @@ function ChatBase({ chat, readonly, canDelete }: ChatBaseProps) {
         // Add this prompt message to the chat
         await chat.addMessage(new ChatCraftHumanMessage({ text: prompt, user }), user);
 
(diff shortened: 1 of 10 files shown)


tarasglek:
I like this, but i feel it would be cleaner for a sngle .messages() to include(includeAppMessages=true) param instead

humphd:
Good suggestion, I've made the change.

PR #117, opened 2023-06-16
Don't lose summary when sharing from ShareModal

### LINKED ISSUE #116: Manual summary lost when sharing
The code for updating the `summary` in a chat is not right.  In the `ShareModal` we have a way to write a summary, or have GPT summarize for you.  There is also a way to `.summarize()` a chat, which gives important sentences from each paragraph.  When we save the chat to the db, it's overwriting the `.summary` with `.summarize()`.

humphd:
Fixes #116.

### DIFF
--- a/src/components/ShareModal.tsx
+++ b/src/components/ShareModal.tsx
@@ -46,8 +46,7 @@ function AuthenticatedForm({ chat, user }: AuthenticatedForm) {
   const handleShareClick = async () => {
     setIsSharing(true);
     try {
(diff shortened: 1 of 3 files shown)


humphd:
Fixed.  I've also pushed another commit without a PR in order to improve the quality of the AI-generated summary in https://github.com/tarasglek/chatcraft.org/commit/96f13d1db3b426e98cb250d36a00aa09c4c2376f

Issue #121, opened 2023-06-19
Shared chats should be read-only (hide version switching in UI)


tarasglek:
1. Switching between gpt-4 and chatgpt doesn't work.
2. The OpenAI key prompt is super confusing as no input box visible..Should probably not show that message in this case.

Repro(in private window):
https://chatcraft.org/c/tarasglek/X7OEaeIoke-7BEugHh6CX

<img width="813" alt="image" src="https://github.com/tarasglek/chatcraft.org/assets/857083/90295b86-bdef-4154-9879-c5c78c9b09ab">

humphd:
I've dealt with 2. in #122.

There are a few ways to deal with 1. (i.e., not being able to switch versions in a shared chat).  The underlying problem is that we rely on the database for our data reactivity; that is, if you want the Chat to change (e.g., "show me a different version"), you have to update it in the database.  When we load a shared chat, we get it via HTTP as JSON and create a Chat in memory.  This has no reactivity (i.e,. the instance never changes), so this UI won't update.

We can fix it by doing:

1. Not showing the version switcher for shared chats.  A remote chat is really "read-only" so we could embrace this and only show you the versions that the author chose.
2. Copy the remote chat into our database and load it from there, which will make it reactive.

Doing 2. would involve creating a pull-through cache in your database from the R2 bucket.  When you request `/username/id` we'd load it over JSON, put it in the database using an id of `username/id` vs. our usual `id` (to differentiate chats I own vs. ones being cached), then load it from the db.

This would complicate the db a bit, since we'd have to filter out shared chats from ones you own based on ID.  We'd also have to add a bit of code to deal with the shared chat updating (e.g., update local copy in db when loading), deleting from local db if shared chat is deleted, etc.

I'm not sure how important this feature is, but it's doable with some work.  Let me know your thoughts.

I had a good chat with ChatCraft about it, if you want to read along for more technical detail: https://chatcraft.org/c/humphd/AUpJW_Jfb6ljp5ZM2N6SB

tarasglek:
I think 1 is ok.

2 feels like overengineering

humphd:
Agreed.  I'll morph this into an issue to make remote chats truly read-only.

PR #122, opened 2023-06-20
Use custom UI for some ChatCraftAppMessages

humphd:
Replaces #66 and fixes part 2. of #121.

This adds some special `ChatCraftAppMessage` types, which are used as part of the app's logic as well as being part of a chat:

1. Greeting message
2. API Key Instructions message + form to enter and validate API Key
3. Help message (not implemented fully in this PR, but included for future)

The idea is that instead of storing a big piece of text in an App Message, we can store `app:*`, for example: `app:greeting`.  The UI can then use that to show custom text + components in a message.

The first real instance of this is adding your API Key inline in the chat:

Screenshot 2023-06-19 at 9 05 56 PM

Later I need to use the same API Key validation logic in the settings modal (I'll do that in a follow-up).  If you put in an invalid API Key, it will show an error and wait for a good one.


tarasglek:
This is perfect. 
The only problem is that now experience of editing key in options dialog feel degraded. Should implement key validation there too.

PR #123, opened 2023-06-21
Add menu to chat header, including customizing system prompt

humphd:
This changes a few things:

1. Consolidates the Share menu in the bottom right, and the Unshare and Delete buttons, into a new Chat header menu:

### Before
Screenshot 2023-06-21 at 5 09 38 PM

### After
Screenshot 2023-06-21 at 5 11 15 PM

2. Adds the ability to override the default system prompt and add a custom one.  What it does is to include a system prompt message in the chat with our default system prompt text.  You can edit it to whatever you want:

Screenshot 2023-06-21 at 5 11 35 PM

You can switch back and forth between the default and custom, if you want to reset.  Here's a simple example of what you can do.  I've switched to a custom system prompt, then added a new bullet point rule to the list of ways I want it to respond: "Answer in French"

Screenshot 2023-06-21 at 5 03 36 PM

The model does exactly what I ask.  Merci!


humphd:
OK, based on discussions with @tarasglek today, I've made some changes to how this works:

1. I've removed the greeting message, and now we use the system prompt as our greeting.
2. By default, the system prompt is collapsed so it doesn't take a ton of space; but it can be easily expanded
3. If the user clicks "Edit" they can change the prompt to whatever they want (e.g., add more rules).  An edited system prompt is always expanded, and I don't include the button or note about editing (i.e., I only show for default).
4. I've had to alter the that the "Just some me the code" feature works: it will happen on next new message vs. current
5. I've added `lib/settings.ts` to allow non-React code to access the current state of the settings
6. I've replaced the `use-system-message.ts` hook with `lib/system-prompt.ts`, which can be used in React or non-React code
7. I've added another rule to our default list to try and have ChatCraft always use the name of languages or data types in code blocks

### Default System Prompt (collapsed)

Screenshot 2023-06-22 at 4 37 35 PM

### Expanded System Prompt

Screenshot 2023-06-22 at 4 37 45 PM

My testing of this seems like it all works, but it could use some testing from you.

tarasglek:
this is fantastic. looking forward to combining this with functions

PR #127, opened 2023-06-22
Handle readonly (from JSON) vs. regular (from DB) chats/messages differently

### LINKED ISSUE #121: Shared chats should be read-only (hide version switching in UI)

humphd:
Fixes #121

This modifies both `ChatCraftChat` and all `ChatCraftMessage` types to include a `.readonly` property.  When we load a chat or message from JSON, we now mark it as `readonly=true` so that the app knows it can't be updated, changed, etc.

The only non-obvious change that this PR will make is that now, when you fork a chat, I'm going to clone the messages.  This will mean we can have duplicates in search results, since multiple messages can have the same text.  That's probably OK, and won't surprise people, but it's different than before.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -19,10 +19,9 @@ import ChatHeader from "./ChatHeader";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
-  readonly: boolean;
 };
 
-function ChatBase({ chat, readonly }: ChatBaseProps) {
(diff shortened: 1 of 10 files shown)


tarasglek:
This is good to land

PR #131, opened 2023-06-25
Add optional chat title to db, model, and UI

humphd:
Adding this based on feedback from Discord:

Screenshot 2023-06-24 at 9 35 31 PM

This adds a new optional field to all chats: `title`.  A `title` is a short name for a chat, different from the existing `summary` (i.e., a `summary` is typically generated and longer, a `title` is manually added and shorter).  The presence of a `title` on a chat gives us an interesting signal later about how much the user values this particular chat (e.g., for inclusion in search, showing in UI, etc).  I don't do anything with this yet, but we could.

It has two ways to access it.  First, in the header for a chat, you will now see a "Add Title..." button.  It's optional to add a title to a chat, and clicking it will reveal a form to enter a title:

Screenshot 2023-06-24 at 9 38 54 PM
Screenshot 2023-06-24 at 9 41 19 PM

Once a title is added to a chat, you will see it in the header, and an edit button will let you access the form again to update it:

Screenshot 2023-06-24 at 9 42 27 PM

A similar thing happens in the sidebar, where the chat's `title` is shown if it exists, otherwise we show the `summary` (i.e., so there is always some text):

Screenshot 2023-06-24 at 9 43 57 PM

Clicking the edit button in the sidebar also lets you edit the title in an inline form, similar to the header:

Screenshot 2023-06-24 at 9 44 59 PM

I've also had to do a db migration to include this new `title` field.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -16,6 +16,7 @@ import NewButton from "../components/NewButton";
 import { useSettings } from "../hooks/use-settings";
 import { useModels } from "../hooks/use-models";
 import ChatHeader from "./ChatHeader";
+import useTitle from "../hooks/use-title";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -40,6 +41,7 @@ function ChatBase({ chat }: ChatBaseProps) {
   const inputPromptRef = useRef<HTMLTextAreaElement>(null);
   const toast = useToast();
   const { user } = useUser();
+  useTitle(chat);
 
   // If we can't load models, it's a bad sign for API connectivity.
   // Show an error so the user is aware.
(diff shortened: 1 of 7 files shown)


tarasglek:
Why not make first sentence of summary serve as title? Then don't need to change schema.

Also I would put title into window title and include it in the first header before date instead of having it as a subtitle to date

humphd:
I thought about having `summary` be the only place we store this, and we could do that.  I'd still need to do a migration, though, since `summary` is currently required, and I'd need to make it optional.

The problem with putting the title/summary in the same line as the date is that it's not going to fit in many cases, especially on smaller widths.

Or are you suggesting using the Title as the main thing you see vs. the date (i.e., make the title bold, clickable, and date smaller)?

humphd:
Ok, I've made another attempt at this based on review:

- removed `title`
- made `summary` optional in schema.  If the user enters a summary, we prefer it; otherwise we generate for them
- show summary instead of date in chat heading, move date beneath and make smaller
- use summary in window title

I think this is better

Screenshot 2023-06-25 at 10 11 40 AM

PR #134, opened 2023-06-25
Update user settings modal to use api key validation

### LINKED ISSUE #130: Copy API Key verification logic to Preferences Modal
We have two different ways to enter an API Key.  Recently, the way you enter it in an App Message was extended to include a check to see if the API Key is valid.  We should re-use this code in the Preferences Modal.

Doing so will also require that the modal switch to use a button for submission vs. updating state on changes (otherwise the key will fail as you type).

humphd:
Fixes #130

This change updates the user settings modal to include API Key validation, similar to the way it works in the instructions app message.  To do it, I had to add `Cancel` + `Save` buttons:

Screenshot 2023-06-25 at 2 07 29 PM

I've also removed the drop-down with models, since we have a way to do this in the prompt form now, and I think this is less useful here.

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -1,4 +1,4 @@
-import { RefObject, useCallback, useEffect, useState } from "react";
+import { FormEvent, RefObject, useCallback, useEffect, useState } from "react";
 import { useCopyToClipboard } from "react-use";
 import {
   Button,
@@ -16,12 +16,12 @@ import {
   Radio,
   VStack,
   Stack,
-  Select,
   Kbd,
   Checkbox,
   Link,
   ButtonGroup,
   useToast,
+  FormErrorMessage,
 } from "@chakra-ui/react";
 import { exportDB } from "dexie-export-import";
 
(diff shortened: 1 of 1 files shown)


tarasglek:
Why do we need cancel and save? I'm not a fan of ui that can lose state and requires explicit save

humphd:
You need a signal that the api key is ready to be checked (i.e., previously we did it on change, but testing every character as you type is terrible UX).

tarasglek:
can we debounce onchange validation with a few seconds of delay?

humphd:
We could, but is this such a problem?  We already do a "Save" button in the other one we landed:

Screenshot 2023-06-27 at 2 20 27 PM

tarasglek:
I really don't like save buttons on dialogs :(

This one is ok as it's "i'm ready to use this", gotta click something

tarasglek:
is this still relevant? do we need a call abou this?

humphd:
Let's close and I'll do something without a "Save" button on the new prefs modal that includes the API URL stuff.

PR #136, opened 2023-06-25
Add common shortcut icons on messages when width allows

humphd:
This adds some common shortcut icons to the message heading when there is enough room.  I can't tell if this is a good idea or not, but I wanted to try it.

On smaller screen widths, I still only show the 3-dots menu.  However, if there is enough room, I also show copy, edit, and delete buttons:

Screenshot 2023-06-25 at 3 18 41 PM

The buttons only show up if a) there is enough room; b) you are hovering your mouse over that message.

This clutters things a bit, but it also makes it much, much faster to copy, edit, and delete (which I do a lot).

See what you think.


tarasglek:
We should do edit for all messages including system

Issue #140, opened 2023-06-27
Anchors in Markdown should open in a target="_blank"


humphd:
I'm working on a chat template for my colleagues to build AWS CloudFormation templates.  As part of it, I want URLs to official AWS docs included.  Here's an example:

https://chatcraft.org/c/humphd/KnVf3iLPqdmMd4ERow6F-

In this example, GPT-4 responds with a bunch of good URLs at the end, and I want to click them and have it open in a new tab vs. replacing my current chat.

We're using [react-markdown](https://www.npmjs.com/package/react-markdown) to parse and render the HTML, and it lets you [specify the syntax for various components](https://www.npmjs.com/package/react-markdown#appendix-b-components).  We need to rewrite `a` components so that they render with `<a target="_blank">...</a>`

The code change needs to happen here: https://github.com/tarasglek/chatcraft.org/blob/main/src/components/Markdown.tsx#L61

PR #142, opened 2023-06-28
Open links in new tabs in Markdown

rashboldb:
Fixes https://github.com/tarasglek/chatcraft.org/issues/140

react-markdown has a separate `linkTarget` prop for this.
Same can be achieved by modifying a component like this:
```js
a({ href, children }) {
  return (
    <a href={href} target="_blank">
      {children}
    </a>
  );
},
``` 
but above is cleaner, plus eslint is warning about breaking the [jsx-no-target-blank](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) rule in that case. Did not dive deep into this though.

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -58,6 +58,7 @@ function Markdown({ previewCode, isLoading, onPrompt, children }: MarkdownProps)
       className="message-text"
       children={children}
       remarkPlugins={[remarkGfm]}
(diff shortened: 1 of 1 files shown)


humphd (approved):
Oh, this is perfect!  I didn't see that they had this as a top-level prop you can pass.  Nice work, thank you!

PR #147, opened 2023-06-30
Make shared chats immutable, db migration v5

### LINKED ISSUE #143: Decide on update strategy for shared chats
I want to explore how shared chats should work, specifically when it comes to updates.  Let me describe how this works at the moment.

You create a chat and decide to share it.   Let's say the chat is https://chatcraft.org/c/KlRWLA8H85RXSrmdgyEIh locally.  You get a public URL, let's say https://chatcraft.org/c/humphd/KlRWLA8H85RXSrmdgyEIh (note the addition of the username `/humphd/` in the path).

When you (as the creator) go to https://chatcraft.org/c/humphd/KlRWLA8H85RXSrmdgyEIh, it will redirect you to https://chatcraft.org/c/KlRWLA8H85RXSrmdgyEIh, that is, you'll be going back to the chat you shared.  If you make a change, it will update the db AND also push that newly updated version to R2 as well.

This has confused @tarasglek, and even I got surprised by it today.  I wonder what the correct way to do this would be?

In the past, we said that we didn't want to have buttons that you had to click to "save" things, that everything should be automatic.  I extended this idea to shared chats, and if you make a change, it auto-updates them.

Should a shared chat be immutable?  Should you have to manually opt into updating it?  Should updates be automatic?

humphd:
Fixes #143

This changes our strategy and data schema for shared chats.  Previously, we included an optional `sharedUrl` property on a chat, indicating that it was shared.  This URL was stable even if the chat changed (e.g. you added/deleted messages), since we silently kept the data in R2 in sync with changes in the local database.

However, we're moving to having shared chats be immutable.  Once you share a chat, you can delete it but not change it.  To help with reactivity in the app, I've created a new table to store the shared chat info locally (I played with various approaches where I queried the API functions, but it got overly complicated).  Our IndexedDB database is our source of truth, so shared chats also get managed there.  This makes it easy to show them in the UI.

To accomplish this, I've removed all uses of `shareUrl`, which meant doing another db migration.  This is the first time that I had to write an `upgrade()` function to convert the data to use the new `shared` table (i.e., I take the chats with `shareUrl` values and copy their data over into the new table before deleting the property).  It's really slick how Dixie allows these upgrades in place.

I've also created a new type, `SharedChatCraftChat`, which you get from calling `.share()` on a `ChatCraftChat` object.  It represents the data as we have it in R2 and the `shared` table.

To make testing easier, I've also added an `Import` feature for the database.  Now you can open `chatcraft.org` and go to Settings, click `Export`, get the `.json` file, then go back to a preview URL or localhost and click `Import` to bring that database into the current app.  Then you can test what the migration will do.  I was able to migrate all my data over to localhost without issue.

I will need to rebase this and test some more, but I think all the necessary pieces are working well.  Hopefully this is the last time I have to rewrite sharing!

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -119,7 +119,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
       try {
         // Add this prompt message to the chat
-        await chat.addMessage(new ChatCraftHumanMessage({ text: prompt, user }), user);
+        await chat.addMessage(new ChatCraftHumanMessage({ text: prompt, user }));
 
         // In single-message-mode, trim messages to last few. Otherwise send all.
         // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
@@ -128,7 +128,7 @@ function ChatBase({ chat }: ChatBaseProps) {
         const response = await callChatApi(messagesToSend);
 
         // Add this response message to the chat
-        await chat.addMessage(response, user);
+        await chat.addMessage(response);
       } catch (err: any) {
         toast({
           title: `OpenAI Response Error`,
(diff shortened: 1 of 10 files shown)


humphd:
Rebased

humphd:
Upgraded my db on prod without issue.  Dexie ftw!

PR #148, opened 2023-07-02
Use tiktoken WASM to count message tokens

humphd:
I'm not sure if this is the right way to do this or not, but I'm trying an experiment using [tiktoken](https://github.com/dqbd/tiktoken).  It's a JS/WASM port of OpenAI's tiktoken, allowing you to get tokens for a string.  It's crazy fast (unlike langchain's equivalent).  The only downside is the extra download size.

If you enable token counting in the settings, this is what it looks like (not sure that I like the prominence of the token count, but this is a draft to discuss):

Screenshot 2023-07-01 at 9 28 31 PM

I've also added dynamic getters to `ChatCraftMessage` and `ChatCraftChat` to easily get the counts.  This would allow me to bypass the bugs with langchain and token counting in streaming messages, and just count everything myself.

I know you wanted to have a "receipt" of token usage.  I'm wondering about putting it in `sessionStorage` so it's tied to your current session/window.  I could connect the code in this PR to the calls to OpenAI, and keep track of things as you chat.

See what you think, and we can discuss/iterate.


tarasglek:
this feels odd. It's kind of a superpower to have this info, but a superpower you don't normally want...could probably have this as optional in settings. It's an interesting feature for devs.

Also doesn't load here.
image

tarasglek:
i'm also concerned about footprint/loadtimes/memory-fitprint..tiktoken is pretty heavyweight

humphd:
Updated to make it only load tiktoken module dynamically when setting is enabled.

I have no idea what this crash is on startup, I'll have to look into that next.

humphd:
OK, preview URL works now https://tokens.console-overthinker-dev.pages.dev/

humphd:
I have this working now:

Screenshot 2023-07-04 at 2 32 25 PM

If the user opts into token/cost tracking via settings, we keep track of the cost for a session cumulatively.

The large dependencies to make this work are only loaded if the setting is enabled at runtime vs. startup.

I'm still not sure if we want this or not, but it works.

tarasglek:
I like it

humphd:
Rebased.  Should I land this, then?

PR #150, opened 2023-07-04
Limit shares to 500 total, 10 in 24 hours

humphd:
Previously we discussed adding some rate limits to the sharing API, but then I forgot to do it.

I've added two more checks:

1. users can't share more than 500 chats total
2. users can't share more than 10 chats in 24 hours

Both of these numbers are arbitrary and could be changed and/or made into env vars if you want.

As always, I can't fully test this without doing it on prod, but I think it's right.  See the [R2 Workers API reference](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#bucket-method-definitions) for details.

### DIFF
--- a/functions/api/share/[[user_id]].ts
+++ b/functions/api/share/[[user_id]].ts
@@ -34,17 +34,38 @@ export const onRequestPut: PagesFunction<Env> = async ({ request, env, params })
     return errorResponse(400, "Expected share URL of the form /api/share/{user}/{id}");
   }
 
-  // Make sure we have a token, and that it matches the expected user
+  // Make sure the access token matches the expected user and limits.
   try {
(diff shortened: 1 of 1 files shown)


tarasglek:
Interesting approach. If we change Id to be yyyy.mm.dd.h.m.s..we can then arbitrarily refine the throttle limit..eg this is scheme limits to one share per second...to do a further limit could scan bucket with a prefix to see how much was shared within a time bucket

humphd:
We can already do that without changing `id`, since we have a separate set of metadata in the [`R2Object`](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2object-definition) we get back from the bucket, and all objects have `uploaded` date stamped into them.

How should I proceed with this?

PR #153, opened 2023-07-09
duplicate chat until message

tarasglek:
I keep wanting to fork my chats to take them in different directions(kind of like threads in other chat apps).

We had a `Duplicate Chat from Message`, which nuked the system message and included all chats after message. It also did that inplace and looked like it lost my very custom system message.

This changes that to 'Duplicate Chat until Message' which now opens a new window and includes all messages up to and including message-being-duplicated


humphd (changes requested):
Remove line 82 to make eslint happy (we don't use `navigate` now).

Looks good!

PR #154, opened 2023-07-09
Allow code within "Run Code" to decide how output gets rendered

tarasglek:
This lets js control what gets rendered. In particular this will enable clickhouse markdown.

image


humphd:
Can you rebase on `main` so you don't have these other commits in here?

humphd:
This looks good other than commit cleanup.

Issue #105, opened 2023-06-13
Keep track of scroll position when retrying AI messages


humphd:
In #99 we're adding the ability to stream a new message into the middle of a chat.  We have logic for auto-scrolling when adding a new message to the bottom, but we don't do this for retries.

It would be nice to refactor our auto-scroll behaviour so it works for any case where we stream content.

humphd:
The code related to this changed in https://github.com/tarasglek/chatcraft.org/pull/172#issuecomment-1654792215:

> I've also made a change to how the auto-scrolling works, and I'll need to build on it in follow-ups. Previously, I only auto-scrolled when a streaming messages was coming in (and changed its value). This breaks when we add new messages to the chat, so was broken for functions. Fixing this is good for the function case, but it will create some weirdness when you delete a message high-up in the middle of a chat, or create new versions of a chat, etc. I'll fix that later, but if it bugs people, we can disable the dependency on chat.date in the autoscroll effect.

We need to have the ability to set a React Ref to an arbitrary element to use as the bottom vs. the bottom of the scroll div.

PR #155, opened 2023-07-10
Have all AI chat calls use same code path and UX

### LINKED ISSUE #152: Have all AI chat calls use same code path and UX (i.e., retires like initial prompt)
Forgot to file this earlier. Got confused as to why esc wasn't working and remembered

humphd:
Fixes #152.

We currently have 3 different places where we interact with OpenAI:

1. When you ask a question in the prompt form
2. When we retry an AI message with a different model inline
3. When a user requests that we summarize a chat in the share modal

These all worked a bit differently, so we had 3 different code paths. However, the one we use for the prompt form was the most mature, and included nice UX features like pause/resume and cancel.

I've amalgamated all three of this into a single one.  The low-level stuff now lives in `src/lib/ai.ts` and the UI components use it via `src/hooks/use-chat-openai.ts`.

Some callers need to be able to stream responses, some don't, so I've made something that you can pass event handlers to, but you can also use it via a promise if all you care about is the result.

I've also reworked the code to have both the prompt form and inline retires use the same UI components.  As part of this I fixed some visual bugs, and one downside was that I had to remove the click-to-pause feature.  I think now that we have an actual pause button, this is less important, but let me know if you disagree and I can find a way to re-enable it.

One thing I haven't done yet is fix the auto-scrolling for retry messages, which is #105, and I'll tackle that separately.

I've also fixed a stray import in `src/components/Message/MessageBase.tsx` we missed earlier today.

This needs a bunch of testing.  Try to look past UI bugs that aren't related to what I've done here (I know that's hard), but I just need to get this new OpenAI logic landed and then I can work on fixing things in common code.

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -27,7 +27,7 @@ import ResizeTextarea from "react-textarea-autosize";
 import { TbDots, TbTrash } from "react-icons/tb";
 import { AiOutlineEdit } from "react-icons/ai";
 import { MdContentCopy } from "react-icons/md";
-import { Link as ReactRouterLink, useNavigate } from "react-router-dom";
+import { Link as ReactRouterLink } from "react-router-dom";
 
 import { formatDate, download, formatNumber } from "../../lib/utils";
 import Markdown from "../Markdown";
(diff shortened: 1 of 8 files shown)


tarasglek:
haven't run into any bugs, appreciate the improvements. Time to finish testing in prod.

PR #161, opened 2023-07-12
openrouter & other openai-compat model support

tarasglek:
Openrouter is neat in that it in theory gives us CC access and claude and open source models. They also support old-fashioned baseURL/token workflow, so easy to start to experiment with.

### 1.  Openrouter also does not support model listing
Ignore this, that was my bug.

### 2. Openrouter streaming does not work with langchain JS:
```
TypeError: Cannot read properties of undefined (reading 'function_call')
    at onmessage (openai.js:371:52)
    at onLine (event-source-parse.js:128:13)
    at onChunk (event-source-parse.js:92:13)
    at getBytes (event-source-parse.js:28:9)
    at async getResponse (axios-fetch-adapter.js:238:13)
    at async fetchAdapter (axios-fetch-adapter.js:176:18)
    at async RetryOperation._fn (index.js:50:12)
```
for some reason delta is null in their responses(I can't get browser devtools to show me SSE to debug further)
Here is unhappy code in langchain. Pretty sure every langchainjs streaming app is broken with them.
image
So when using non-openai models so I disable streaming for non-openai baseurls for now. We could probably do a smooth fallback for when streaming fails, but completion still comes in.

### 3. Our parsing of / in model names is a complete mess

Disabled it for now.

### 4. Our state management around preferences modal is pretty bad
Need to reload after adding baseurl

Proof that stuff works:
image

Feel free to try this with my key:
```
key: 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
baseurl: https://openrouter.ai/api/v1
```

@humphd I would like to land something like this. I think it's a good first step towards 3rd-party models. I just need @alexanderatallah to fix model-listing as hardcoding models is a no-go.

Issue #156, opened 2023-07-10
Drop langchainjs in favour of official openai


humphd:
I was updating deps tonight and ran into a wall when I failed a build due to [deprecations and renaming in langchainjs v0.0.105](https://github.com/hwchase17/langchainjs/commit/90d7edebe91b27a75229c465013e72bd32394598).

It's not a huge deal, and I can do it, but this is probably a good time to ask: are we ever planning to use langchain?  Right now we don't.  We could in various places, which I've asked about in https://github.com/tarasglek/chatcraft.org/issues/119#issuecomment-1608536293.

@tarasglek can we put some thought into this?  If we aren't ever going to use it, I'll rip it out.  If we are, I'll do this refactor.

humphd:
https://github.com/lgrammel/ai-utils.js is interesting, as is https://vercel.com/blog/introducing-the-vercel-ai-sdk.

tarasglek:
I endorse ditching it for anything with decent functions support.

humphd:
Reading the code for these, I don't think either is perfect for what we need.

I think what I'm going to do is isolate our code even further from Langchain in order to prepare for a time in the future when we replace it with something different.

For now I'll stick with it.

tarasglek:
langchain both hurt us in adoption of openrouter (eg broke us where we couldn't do workarounds) and made it easier to argue for them to fix support for it :)

I think it's either official openai api or langchain. I would prefer official openai api.

humphd:
Official OpenAI node module is not really aimed at browsers, see https://github.com/openai/openai-node#openai-nodejs-library, though work seems to be happening in v4 https://github.com/openai/openai-node/issues/102#issuecomment-1627892242.

I'd like to move forward with my changes in #165, can you review that?  Regardless of what we do, those changes will help isolate us from how the AI pieces work, and make switching easier.

tarasglek:
Lets switch to raw openai. Langchain/openrouter compat is too painful. I want to be able to say that we support llama, but can't atm. it's too glitchy

humphd:
Adding some more context to this, in case it's helpful. We largely isolate our use of langchain to a few spots (mostly in anticipation of this issue being solved).  See:

https://github.com/search?q=repo%3Atarasglek%2Fchatcraft.org%20from%20%22langchain%2F&type=code

[...]

PR #162, opened 2023-07-12
Revert "Have all AI chat calls use same code path and UX"

tarasglek:
Reverts tarasglek/chatcraft.org#155

This caused
```
> chatcraft.org@0.0.0 build /Users/taras/Documents/chatcraft.org
> tsc && vite build

src/components/Message/MessageBase.tsx:157:17 - error TS2540: Cannot assign to 'text' because it is a read-only property.

157         message.text = text;
                    ~~~~

src/components/Message/NewMessage.tsx:17:8 - error TS2739: Type '{ key: string; message: ChatCraftAiMessage; chatId: string; isLoading: true; heading: string; }' is missing the following properties from type 'Omit<MessageBaseProps, "message" | "avatar">': editing, onEditingChange

17       <OpenAiMessage
          ~~~~~~~~~~~~~

src/lib/ChatCraftMessage.ts:194:10 - error TS2540: Cannot assign to 'text' because it is a read-only property.

194     this.text = text;
             ~~~~


Found 3 errors in 3 files.

Errors  Files
     1  src/components/Message/MessageBase.tsx:157
     1  src/components/Message/NewMessage.tsx:17
     1  src/lib/ChatCraftMessage.ts:194
 ELIFECYCLE  Command failed with exit code 2.
 ```

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -27,7 +27,7 @@ import ResizeTextarea from "react-textarea-autosize";
 import { TbDots, TbTrash } from "react-icons/tb";
 import { AiOutlineEdit } from "react-icons/ai";
 import { MdContentCopy } from "react-icons/md";
-import { Link as ReactRouterLink } from "react-router-dom";
+import { Link as ReactRouterLink, useNavigate } from "react-router-dom";
 
 import { formatDate, download, formatNumber } from "../../lib/utils";
 import Markdown from "../Markdown";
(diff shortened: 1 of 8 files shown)


humphd:
This error isn't from that change.  You're hitting the regression I mentioned in #156 with [your update to langchainjs v0.0.v105](https://github.com/tarasglek/chatcraft.org/pull/161/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R27), which includes them reworking a bunch of stuff.

I'm going to isolate our code from langchain's so we don't break when they update.

PR #166, opened 2023-07-13
Model validation in prefs modal, better model/provider handling

humphd:
This updates our Preferences modal to properly handle changes to the API Key.  It uses no "save" button (you're welcome, @tarasglek), and debounces changes on the input instead.  One issue I ran into is that the models API endpoint for OpenRouter.ai does NOT require an api key (i.e., it always works, so this won't tell us if your key is working or not).

I've also fixed the `useModels()` hook to update whenever the `apiKey` or `apiUrl` are changed.

Doing this work, I've realized that OpenAI and OpenRouter have different model names: OpenRouter includes a vendor `openai/*` and OpenAI does not.  This means switching providers is a bit hairy, since you have to also change the default model, which I've done.

We could do more work to separate how we store the model (without vendor) and how we provide it for the model, depending on the provider.  But I think this is a big edge case we can fix if/when necessary.

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -1,5 +1,5 @@
 import { ChangeEvent, RefObject, useCallback, useEffect, useRef, useState } from "react";
-import { useCopyToClipboard } from "react-use";
+import { useCopyToClipboard, useDebounce } from "react-use";
 import {
   Button,
   FormControl,
@@ -23,6 +23,7 @@ import {
   Link,
   ButtonGroup,
   useToast,
+  FormErrorMessage,
 } from "@chakra-ui/react";
 import { exportDB, importDB } from "dexie-export-import";
 
@@ -33,6 +34,7 @@ import db from "../lib/db";
 import { useModels } from "../hooks/use-models";
 import { ChatCraftModel } from "../lib/ChatCraftModel";
 import { OPENAI_API_URL, OPENROUTER_API_URL } from "../lib/settings";
+import { validateOpenAiApiKey } from "../lib/ai";
 
 // https://dexie.org/docs/StorageManager
 async function isStoragePersisted() {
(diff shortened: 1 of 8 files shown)


tarasglek:
2 thoughts:
1. it makes sense to use all providers in parallel instead of switching. 
2. Because right now we switch, we now send openai key to openrouter or vise-versa for validation. that's not good. At the very minimum we should use separate key field for each key...in ui we can just hide openai key field when openrouter is chosen, etc
3. I'll still take this as that's a rare corner case and otherwise chatcraft is broken after my last change

humphd:
> 1. it makes sense to use all providers in parallel instead of switching.

What does this mean, I don't understand?  We have to know where we are sending your request.

> 2. Because right now we switch, we now send openai key to openrouter or vise-versa for validation. that's not good. At the very minimum we should use separate key field for each key...in ui we can just hide openai key field when openrouter is chosen, etc

Easy fix for this is to clear the API Key when switching providers.

Another thought: should the OpenRouter.ai flow always use OAuth instead of manual API Key entry?  That is, does an OpenRouter.ai user ever want to enter this themselves, or is it better that we use their [OAuth PKCE](https://openrouter.ai/docs#oauth) method to get the API Key for them?

humphd:
RE: OAuth PKCE for OpenRouter...

If we switched to an "Authenticate with OpenRouter.ai" button, and got the API Key via OAuth flow, we could eliminate the need to validate that key.  I wonder if this is the answer.

So in the forms where we have provider drop-down, if you choose OpenAI we show an input for manual API Key entry; and if you choose OpenRouter.ai, we show a button to authenticate and get key from them.

tarasglek:
yes i think OAuth is the fix for this.

>>it makes sense to use all providers in parallel instead of switching.
> What does this mean, I don't understand? We have to know where we are sending your request.

I meant might make sense to use my openai credits via openai account, and still have access to "retry with claude-v1"

Issue #159, opened 2023-07-10
Mermaid diagrams can randomly fail to render then start working


humphd:
From Discord, there was discussion of the fact that a correct Mermaid diagram can fail to render, then suddenly start working.  I've seen this a few times as well, and I suspect that it has something to do with us pushing a new version to prod, and the local cached version of Mermaid.js being different (i.e., it can't load the dependency for some reason).

A refresh of the page will usually fix it (which is why it "suddenly" starts working for "no reason").  We might be able to have updates go smoother.

humphd:
Our service worker (see `build/sw.js`) does this:

```
    s.precacheAndRoute(
      [
        { url: "android-chrome-192x192.png", revision: "9338ee34f2a337c375cf60b0cdb5b3c6" },
        { url: "android-chrome-512x512.png", revision: "05fed013f47b2b60a3ddf91dc97456ec" },
        { url: "apple-touch-icon.png", revision: "01a521e3bd9203e755ddc4fd77e3ac27" },
        { url: "favicon-16x16.png", revision: "dcc8379b51266819244f1a117ce3c5bc" },
        { url: "favicon-32x32.png", revision: "d92930e5701f88aec4bd7dd13fc52f85" },
        { url: "favicon.ico", revision: "36ef512c6eb3b474f57b356ec46b26c5" },
        { url: "openai-logo.png", revision: "de0068828c32f2be9a0ea36b1196d580" },
        { url: "assets/arc-bd7316d5.js", revision: null },
        { url: "assets/array-9f3ba611.js", revision: null },
        { url: "assets/c4Diagram-9cddb37f-8fbadd15.js", revision: null },
        { url: "assets/classDiagram-bc733c3b-d2b53d56.js", revision: null },
        { url: "assets/classDiagram-v2-8931bdaf-803d3af2.js", revision: null },
        { url: "assets/core-b8b80318.js", revision: null },
        { url: "assets/createText-3df630b5-9c82b1dc.js", revision: null },
        { url: "assets/css-c2652ddb.js", revision: null },
        { url: "assets/edges-49ac43a2-a1b7f1ab.js", revision: null },
        { url: "assets/erDiagram-f6946109-da96a775.js", revision: null },
        { url: "assets/flowchart-elk-definition-5082a990-cf1b168e.js", revision: null },
        { url: "assets/flowDb-6a57c1b4-21e3e0ee.js", revision: null },
[...]

PR #168, opened 2023-07-14
Update deps for July 14, 2023

humphd:
This updates all of our deps except vite, which fails on macOS with a missing esbuild binary:

```
error Couldn't find package "@esbuild/darwin-x64@0.18.12" required by "esbuild@^0.18.10" on the "npm" registry.
```

NOTE: this change updates a number of things in our dev dependencies to new major versions.  Everything seems OK, but we should keep our eyes open for problems.

Also, this includes an update to Mermaid.js, which might help us trigger #159 when we land on prod.  If/when people hit this, I want to know what it's trying to load vs. what's on the server.


humphd:
I can't recreate these failures locally, the must be network/cache issues on npmjs.  I'll wait a bit and re-trigger.

humphd:
Looks like it's known: https://twitter.com/npmstatus/status/1679957928294649858

humphd:
@tarasglek, @rashboldb when this gets merged, we might be able to trigger #159 and have Mermaid fail.  If you don't mind, I'd suggest that you do the following:

1. open a chat with a Mermaid diagram in it, confirm it works
2. open dev tools > Network
3. refresh to get latest from prod
4. see if diagrams fail, and if so, what the network is showing (e.g., 404s)

humphd:
This worked for me (no bug).

PR #171, opened 2023-07-19
replit support

tarasglek:
Experimenting with bounties there

https://replit.com/@tarasglek1/chatcraftorg

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,29 @@
+run = "vite --host 0.0.0.0"
+hidden = [".build", ".config", "package-lock.json", "tsconfig.json"]
+
+[packager]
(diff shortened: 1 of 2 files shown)


humphd:
What is this one about, being able to open it on Repl.it?

tarasglek:
yes, just lets one edit on https://replit.com/. However their 500mb RAM is insufficient to bundle react with vite 🤷

humphd:
We could add GitPod / Code Spaces too if you want that style of dev env.

tarasglek:
I'm just experimenting, replit does a lot of PR, but I'm not impressed so far

Issue #86, opened 2023-06-02
Let's try to define and prototype tools


humphd:
We've spoken a lot about "tools," and the abilities they'd unlock.  Now that we have the message types in place, routing, and the db backend, I think we have most of the pieces necessary to start playing around with this.

It likely makes sense to consider this in the context of Langchain, which already gives us access to [Agents](https://js.langchain.com/docs/modules/agents/) and [Tools](https://js.langchain.com/docs/modules/agents/tools/).  Here, a "tool" means:

```ts
interface Tool {
  call(arg: string): Promise<string>;
  name: string;
  description: string;
}
```

So we could begin by providing a way to write/upload/store JS (or TS and transpile, or I guess whatever lang and compile to WASM) which exports a `call(arg: string): Promise<string>` function.

We now have `/c/*` for chats.  What about `/t/*` for tools?

How do we use these tools?  Do we create a [Toolkit](https://js.langchain.com/docs/modules/agents/toolkits/) for an agent to use, based on all the `/t/*` tools owned by the user?  Or do we have some way to insert a Tool as context in a message?  Or both?

humphd:
[OpenAI announced](https://openai.com/blog/function-calling-and-other-api-updates) changes to how models interact with functions:

> Function calling
>
> Developers can now describe functions to gpt-4-0613 and gpt-3.5-turbo-0613, and have the model intelligently choose to output a JSON object containing arguments to call those functions. This is a new way to more reliably connect GPT's capabilities with external tools and APIs.
>
>These models have been fine-tuned to both detect when a function needs to be called (depending on the user’s input) and to respond with JSON that adheres to the function signature. Function calling allows developers to more reliably get structured data back from the model. For example, developers can:
>
>Create chatbots that answer questions by calling external tools (e.g., like ChatGPT Plugins)
>
Convert queries such as “Email Anya to see if she wants to get coffee next Friday” to a function call like send_email(to: string, body: string), or “What’s the weather like in Boston?” to get_current_weather(location: string, unit: 'celsius' | 'fahrenheit').
>
>Convert natural language into API calls or database queries
>
>Convert “Who are my top ten customers this month?” to an internal API call such as get_customers_by_revenue(start_date: string, end_date: string, limit: int), or “How many orders did Acme, Inc. place last month?” to a SQL query using sql_query(query: string).
>
>Extract structured data from text
>
[...]

PR #172, opened 2023-07-19
Initial work to add functions UI and db support

humphd:
This is the start of work on adding functions to ChatCraft.  I'm initially targeting the database and UI, so we can create/manage functions.  Using them comes after this.

This PR adds a new set of "function" routes to the front-end, similar to what we have for Chats with `/c/*`: 

- `/f/new` creates a new function
- `/f/:id` loads an existing function
- `/f/:id/delete` deletes an existing function

Screenshot 2023-07-19 at 9 02 47 AM

I've using CodeMIrror 6 as the editor component.

As discussed in #86, I've done this first version by having all of the metadata get extracted from the code.  I'm not sure that what's required to make this happen is efficient, but it's working (i.e., the code has to be parsed and saved on every keystroke in the editor).  My computer is really fast, so it would be good to test this on something slower.

The alternative to what I'm doing here would be:

- add a "Save" button to only save when you want (I already know what you'll say to this...)
- debounce keystrokes, which means editor state and db are not totally in sync
- move metadata out of code so it can be managed separately

However, maybe what I have here is fine.  I'm able to parse the JS code into a ES Module and get the default function.  If we wanted to call it for an LLM, that would be easy.

NOTE: I haven't bothered to add this to the Sidebar yet, since I want to figure out the db and editor part first, in case the data layout changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,10 @@
   },
   "dependencies": {
     "@chakra-ui/react": "^2.7.1",
+    "@codemirror/lang-javascript": "^6.1.9",
     "@emotion/react": "^11.11.1",
     "@emotion/styled": "^11.11.0",
+    "@uiw/react-codemirror": "^4.21.7",
     "cookie": "^0.5.0",
     "dexie": "^3.2.4",
     "dexie-export-import": "^4.0.7",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "dexie-react-hooks": "^1.1.6",
     "framer-motion": "^10.12.21",
     "jose": "^4.14.4",
-    "langchain": "^0.0.109",
+    "langchain": "^0.0.114",
     "mermaid": "^10.2.4",
     "nanoid": "^4.0.2",
     "react": "^18.2.0",
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -8,13 +8,14 @@ import MessagesView from "../components/MessagesView";
 import Header from "../components/Header";
 import Sidebar from "../components/Sidebar";
 import useChatOpenAI from "../hooks/use-chat-openai";
-import { ChatCraftHumanMessage } from "../lib/ChatCraftMessage";
+import { ChatCraftFunctionCallMessage, ChatCraftHumanMessage } from "../lib/ChatCraftMessage";
 import { ChatCraftChat } from "../lib/ChatCraftChat";
 import { useUser } from "../hooks/use-user";
 import NewButton from "../components/NewButton";
 import { useSettings } from "../hooks/use-settings";
 import { useModels } from "../hooks/use-models";
 import ChatHeader from "./ChatHeader";
+import { ChatCraftFunction } from "../lib/ChatCraftFunction";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -61,13 +62,14 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Auto scroll chat to bottom, but only if user isn't trying to scroll manually
   // Also add a dependency on the streamingMessage, since its content (and therefore
-  // the height of messageList) will change while streaming.
+  // the height of messageList) will change while streaming, and the chat's date
+  // since we update that whenever the chat is re-saved to the db
   useEffect(() => {
-    if (messageListRef.current && shouldAutoScroll && streamingMessage) {
+    if (messageListRef.current && shouldAutoScroll) {
       const messageList = messageListRef.current;
       messageList.scrollTop = messageList.scrollHeight;
     }
-  }, [messageListRef, streamingMessage, shouldAutoScroll]);
+  }, [messageListRef, streamingMessage, shouldAutoScroll, chat.date]);
 
   // Disable auto scroll when we're loading and the user scrolls up to read previous content
   const handleScroll = useCallback(() => {
@@ -132,9 +134,28 @@ function ChatBase({ chat }: ChatBaseProps) {
 
         // Add this response message to the chat
         await chat.addMessage(response);
+
+        // If it's a function call message, invoke the function
+        if (response instanceof ChatCraftFunctionCallMessage) {
+          const func = await ChatCraftFunction.find(response.func.id);
+          if (!func) {
+            toast({
+              title: `Function Error`,
+              description: `No such function: ${response.func.name} (${response.func.id}`,
+              status: "error",
+              position: "top",
+              isClosable: true,
+            });
+            return;
+          }
+
+          const result = await func.invoke(response.func.params);
+          // Add this result message to the chat
+          await chat.addMessage(result);
+        }
       } catch (err: any) {
         toast({
-          title: `OpenAI Response Error`,
+          title: `Response Error`,
           description: "message" in err ? err.message : undefined,
           status: "error",
           position: "top",
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -15,8 +15,7 @@ import NewButton from "../components/NewButton";
 import { useSettings } from "../hooks/use-settings";
 import { useModels } from "../hooks/use-models";
 import ChatHeader from "./ChatHeader";
-import db from "../lib/db";
-import { ChatCraftFunction } from "../lib/ChatCraftFunction";
+import { functionsFromPrompt } from "../lib/ChatCraftFunction";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -125,12 +124,11 @@ function ChatBase({ chat }: ChatBaseProps) {
         const messages = chat.messages({ includeAppMessages: false, includeSystemMessages: true });
         const messagesToSend = singleMessageMode ? [...messages].slice(-2) : messages;
 
-        // HACK: pass our functions through
-        const records = await db.functions.toArray();
-        const funcs = await Promise.all(records.map((record) => ChatCraftFunction.fromDB(record)));
-        console.log({ funcs });
-
-        const response = await callChatApi(messagesToSend, undefined, funcs);
+        // See if the user wants us to use any functions
+        const functions = await functionsFromPrompt(prompt);
+        const response = await callChatApi(messagesToSend, {
+          functions,
+        });
 
         // Add this response message to the chat
         await chat.addMessage(response);
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -8,14 +8,13 @@ import MessagesView from "../components/MessagesView";
 import Header from "../components/Header";
 import Sidebar from "../components/Sidebar";
 import useChatOpenAI from "../hooks/use-chat-openai";
-import { ChatCraftHumanMessage, ChatCraftSystemMessage } from "../lib/ChatCraftMessage";
+import { ChatCraftHumanMessage } from "../lib/ChatCraftMessage";
 import { ChatCraftChat } from "../lib/ChatCraftChat";
 import { useUser } from "../hooks/use-user";
 import NewButton from "../components/NewButton";
 import { useSettings } from "../hooks/use-settings";
 import { useModels } from "../hooks/use-models";
 import ChatHeader from "./ChatHeader";
-import { functionNamesFromChat, loadFunctions } from "../lib/ChatCraftFunction";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -124,13 +123,9 @@ function ChatBase({ chat }: ChatBaseProps) {
         const messages = chat.messages({ includeAppMessages: false, includeSystemMessages: true });
         const messagesToSend = singleMessageMode ? [...messages].slice(-2) : messages;
 
-        // Scan entire msg history for functions so we can use more natural interaction form
-        // Seems more clear to do this directly
-        const humanAndSystemMessages = messagesToSend.filter(
-          (x) => x instanceof ChatCraftHumanMessage || x instanceof ChatCraftSystemMessage
-        );
-        // See if the user wants us to use any functions
-        const functions = await loadFunctions(functionNamesFromChat(humanAndSystemMessages));
+        // If there are any functions mentioned in the chat (via @fn or @fn-url),
+        // pass those through to the LLM to use if necessary
+        const functions = await chat.functions();
         const response = await callChatApi(messagesToSend, {
           functions,
         });
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -8,14 +8,14 @@ import MessagesView from "../components/MessagesView";
 import Header from "../components/Header";
 import Sidebar from "../components/Sidebar";
 import useChatOpenAI from "../hooks/use-chat-openai";
-import { ChatCraftHumanMessage } from "../lib/ChatCraftMessage";
+import { ChatCraftHumanMessage, ChatCraftSystemMessage } from "../lib/ChatCraftMessage";
 import { ChatCraftChat } from "../lib/ChatCraftChat";
 import { useUser } from "../hooks/use-user";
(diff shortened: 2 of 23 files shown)


tarasglek:
This approach seems logical, feel free to land it.

humphd:
Added Functions to the sidebar so it's easier to navigate back and forth between chats and functions:

Screenshot 2023-07-20 at 6 59 24 PM

I suspect that this will change once we get code that can call these, but we need something to start with, so here it is.

NOTE: I've added "New" buttons in the sidebar beside "Recent Chats" and "Functions" so that you can easily get from one to the other.  Probably we'll need more links/UI.

humphd:
@tarasglek, check this out.  Here's a function add an arbitrary list of numbers and format it:

Screenshot 2023-07-21 at 9 43 29 AM

Then I use it to do math:

Screenshot 2023-07-21 at 9 43 15 AM

The demo above passes all of your functions, which is obviously wrong, but let's me test it.  It then gets the response back, finds the function to call in the DB, invokes it with the arguments, and returns the result.

Lots of hacks and rough edges here, but I'm excited!  I need to clean up the code so it's generalized, and then find a way to tell the model to use particular functions in UI, but IT WORKS!

humphd:
I pushed my hacky code, if you want to play along: don't land this.

To use it: create one or more functions, then write a prompt that can use them.  It will print the result if it can.

humphd:
Added some basic UI to the message so it's clear that a function was called, which args were passed, and what the result was:

Screenshot 2023-07-21 at 1 19 07 PM

humphd:
@tarasglek I've got this to the point that you can use it.  Here's how it works:

1. Create one or more functions (i.e., go to `/f/new`).  Here's an example:

```js
/**
 * Example Function Module. Each function needs you to define 4 things:
 */

/* 1. Name of your function */
export const name = "sum";

/* 2. Description of function, used to describe what it does to an LLM */
export const description = "Adds all numbers passed to it, returning the total.";

/**
* 3. A JSON Schema defining the function's parameters. See:
*
* - https://platform.openai.com/docs/guides/gpt/function-calling
* - https://json-schema.org/learn/getting-started-step-by-step
*/
export const parameters = {
  type: "object",
  properties: {
    "numbers": {
      items: {
        type: "number"
      },
      description: "Array of numbers",
    }
  }
};

/**
 * 4. The function itself. Accepts an Object matching the schema
 * defined in params, returning a Promise<string> (i.e., should be
 * an async function).
 */
export default async function (data) {
  const { numbers } = data;
  const result = numbers.reduce((acc, current) => acc + current, 0);
  return result.toLocaleString();
}
```

2. Write a prompt and let ChatCraft know you want to use your function.  For example: `Add these numbers: 10, 1000, 13, 134, 1345 @function: sum`. I look for `@functions: ...`, `@function: ...`, `@fns: ...`, and `@fn: ...` (the prefix/format is something to figure out, but for now it works) and then extract the list of one or more function names.
3. If you include function(s) in your prompt, the LLM will use them, otherwise it's going to do the usual thing.

I struggled a lot to figure out the right UI to include functions easily in your chat, and it suddenly dawned on me that using plain text was probably the easiest.  It's not so discoverable, so we might want visual UI as well later.

Try this branch.  It lets us experiment with the ideas.

tarasglek:
not having any luck..i defined a function and it's not passed into api

image

humphd:
What is your prompt?

humphd:
i.e., did you include "@function: runClickhouseQuery` to tell it to use that function?

tarasglek:
Ok, sorry I was really confused by the UX. I got it now. It works!

tarasglek:
image

Here is a small refinement of your work. I made it scan all human + system message for functions. this allows a more natural convo form.
I also allowed functions to provide own formatting.

1. I find the experience of having functions listed in sidebar to be good..nice to be able to switch between function and chat in sidebar
2. I find using chatcraft to edit functions frustrating. I propose we just list urls to gists or other CORS-fetch()able content..then just add a "reload functions" button(if functions are present) for handling the corner case of gists changing
3. I think in a followup we can get rid of special syntax for listing functions..we can just run a special chatgpt call every time system message changes, that extract all functions using some gpt parsing-using-functions trick..this would also validate the json schema for loaded functions..and could even pass those through another gpt prompt to suggest fixes :)

humphd:
> Here is a small refinement of your work. I made it scan all human + system message for functions. this allows a more natural convo form. I also allowed functions to provide own formatting.
> 

This is useful, thanks.  We should make `chat.messages({...})` take options to allow easily getting just those message types so you don't have to write code manually.

> 1. I find the experience of having functions listed in sidebar to be good..nice to be able to switch between function and chat in sidebar
> 2. I find using chatcraft to edit functions frustrating. I propose we just list urls to gists or other CORS-fetch()able content..then just add a "reload functions" button(if functions are present) for handling the corner case of gists changing. I provided a sample of how that would look in the screenshot...Would still like to be able to jump to them using sidebar per point_1
> 3. I think gists-for-functions also keep us to true to cause for sharing chats...no need to figure out how to share functions

I can see the value of being able to point at a URL and have it get pulled in, similar to how we pull code in from db.  However, I disagree that this is the best experience.  I think we should support both local (in db, edited in ChatCraft) and remote (URL) functions.

In addition, I'd be interested to her what's frustrating about editing locally.

Some thoughts about functions as URLs:

1. There will be two cases: 1) function served as `text/plain` (e.g., gist), which needs to be downloaded and parsed into a module (i.e., like what we do from db now); 2) function served as ES Module `text/javascript` (i.e., CDN), which we can simply `import(url)`.  Both assume CORS, but we could proxy through Cloudflare too.  The latter is already usable by doing `import whatever from 'url';` in your function.
2. What happens when I use a URL function in a chat?  Do we store the metadata info in memory (i.e., just for this session) or in db (i.e. ,survives sessions)?  The answer will help guide what happens in sidebar: should the sidebar show you all functions currently "in scope" (i.e., what's in db + what's memoized from previous use)?  What happens when you switch to a new chat?  Do the URL functions from the other chat live there too, or are they only accessible in the one where they were mentioned?
3. Locking us into gist as our function backend seems short-sighted.  A URL to a gist requires us to build infra for loading modules from any URL, so that opens the door to the whole web.
4. How would a URL function in the sidebar act when you click it? Would it redirect you?  Would it open the code in the local functions editor (maybe read-only)?

You talk about using functions in the system message, which I agree is a logical place for them.  However, I'm not sure you'd only want that.  For example, if I'm chatting with an LLM, and it suddenly gives me a bunch of numbers I want to do math on, I'd need to use a function part-way through the chat vs. from the start.  I think we should support functions being called anywhere in the chat.

> 4. I think we could also teach chatcraft to comment in gists when we share conversations with functions for SEO and general goodness :)
> 5. I think in a followup we can get rid of special syntax for listing functions..we can just run a special chatgpt call every time system message changes, that extract all functions using some gpt parsing-using-functions trick..this would also validate the json schema for loaded functions..and could even pass those through another gpt prompt to suggest fixes :)

You'll have to expand on these, I don't fully understand.

> 6. This is so freaking cool to finally be able to hack on this!

Thanks!  I'm really happy to have something less hand-wavy that we can talk about.  It's been hard getting to this point, and I'm excited it works.

humphd:
As an experiment to play with your ideas, I've pushed another commit that lets you do this:

```text
// Use one or more function names in the db
@fn: localFunctionName

// Use a URL to a function we load externally
@url: https://gist.github.com/...
```

I have a gist at https://gist.github.com/humphd/647bbaddc3099c783b9bb1908f25b64e which you can use like this:

Screenshot 2023-07-23 at 11 18 45 AM

One thing that sucks about gists is that to get a raw body (vs. HTML), you have to use https://gist.githubusercontent.com/humphd/647bbaddc3099c783b9bb1908f25b64e/raw/811dac8a21948444fdd2c63dc05ac53028505425/sum.js.  We could add some code to convert the format of the URL I guess.

Anyway, this works.  I've only lightly tested this code, so it might have bugs if you mix local + remote functions.

tarasglek:
> One thing that sucks about gists is that to get a raw body (vs. HTML), you have to use https://gist.githubusercontent.com/humphd/647bbaddc3099c783b9bb1908f25b64e/raw/811dac8a21948444fdd2c63dc05ac53028505425/sum.js. We could add some code to convert the format of the URL I guess.
> 
> Anyway, this works. I've only lightly tested this code, so it might have bugs if you mix local + remote functions.

per chatcraft:
```markdown
For the given Gist URL (https://gist.github.com/humphd/647bbaddc3099c783b9bb1908f25b64e), the raw URL would be:

https://gist.githubusercontent.com/humphd/647bbaddc3099c783b9bb1908f25b64e/raw

You can use this raw URL to access the raw content of the Gist programmatically.
```

humphd:
@tarasglek I went over the code, I think it's probably good/safe to land.  But I have one question: will what we did here break interactions with models that don't support functions?  That is, do we need to add code to specifically check if the model being used can handle functions, or will it safely ignore them?  Can you try with your OpenRouter account?

If that's OK, I think this can land.

humphd:
Reading the [OpenAI docs on function calling](https://platform.openai.com/docs/guides/gpt/function-calling), I think we need to append a message type for function responses, with `role: function`.  Here's an example:

```python
import openai
import json


# Example dummy function hard coded to return the same weather
# In production, this could be your backend API or an external API
def get_current_weather(location, unit="fahrenheit"):
    """Get the current weather in a given location"""
    weather_info = {
        "location": location,
        "temperature": "72",
        "unit": unit,
        "forecast": ["sunny", "windy"],
    }
    return json.dumps(weather_info)


def run_conversation():
    # Step 1: send the conversation and available functions to GPT
    messages = [{"role": "user", "content": "What's the weather like in Boston?"}]
    functions = [
        {
            "name": "get_current_weather",
            "description": "Get the current weather in a given location",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g. San Francisco, CA",
                    },
                    "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
                },
                "required": ["location"],
            },
        }
    ]
    response = openai.ChatCompletion.create(
        model="gpt-3.5-turbo-0613",
        messages=messages,
        functions=functions,
        function_call="auto",  # auto is default, but we'll be explicit
    )
    response_message = response["choices"][0]["message"]

    # Step 2: check if GPT wanted to call a function
    if response_message.get("function_call"):
        # Step 3: call the function
        # Note: the JSON response may not always be valid; be sure to handle errors
        available_functions = {
            "get_current_weather": get_current_weather,
        }  # only one function in this example, but you can have multiple
        function_name = response_message["function_call"]["name"]
        fuction_to_call = available_functions[function_name]
        function_args = json.loads(response_message["function_call"]["arguments"])
        function_response = fuction_to_call(
            location=function_args.get("location"),
            unit=function_args.get("unit"),
        )

        # Step 4: send the info on the function call and function response to GPT
        messages.append(response_message)  # extend conversation with assistant's reply
        messages.append(
            {
                "role": "function",
                "name": function_name,
                "content": function_response,
            }
        )  # extend conversation with function response
        second_response = openai.ChatCompletion.create(
            model="gpt-3.5-turbo-0613",
            messages=messages,
        )  # get a new response from GPT where it can see the function response
        return second_response


print(run_conversation())
```

The key bit here is `Step 4`:

```python
# Step 4: send the info on the function call and function response to GPT
        messages.append(response_message)  # extend conversation with assistant's reply
        messages.append(
            {
                "role": "function",
                "name": function_name,
                "content": function_response,
            }
        )  # extend conversation with function response
        second_response = openai.ChatCompletion.create(
            model="gpt-3.5-turbo-0613",
            messages=messages,
        )  # get a new response from GPT where it can see the function response
        return second_response
```

We sort of do this now, but put it into an AI message. That should get fixed to add a `ChatCraftFunctionMessage` and some UI to display it.

Either I should do that in this PR, or in a follow-up.

humphd:
I've pushed a new commit that does the work described above, namely, we store all the function data in a single `ChatCraftFunctionMessage` and display it with the `FunctionMessage` component; however when we send it to OpenAI via langchain, we split the function call and function result into two messages (AI Message + Function Message).  This way the chat has the proper context it needs to remember what it did (e.g., which function was called, which args were passed, what was the result, etc).

Probably this is all I should do on this PR.  Test and see if you're happy.

One thing I know this broke is our token counting when functions are used, since OpenAI does that differently.  It will still work for normal chats.

tarasglek:
> That is, do we need to add code to specifically check if the model being used can handle functions, or will it safely ignore them? Can you try with your OpenRouter account?

This only works with the one chatgpt model, but it works on both openrouter and openai...So basically add a hardcoded a check to our model wrapper class to see if functions are supported, right next to [logoUrl()](https://github.com/tarasglek/chatcraft.org/blob/4ba2d0ec9393df8e9240fcb5e4211ef5073dd01a/src/lib/ChatCraftModel.ts#L20)

humphd:
@tarasglek good idea, I've added a check to see if the model supports function calling.  In my experience, all of the OpenAI gpt models *can* do it, but some are better at it.  I've also noticed that I don't need to turn off streaming when using functions, which was bugging me when it responded with text, but I had to wait for it.

Can you review/test this?  I think it's probably good to land if you're happy.

humphd:
From call with @tarasglek today, some things we need to change and/or adapt with how this works:

1. Currently, we store a function call + result in a `ChatCraftFunctionMessage`.  This doesn't map well to how functions work, not least because OpenAI requires 2 separate messages (one for call, one for result) to build the full context.  We should split `ChatCraftFunctionMessage` into a call and result as well.  One nice benefit of this is that it lets us show that something is happening while a function is running (e.g., might take some time).
2. We need a way to send the function call + result back to the LLM, so it can answer follow-up questions based on that new context.  We talked about adding some UI to let you trigger that, optionally being able to specify to always send results back during this chat session.
3. If an `fn` or `fn-url` are referenced in a system prompt, we could let the LLM decide whether to use it.  However, if referenced in a human message, and there's only one, we should probably send that function via the `function_call` param to the model, indicating that we want it to call this function (i.e., mentioning a function in a human prompt is intent vs. suggestion).

I think I should tackle the db side of 1. before we land this (since it will change things), and the rest can happen in follow-ups.

humphd:
OK, I've completed 1. above, see https://github.com/tarasglek/chatcraft.org/pull/172/commits/86359eec66a26d2193c48d7fbd5c5eb1567f1a6a.

Here's what this does:

- Like OpenAI, a "function" now has two parts: a function call, and a result.  I've created a `ChatCraftFunctionCallMessage` and a `ChatCraftFunctionResultMessage` to handle these.
- In the UI, `FunctionCallMessage` and `FunctionResultMessage` components take care of rendering these.
- I've moved the function invocation step out of the `ai`/`use-chat-openai` code, and put it *after* this code runs in `ChatBase`.  When a response comes back from the LLM, we add it to the message list, then check if it's a function and execute it if it is, adding the result when it completes.  It's the same logic, but split up across multiple files, since we now do it in stages.

I've also made a change to how the auto-scrolling works, and I'll need to build on it in follow-ups.  Previously, I only auto-scrolled when a streaming messages was coming in (and changed its value).  This breaks when we add new messages to the chat, so was broken for functions.  Fixing this is good for the function case, but it will create some weirdness when you delete a message high-up in the middle of a chat, or create new versions of a chat, etc.  I'll fix that later, but if it bugs people, we can disable the dependency on `chat.date` in the autoscroll effect.

This code was a bit of a pain to write and think through, but it's a much better design than what I had before.

@tarasglek please test this again, and I think it's ready when you're happy.

tarasglek:
Works fantastic in my testing. Lets get this landed

humphd:
OK, I've filed a bunch of follow-ups, and we'll need to probably file new things to improve this as we work with it.  Going to land.

tarasglek:
This is the greatest pull request ever, excited :)

humphd:
> This is the greatest pull request ever, excited :)

I think we're on to something!  Thanks for helping me figure it out.

PR #174, opened 2023-07-22
WIP: add bun + mem tracking to support free replit nodes

tarasglek:
I noticed that bun is a lot faster and can use less ram than vite.
### vite
```bash
`which time` -v pnpm vite build
```
Results in
```
User time (seconds): 27.39
Maximum resident set size (kbytes): 2172708
```
Which is over the replit 1gb limit

### bun
```bash
`which time` -v pnpm build-bun
```
Whereas bun(without minification) gets us down to 800mb
```
User time (seconds): 0.96
Maximum resident set size (kbytes): 798232
```

Downside is that bun doesn't produce a working site yet

This is pretty hopeless to get to work with bun directly. bun is too immature, they announced vite support in latest release.

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,12 +1,5 @@
-run = "vite --host 0.0.0.0"
+run = "bun --bun vite dev --host 0.0.0.0"
 hidden = [".build", ".config", "package-lock.json", "tsconfig.json"]
-
-[packager]
-language = "nodejs"
-  [packager.features]
-  enabledForHosting = false
-  packageSearch = true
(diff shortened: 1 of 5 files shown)


tarasglek:
I don't know why I keep falling for playing with bleeding edge software. They clearly marked their vite support as "lol, good luck" https://bun.sh/blog/bun-v0.7.0

humphd:
Bun has lots of potential, but it's really not ready (no support on Windows either).

Issue #101, opened 2023-06-09
Cant login on Chrome


harrytormey:
1) Tapped icon to share.
2) Clicked github signin.
3) Went through github auth, landed back on chatcraft not logged in despite saying yes.
4) Not tapping sign in does not allow me to sign in or do anything so I cant share urls.

harrytormey:
Recording:
https://github.com/tarasglek/chatcraft.org/assets/79118283/d55d8abb-f06e-4fde-9c3f-4ff24763a961

tarasglek:
heard through backchannel that it's unclear why we require signin for sharing..should add a blurb saying that this is a spam-prevention measure

humphd:
I'm going to close this, since I can't reproduce.  Please re-open or file again if it becomes a blocker to anyone.

humphd:
@harrytormey we figured out your issue, and I have a fix up for it in https://github.com/tarasglek/chatcraft.org/pull/179.  Sorry we couldn't figure this out sooner!

PR #179, opened 2023-07-25
Use GitHub login name if user's name is missing

### LINKED ISSUE #175: Github signin issue
**Issue**

I tried to sign in with my other github acct, but even after authorizing it doesn't connect to chatcraft.

Tested on normal chrome browser and a new incognito window. Also tried doing it in a separate fresh chrome profile but still not working

**Demo/How to Reproduce**

https://drive.google.com/file/d/1yhWhMAjPZf1XdaOvqSpuyif_Q0DBWoHF/view?usp=sharing

Steps:
- Sign out of GH account
- Use another GH account to sign in
- Authorize the GH oauth page
- Go back to chatcraft
- Still not connected + unable to share prompts

humphd:
Fixes #175 

@steven4354 and I figured out that an account on GitHub may or may not include a user's name (it's optional).  What that happens, and we don't find a name in the cookie, we fail to use the data.

This fixes things so we re-use the GitHub login if the name.

I've also added a warning in the user context/hook, so we know if this happens again.

### DIFF
--- a/functions/github.test.ts
+++ b/functions/github.test.ts
@@ -13,18 +13,25 @@ export const githubMocks = () => ({
       .reply(200, { access_token: mockAccessToken });
     return origin;
   },
-  "api.github.com": () => {
+  "api.github.com": ({
+    login,
+    name,
+    avatar_url,
+  }: {
+    login: string;
+    name: string | null;
+    avatar_url: string;
+  }) => {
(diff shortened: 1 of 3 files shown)


humphd:
Also fixes #101 (same issue).

tarasglek:
Nice catch, good to land

PR #180, opened 2023-07-25
Fix eslint issue in use-settings.tsx

humphd:
Noticed we had an eslint issue in the changes we made in `use-settings.tsx` for OpenRouter.  I've moved the `handleOpenRouterCodeExchange` function into the effect, and disabled the warning for exhaustive deps (I don't want to re-trigger this for anything other than the `openRouterCode` changing). 

@Chigala can you review this please?


Chigala (approved):
It looks ohk, sorry I'm just replying now.

PR #183, opened 2023-07-27
Add errorElement with better fatal error UX

### LINKED ISSUE #149: Deal with db upgrades without crashing
This is a bit of an edge case, but since we encourage people to open ChatCraft in multiple tabs, we should probably make db migrations not crash secondary tabs.

When doing a db migration, the database gets updated in the current tab.  This will force the db to get closed in any other open tabs, causing an error (we could catch it, but currently don't).

We might want to [close and re-open](https://stackoverflow.com/questions/39000780/errors-with-indexeddb-versions-and-dexie-js/39015986#39015986) the db to deal with the `versionchange` event instead.

We could do something like:

```js
const db = new Dexie("MyDatabase");

db.version(1).stores({ friends: "++id,name,age" });

db.version(2).stores({ pets: "++id,name,breed" }).upgrade(tx => {
  // Upgrade logic here
});

db.on('versionchange', (event) => {
  // Close the database connection
  db.close();
  // Reopen the database
  db.open().catch(err => {
    console.error(`Failed to reopen db: ${err.stack || err}`);
  });
});

db.open().catch(err => {
  console.error(`Failed to open db: ${err.stack || err}`);
});
```

It might be good to have `versionchange` and [`DatabaseClosed` errors](https://dexie.org/docs/DexieErrors/Dexie.DatabaseClosedError) trigger UI to let the user know they need to refresh, or we could do it automatically?

### LINKED ISSUE #181: Provide an errorElement to show when the router fails (e.g., chat.save())
image

Looks like save() can throw, there are a low of unprotected saves https://github.com/tarasglek/chatcraft.org/blob/4ba2d0ec9393df8e9240fcb5e4211ef5073dd01a/src/router.tsx#L31

They are intermittent cos they happen during routing. This renders app unusable for some users.

humphd:
Fixes #181
Fixes #149 

This adds top-level error handling via [`errorElement` in React Router](https://reactrouter.com/en/main/route/error-element).  If our routes fail, we'll show better error UI.

The vast majority of errors are handled, and those that aren't are likely show-stoppers (e.g., database not accessible for some reason).

In the case of database connection issues, closing other tabs and reloading should fix.

Screenshot 2023-07-27 at 12 09 22 PM

In the example above, you can see that my db won't open locally because it's version is newer than the one it wants to use on my current branch (i.e., I ran a migration on another branch).  This is an example of what a db error would look like.

### DIFF
--- a/src/AppError.tsx
+++ b/src/AppError.tsx
@@ -1,0 +1,97 @@
+import {
+  Flex,
+  Grid,
+  GridItem,
+  Heading,
+  Center,
+  Card,
+  CardBody,
+  CardHeader,
+  Text,
+  Container,
+  Icon,
+  CardFooter,
+  Button,
+  Box,
+  Code,
+} from "@chakra-ui/react";
+import { MdErrorOutline } from "react-icons/md";
+
+import Header from "./components/Header";
+import { useRouteError } from "react-router-dom";
+
+export default function AppError() {
+  const err = useRouteError() as any;
+  console.error(err);
+
+  return (
+    <Grid
+      w="100%"
+      h="100%"
+      gridTemplateRows="min-content 1fr"
+      gridTemplateColumns="1fr"
+      bgGradient="linear(to-b, white, gray.100)"
+      _dark={{ bgGradient: "linear(to-b, gray.600, gray.700)" }}
+    >
+      <GridItem colSpan={2}>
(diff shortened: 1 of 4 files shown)


tarasglek (approved):

tarasglek:
Good to land

PR #186, opened 2023-07-28
Add hint to function example about formatting output

humphd:
Adds a note to the function sample code so people know you can return Markdown code blocks.


tarasglek:
How are people supposed to discover this example?

humphd:
This is what you get when you create a new function in the app.

We need proper docs for this (plus everything else) in addition.

PR #187, opened 2023-07-29
Handle sending function results back to LLM in UI

### LINKED ISSUE #184: Provide way to send function results to LLM
Spin-off from https://github.com/tarasglek/chatcraft.org/pull/172#issuecomment-1652537723:

> We need a way to send the function call + result back to the LLM, so it can answer follow-up questions based on that new context. We talked about adding some UI to let you trigger that, optionally being able to specify to always send results back during this chat session.

humphd:
Fixes #184 

When a function result comes back, we need a way to send it to the LLM for further processing (e.g., if result is part of a larger discussion, and we need to continue).  This adds UI in a function result message to allow sending the result back to the LLM.

Users can either manually send, or specify that results should always be sent.  Their preference is remembered.

### User must manually click "Send" by default
Screenshot 2023-07-29 at 7 34 06 AM

### Checking "Always Send" does it automatically
Screenshot 2023-07-29 at 7 33 57 AM

I wonder if the preference to "Always Send" should be true by default, which would mimic what OpenAI docs do for function calling.  I think OpenAI imagines that function calls/results are internal to an app vs. something you display (I like that we display it, just reflecting on what they do).

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -112,13 +112,15 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt: string) => {
+    async (prompt?: string) => {
       setShouldAutoScroll(true);
       setLoading(true);
 
       try {
-        // Add this prompt message to the chat
-        await chat.addMessage(new ChatCraftHumanMessage({ text: prompt, user }));
+        if (prompt) {
+          // Add this prompt message to the chat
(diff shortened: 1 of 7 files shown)


humphd:
Here's a question: do we care that the `Send` button is going to be visible/functional/enabled after the user sends the result?  Do I need to hide/disable it when the result has been seen by the LLM?  If so, I think I have to add state to the db, which sucks.  I can't figure out what the right answer is.

tarasglek:
that's sexy. I think for database use-cases, it would be handy to have an "exclude" option where we never send the results back to llm...eg we'd switch from checkbox to a menu that has "send ondemand|always|exclude"

PR #189, opened 2023-07-29
Tell LLM to use function if mentioned in prompt

### LINKED ISSUE #177: Add syntax for forcing LLM to use a function
In #172 we're adding support for function calling, introducing the `@fn: name` or `@fn-url: https://...` syntax to include references to `functions` in a chat when sent to OpenAI.

The way function calling works, you can also also set the `function_call` parameter.  By default it's value is `auto`, meaning that the LLM can decide to use or ignore the functions.  You can also set it to `none` (don't use any functions) or `{"name": "function-name-to-call"}` to force it to use a function.

It might be nice to add a way to signal your intention to do this when referring to the function.  For example, one of these:

```
@fn: foo!
@fn!: foo
@!fn: foo
```

I think putting it at the end might work `@fn: foo!`, as it feels a bit like `!important` in CSS.

If a function is referred to with this special syntax, we'd add that to the `function_call` property we send to the model.

What do you think?

### LINKED ISSUE #185: Be smarter about when to recommend vs. force the use of functions
Spin-off from https://github.com/tarasglek/chatcraft.org/pull/172#issuecomment-1652537723:

> If an fn or fn-url are referenced in a system prompt, we could let the LLM decide whether to use it. However, if referenced in a human message, and there's only one, we should probably send that function via the function_call param to the model, indicating that we want it to call this function (i.e., mentioning a function in a human prompt is intent vs. suggestion).

humphd:
Fixes #185
Fixes #177 

When a chat includes functions (`fn` or `fn-url`), we send those functions to the LLM.  Previously, we let the LLM decide whether or not to call a function.  However, this PR changes things so that if a user explicitly mentions a **single** function in the current prompt, we ask the LLM to use it.  If previous messages, or the system prompt, included functions, we still let it decide.

In this example you can see that the LLM doesn't actually need a function to answer `1+1`.  However, when I mention the `sum` function in the prompt, it uses it.

Screenshot 2023-07-29 at 8 39 30 AM

NOTE: this code needs to land after my earlier PRs, since I fixed some bugs that this exposes.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -117,9 +117,11 @@ function ChatBase({ chat }: ChatBaseProps) {
       setLoading(true);
 
       try {
+        let promptMessage: ChatCraftHumanMessage | undefined;
         if (prompt) {
           // Add this prompt message to the chat
-          await chat.addMessage(new ChatCraftHumanMessage({ text: prompt, user }));
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, user });
+          await chat.addMessage(promptMessage);
         }
 
         // In single-message-mode, trim messages to last few. Otherwise send all.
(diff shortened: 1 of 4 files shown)


tarasglek:
This seems good. Code is a little hairy, but we can simplify it later once we understand where this is going better

PR #193, opened 2023-07-31
Update deps July 31, 2023, switch to pnpm

humphd:
Updating deps and switching over to pnpm from yarn.


humphd:
CloudFlare is using `npm` without our lock file to do the builds.  We can switch to the build environment v2 to get pnpm installed for us.  I won't do that until this lands though, and can test.

humphd:
https://community.cloudflare.com/t/add-pnpm-to-pre-installed-cloudflare-pages-tools/288514/8

humphd:
I'm going to test upgrading the CloudFlare build env to v2, which failed for me before.  I'll start with the preview builds and see if it goes OK...

humphd:
OK, that worked: CloudFlare v2 build env and `pnpm build`.

I'll do the same for prod and land this.

Issue #194, opened 2023-08-01
Help Wanted: ChatCraft User Feature Requests


humphd:
@tarasglek and I have spent the spring and summer implementing what *we* want in ChatCraft.  I think we've ended up making something pretty great.

However, I'd be interested to hear what *you*, our users would like to see us to do next.  If you're using ChatCraft and have suggestions, please add a comment here.  What would make this app better for you?  What would get your friends and colleagues to start using it?  How can we improve it for your needs?

No promises, but it would be great to build a list.

rashboldb:
> > I'd really love to see it has some ASR system support, like Whisper!
> 
> Can you describe this a bit more?

I'd like to have the ability to communicate with chat (I can imagine it'll be mostly via mobile) with my voice. So once I can do this I can use it like [audiopen](https://audiopen.ai/) (pretty cool stuff) by modifying the prompt and let's say asking it to summarize & clean up my thoughts. It's only the example, and it's definitely not the most technical use case of the chat (which was the original idea, as far as I know) but that'll be awesome:)

rashboldb:
BTW, someone already [did](https://github.com/chengsokdara/use-whisper) react hook for smth like this in the past, open source, it was update like 5 month ago the last time though.

humphd:
Right, I've seen that hook before, it might be a good option.  I'll do some reading...

tarasglek:
I think he means something like https://gitlab.com/attention1/ai-interface

steven4354:
i'd love the ability to save the system messages to my profile so I can quickly swap between a few system messages I have (perhaps via a dropdown) and set a "default" system message to use whenever I open new blank chats

humphd:
> i'd love the ability to save the system messages to my profile so I can quickly swap between a few system messages I have (perhaps via a dropdown) and set a "default" system message to use whenever I open new blank chats

I'm literally working on this right now!  I'll have a PR up soon, and you can help me figure out the UX, it's tricky to get right and make clear to users...

humphd:
Also, "ability to set system prompt default" is #138

humphd:
https://github.com/tarasglek/chatcraft.org/pull/196 is up with a partial fix for system default messages.

alexisargyris:
Here are a few random thoughts:

* the UI looks nice and is quite flexible (small/large screen, landscape/portrait, etc.). Sometimes, when entering code, it doesn't get properly formatted or a line at the top or bottom gets skipped
[...]

PR #196, opened 2023-08-01
Ability to set and reset default system prompt

### LINKED ISSUE #138: Allow switching to previous system prompts, setting default
Now that we can see and edit the system prompt, something I think we should do next is allow the user the ability to easily switch between previous system prompts they've entered.  @steven4354 has a video that hints at this, and in Discord another user requested something like this:

1. https://www.youtube.com/watch?v=kZEBevvoPYM
2. "what'd be even more cool imo is to have the ability to save 'favourite' prompts you like for the future uses. just like with chats, but with prompts. to have smth like templates. because the one like I linked above is kinda a big one, a I don't like I have to copy paste and/or modify it each time I wanna use it."

I had a brief discussion with ChatCraft about it as well, https://chatcraft.org/c/humphd/aREmDBWjgDJjkKVHPe9QB

I haven't figured out the right UX yet, but here's what I imagine:

1. You start a new chat, and we show you the default system prompt.  There is a button to "edit" (we already have this in the menu, and #136 is exposing it top-level), but also a button to "Use previous" (wording TBD).
2. Clicking "Use previous" shows you a new UI in place of the current system prompt message, with an easy way to jump between all the old system prompts you've entered, sorted by date (most recent first).
3. You can select one of these old prompts, and it updates the current prompt
4. A custom prompt message like this can include another button in the bottom-right or somewhere to "Set as Default" and another to "Switch to Default" to get back to the ChatCraft prompt.
5. If you start a new chat with a different default prompt set, we automatically use that instead.

This would allow users to completely change how ChatCraft works for their needs, and use that all the time.  It would also allow users to re-use old workflows more easily.

humphd:
Fixes #138, addresses part of https://github.com/tarasglek/chatcraft.org/issues/194#issuecomment-1661036662.

When we create a new chat, we always begin with a default system prompt.  The text of this system prompt is baked into the app's code.  We allow the user to edit it in the current message, but you can't use it again in a new chat.

This change allows the default system prompt to be overridden by a user (or reset back to the default).  The process is this:

1. User edits the current system prompt
2. In the 3-dots menu for the system prompt, a new set of options is shown when the system prompt has been changed (i.e., it is NOT shown if it's not needed):
    1. Users can set the default system prompt to the current system prompt (i.e., override)
    2. Users can reset the overridden system prompt back to the default (i.e., reset)

Here's a video showing how this works in practice (sorry it's small, GitHub won't let me post anything > 10M):

https://github.com/tarasglek/chatcraft.org/assets/427398/2491af38-6bca-4c89-b981-b7fbb69f9450

@steven4354 can you try this and comment?  NOTE: I'm not tackling the part about saving multiple custom prompts and choosing between them, mainly because I'm not sure how to do the UI.  I wonder if we should start with this and iterate?  Maybe you have ideas now.

### DIFF
--- a/src/components/DefaultSystemPromptModal.tsx
+++ b/src/components/DefaultSystemPromptModal.tsx
@@ -1,0 +1,96 @@
+import { RefObject, useState } from "react";
+import {
+  Button,
+  Flex,
+  FormControl,
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalFooter,
+  ModalBody,
+  ModalCloseButton,
+  VStack,
+  Text,
+  Textarea,
+} from "@chakra-ui/react";
+import { debounce } from "lodash";
+
+import { useSettings } from "../hooks/use-settings";
+import { defaultSystemPrompt } from "../lib/system-prompt";
+
+type DefaultSystemPromptModalProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  finalFocusRef?: RefObject<HTMLTextAreaElement>;
+};
+
+function DefaultSystemPromptModal({
+  isOpen,
+  onClose,
+  finalFocusRef,
+}: DefaultSystemPromptModalProps) {
+  const [prompt, setPrompt] = useState(defaultSystemPrompt());
+  const { settings, setSettings } = useSettings();
+
+  // Update the default prompt, but not on every keystroke
+  const handleSave = debounce((value: string) => {
+    setSettings({ ...settings, customSystemPrompt: value });
+  }, 250);
+
+  // When the prompt changes, update state, and trigger a save for later
+  const handleUpdate = (value: string) => {
+    setPrompt(value);
+    handleSave(value);
+  };
+
+  // Reset the default system prompt back to the app's default
+  const handleReset = () => {
+    setSettings({ ...settings, customSystemPrompt: undefined });
+    setPrompt(defaultSystemPrompt());
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} size="lg" finalFocusRef={finalFocusRef}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Default System Prompt</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <VStack gap={4}>
+            <Text fontSize="sm">
+              The default system prompt is used as the first message in every chat. It provides the
+              LLM important context, behavioral cues, and expectations about responses and desired
+              response formats.
+            </Text>
+
+            <Text as="em" fontSize="sm">
+              NOTE: changes to the default system prompt will take effect in new chats, but
+              won&apos;t affect existing ones.
+            </Text>
+
+            <FormControl>
+              <Textarea
+                autoFocus
+                rows={15}
+                value={prompt}
+                onChange={(e) => handleUpdate(e.target.value)}
+              />
+            </FormControl>
+          </VStack>
+        </ModalBody>
+
+        <ModalFooter>
+          <Flex w="100%" justifyContent="space-between">
+            <Text fontSize="sm">Reset to the default ChatCraft system prompt</Text>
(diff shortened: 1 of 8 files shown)


humphd:
For showing previous system prompts, I could easily query the db to get all previous prompts, maybe sorted by date used or something (i.e., I don't need to add any new storage for this to work).  But how do I show it?  A drop-down for something that fits in a textarea isn't really going to work (not enough context to pick them).

steven4354:
Just took a look at the video, that looks great and would solve the problem for me

**A drop-down for something that fits in a textarea isn't really going to work (not enough context to pick them).**

Not sure if I'm 100% on the same track (pls lmk if I am), but are you saying the drop-down of the prompt titles? 

If so, I guess you would need a flow to "name" each system prompt as well right, maybe set a max length for the title and then only show in the dropdown the ones that they "named" + "saved"? 

So "saving" and "naming" a system prompt is required to show in the dropdown

humphd:
Thanks for the feedback.  Here's a concept to let the user choose from previous system prompts.  I'm borrowing UX from the way that AI Message versions work, giving the system message a dropdown in the heading:

Screenshot 2023-08-01 at 9 14 49 PM

Clicking that reveals your old system prompts, sorted by most recently used, and clicking one updates the current system prompt to use it:

Screenshot 2023-08-01 at 9 15 23 PM

There's a scrolling/height bug that this exposes, but it can be fixed in a follow-up.

I also realized that I need a way to set a *different* custom system prompt in my menu solution (i.e., if you already have a custom system prompt, but you want to use a different one, we need to somehow differentiate it).

Take a look at this and see if things make sense, or what you'd do differently.

humphd:
I've pushed my updated code, if people want to try it out.  You might want to export/import your db from production so you can get the full effect (do that in the Settings modal).

tarasglek:
I like the system prompt dropdown concept. Why not add a rename button in *ChatCraft (System Prompt)*, so we could list them by name

tarasglek:
Options 2 would be ok for me.

humphd:
OK, I've reworked this to allow users to "star" a system prompt, and then only show the ones they've starred.  I also reworked the way the default system prompt works to let you do it via its own dedicated modal.  Trying to do this via the menu was too complicated, and I think this is simpler.

Screenshot 2023-08-10 at 12 21 45 PM
Screenshot 2023-08-10 at 12 17 26 PM
Screenshot 2023-08-10 at 12 17 42 PM

Screenshot 2023-08-10 at 12 17 14 PM

NOTE: this change needs #198 to get scrolling to not jump to the bottom when you star/unstar the system prompt.

tarasglek:
I just tried this.
1. edited the system prompt
2. stared it
3. now I have no more default prompt in the menu. This is scary.
image

"new blank chat" brings the old prompt back, but menu still doesn't show it. We should star the default prompt.

Feels like a scary amount of state for a very corner-case feature...How about tagging shared chats as "chat-templates".

humphd:
To clarify what I'm reading above, you're saying the default system prompt should also be in that drop-down menu?  If so, that's a good idea.

humphd:
Rebased and updated to include the default system prompt as first item in list of starred entries.

tarasglek:
I wonder about having this sort of complexity in indexeddb...For me it would be better to keep such a "personal prompt library" in cloud so i could use it across all my devices. Same goes for the /commands and @functions we are talking about in other issue

humphd:
"keep...in the cloud" is scope creep.  If we want to explore that, let's not do it in this PR.

How about the functionality of this change?  What do I need to do in order to finish it?  Same question to @steven4354.

tarasglek:
I def think this is sufficient. Not asking for scope creep, just noting that you are taking on a more complex client-side schema with this change.

tarasglek:
lets land the last version you posted

humphd:
Rebased, updated `toast()` calls to use new `useAlert()` hook, will land when this goes green...

humphd:
Go forth and create reusable system prompts!

steven4354:
I'm going to use this a bunch :)

PR #198, opened 2023-08-03
Add useAutoScroll() hook and context to fix autoscrolling

### LINKED ISSUE #105: Keep track of scroll position when retrying AI messages

humphd:
Fixes #105 

This fixes a bunch of bugs in how the auto scrolling feature works.  Specifically:

1. Adds a new React context and hook `useAutoScroll()`.  This keeps track of a React ref that points to the "bottom" of the message list (i.e. where we should auto scroll to in the messages), whether or not the UI should autoscroll, and also the current progress of scrolling (i.e., every time there's some new content added that requires autoscroll logic to fire).  Previously a bunch of this stuff was done as state in `ChatBase`, but I needed to share it across components.
2. When we render the `NewMessage` component, we use it as our "bottom."  This works both when doing a normal chat, and also retrying with a different LLM midway through the chat.
3. When the content of what we're streaming (i.e., a new AI Message) changes, we increment the `scrollProgress` in the autoscroll context, telling effects in the UI to fire and perform various autoscroll logic.
4. When we aren't streaming a response (i.e., `scrollProgress === 0`), we no longer scroll the whole chat (e.g., when deleting or editing a message part-way through).
5. I've increased the fudge-factor threshold for when to break autoscrolling as the user scrolls up.  Previously, it was happening too easily when the LLM added "tall" content (e.g., an iframe or mermaid diagram).  I think using `100` pixels is better.

A key idea to how this code works:

- `useChatOpenAI()` is usable in different parts of the app without sharing logic.  That is, the `streamingMessage` you get back is specific to the component where you're working.  Retrying a message with an LLM won't trigger any logic in the main chat, which also uses the same hook (i.e., it's just a hook, not a shared context).
- `useAutoScroll()` is shared across all the chat-related components in the app.  If you update something in it, it affects all other parts of the app (i.e., it's a context vs. only a hook).

I need to spend more time testing this, and thinking through the code, but I think it's just about ready to go, and makes the UI so much better to use!

Try it and see if you notice any problems.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -16,6 +16,7 @@ import { useSettings } from "../hooks/use-settings";
 import { useModels } from "../hooks/use-models";
 import ChatHeader from "./ChatHeader";
 import { ChatCraftFunction } from "../lib/ChatCraftFunction";
+import { useAutoScroll } from "../hooks/use-autoscroll";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -33,7 +34,8 @@ function ChatBase({ chat }: ChatBaseProps) {
     defaultIsOpen: settings.sidebarVisible,
   });
   const [loading, setLoading] = useState(false);
-  const [shouldAutoScroll, setShouldAutoScroll] = useState(true);
+  const { scrollProgress, shouldAutoScroll, setShouldAutoScroll, scrollBottomRef } =
+    useAutoScroll();
   const messageListRef = useRef<HTMLDivElement | null>(null);
   const inputPromptRef = useRef<HTMLTextAreaElement>(null);
   const toast = useToast();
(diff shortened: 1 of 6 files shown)


humphd:
@tarasglek when you have a chance, can you test this a bit?  I think it's good to go (minus bug with scrolling function stuff, which I'll do in follow-up).

tarasglek:
I can't see any bugs with either prod or your version atm. If you feel this is better, lets land :)

humphd:
OK, I'll do this and file a follow-up to figure out how to scroll function results into view.

PR #201, opened 2023-08-07
Update README for recent features, improve marketing

humphd:
I took a stab at updating our README to be more reflective of the current state of the app.  I've also tried to make it more compelling for someone landing on this with no knowledge of what we do.

This doesn't attempt to be proper documentation, but gives users a sense of what's possible.


tarasglek:
Looks good. Not sure how much value there is in having that within github vs chatcraft itself, maybe should do both?

tarasglek:
It's a good improvement over what we have atm

humphd:
I'll do another to add something similar to ChatCraft itself.  I think having both landing places giving useful info is good.

PR #202, opened 2023-08-07
Give reasons why ChatCraft is worth using when asking for API key

### LINKED ISSUE #133: Explain why chatcraft is good when asking for key
I went through all the alternatives I could find. Was really hard too find motivation to try them because most of them including us don't explain what user gains for the effort of signing up.

List of features plus link to youtube demo would probably be a good start.

https://www.futuredesk.io/welcome seemed most honest but didn't sign up up for them cos unclear as to how pricing works

https://speechgpt.app/ signed up for that cos I thought they would have whisper voice understanding...but turned out I need azure keys for voice 🤡.

I think longer term we need some easy sign up and billing...eg login via github and get free credits and option to pay via stripe....but either way need better job at explaining value

humphd:
Fixes #133 

I've migrated the text from the README and included your comparison table.  See if you think this is too long, needs editing, etc.

I've also removed the way the page title logic works, since it was always losing the summary for shared chats.

### DIFF
--- a/src/components/Message/AppMessage/Instructions.tsx
+++ b/src/components/Message/AppMessage/Instructions.tsx
@@ -17,21 +17,56 @@ import { useSettings } from "../../../hooks/use-settings";
 import { openRouterPkceRedirect, validateOpenAiApiKey } from "../../../lib/ai";
 import { OPENAI_API_URL, OPENROUTER_API_URL } from "../../../lib/settings";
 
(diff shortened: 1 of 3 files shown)


tarasglek:
1. Lets put the comparison table as high as reasonable after a brief intro
2. would prefer [more info on github] for readme copy...but can also leave it as is

humphd:
OK, made an update.  See if this is better, and please confirm that my checkmarks/Xs are correct in table (I added a few):

Screenshot 2023-08-08 at 10 00 10 AM

Issue #210, opened 2023-08-12
Save some context


tarasglek:
With functions it's easy to blow past all context limits. Atm i delete messages one by one once i feel I have too many.

1) Add "Delete Messages After This"
2) should be able to choose to not send function output to gpt when sending next message eg: [x] never send

humphd:
> 1. Add "Delete Messages After This"

Created https://github.com/tarasglek/chatcraft.org/pull/211 to fix this part.

PR #211, opened 2023-08-12
Add Delete Messages Before/After feature

humphd:
Part of #210

This adds two new menu items to a message:

1. Delete Messages Before - will remove all messages from the chat *before* this message (but not including it).  The system message is not affected.
2. Delete Messages After - will remove all messages from the chat *after* this message (but not including it).

These options only appear if they make sense (e.g., you can't delete messages after last message, etc).


tarasglek:
wow, how did i live without this.

Issue #209, opened 2023-08-12
replace functions in messages, not just detect em


tarasglek:
We currently scan messages for `@fn-url:..` to identify function calls... 
We should replace the url with function name when we serialize msgs to send to openai. It's not at all helpful to the model to see those urls, and it costs us tokens

humphd:
Adding some notes for anyone interested in doing this work.

Fixing this would require modifications to https://github.com/tarasglek/chatcraft.org/blob/bb69c02a989246e0a4922f6666af5090b8137723/src/lib/ai.ts#L175-L186

We receive a list of `messages` and possibly `functions`.  When we get `functions`, we'd have to:

1. Loop over all `message.text` and look for `@fn-url`.  There is already code to do this https://github.com/tarasglek/chatcraft.org/blob/bb69c02a989246e0a4922f6666af5090b8137723/src/lib/ChatCraftFunction.ts#L25-L57
2. Find the function `name` for the given `url` (it will be the function object's `id`)
3. Replace the `@fn-url: <url>` with the function's name in the content we send to the LLM (i.e., don't modify the original message content and lose the `@fn-url: <url>` syntax)

PR #212, opened 2023-08-12
Initial experiment adding slash commands

humphd:
This is an initial attempt to add "slash commands."  I'm not 100% sure about the UX yet, so I'm hoping that we can play with this PR and figure it out together.

A command is something a user enters in the prompt, and starts with `/`.  Currently, there are 4 commands defined:

1. `/new` creates a new blank chat
2. `/clear` clears the messages in the current chat
3. `/summary` or `/summary <word count>` creates a summary of the current chat, using `word count`.  For example, `/summary 15` would summarize in at most 15 words
4. `/github <github code URL>` fetches the file at the given GitHub URL and adds it to the chat as if you copy/pasted it.  For example: `/github https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/summarize.ts`

See what you think and we can iterate.

### DIFF
--- a/functions/api/proxy.ts
+++ b/functions/api/proxy.ts
@@ -1,0 +1,61 @@
+import { errorResponse, createResourcesForEnv } from "../utils";
+
+interface Env {
+  JWT_SECRET: string;
+}
+
+// Transform known HTML URLs to get raw content
+const fixUrl = (url: string) =>
+  url
+    // Deal with GitHub HTML URLs to code.  For example:
+    // https://github.com/<owner>/<repo>/blob/<branch>/<path> -> https://raw.githubusercontent.com
+    .replace(/^https:\/\/github\.com\/(.*)\/blob\/(.*)$/, "https://raw.githubusercontent.com/$1/$2")
+    // Fix gist URLs to get /raw data vs. HTML
+    .replace(
+      /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)$/,
+      "https://gist.githubusercontent.com/$1/$2/raw"
+    );
+
+// GET https://chatcraft.org/api/proxy?url=<encoded url...>
+// Must include JWT in cookie, and user must match token owner
+export const onRequestGet: PagesFunction<Env> = async ({ request, env }) => {
+  const { ENVIRONMENT } = env;
+  const { tokenProvider } = createResourcesForEnv(ENVIRONMENT, request.url);
+
+  // There should be tokens in the cookies
+  const { accessToken } = tokenProvider.getTokens(request);
+  if (!accessToken) {
+    return errorResponse(403, "Missing Access Token");
+  }
+
+  // Make sure the access token is valid and has API role
+  try {
+    const payload = await tokenProvider.verifyToken(accessToken, env.JWT_SECRET);
+    if (!payload) {
+      return errorResponse(403, "Invalid Access Token");
+    }
+
+    // Make sure this user has the `api` role
+    if (payload.role !== "api") {
+      return errorResponse(403, "Access Token missing 'api' role");
+    }
+  } catch (err) {
+    return errorResponse(400, err.message);
+  }
+
+  // Proxy the request to url
+  try {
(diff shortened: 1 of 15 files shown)


tarasglek:
I like where this is going, but this particular approach feels too object-oriented and isn't extensible.

1) we should have a mapping of functions + /commands 
2) mapping should be extensible via prompting/system-message just like we do with functions
eg I wanna make a chatcraft resume-review tool where i have an /approve and /reject commands that call my functions
3) we should delete command specifications from msg history ala #209 
4) I would like to define pre-existing commands in same way as user commands would be defined, except maybe the js for them is hosted on /chatcraft-commands/url.js

humphd:
Added an `/import <url>` command, which proxy's a `fetch()` through a Pages Function so that CORS isn't an issue.

This could be extended to to post processing, for example:

```
# load the text at example.com into a message
/import http://example.com

# load a PDF and get the text out of it, putting into a message. Second arg specifies type
/import http://example.com/file.pdf pdf
```

The pages function requires you to be logged in (like sharing) so we avoid people abusing proxy endpoint.  I don't know if it will work until it's pushed to prod, but it works locally if I comment out the security checks in `proxy.ts` and run `pnpm dev-functions` locally.

tarasglek:
we really should sort out a staging env, so server-side functionality should be tested

humphd:
agreed, it's a total pain to do this dev work now

humphd:
Let's get #217 landed so this is easier to test

humphd:
I've rebased this so that people can test on staging or locally.  I've also made the `/import <url>` command smarter for GitHub URLs.

Things to try (you need to login before you can run `/import`):

```
# Pull in raw content of a gist
/import https://gist.github.com/humphd/647bbaddc3099c783b9bb1908f25b64e

# Pull in a file from GitHub
/import https://github.com/jsdom/whatwg-mimetype/blob/master/scripts/get-latest-platform-tests.js

# Pull in a blog post's HTML
/import https://taras.glek.net/post/overlooked-hn-databases-that-process-data-faster-than-memory-bandwidth/
```

tarasglek:
`/help` gave me this
image

tarasglek:
`/load` fails when not logged in, doesn't tell you why

tarasglek:
I used `/load https://taras.glek.net/post/overlooked-hn-databases-that-process-data-faster-than-memory-bandwidth/index.md`, then asked chatcraft to be an editor and it gave me helpful feedback

chatcraft should show /commands that were run in history

I like this functionality a lot!

humphd:
OK, I'll add a `/help`, which is something we've needed forever anyway.

When you say `/load` do you mean `/import`?  Should I rename it to `/load` or support both?

For this:

> chatcraft should show /commands that were run in history

So I should add a message (app message?) to the chat showing that you typed `/whatever`?

tarasglek:
I'm ok just making it look like a "/load" function call, it's same thing with mechanics basically

humphd:
I've added a `/help` command (using `help` will also work), and made a note in the prompt form's default placeholder.

It will show a more comprehensive help message.  Please see if you think it's good, what needs to be added, etc.

I've added an error message about not being logged in for `/import`.

I'm still unclear on what you want me to do with the naming of `/import` vs. `/load`

tarasglek:
Do as you wish re /load,  load implies less conversion

tarasglek:
> So I should add a message (app message?) to the chat showing that you typed /whatever?

Yes. I think it should just have function call formatting
image
Function Call: load
load(url)

humphd:
@iramykytyn would you mind reviewing how the `/help` command works in this PR?  I'm trying to build in some help for totally new users (I think you've worked with some, based on what Taras has said?).

If you're willing, please let me know if it's too much, not enough, needs more or different examples, etc.

I know we'll need proper docs at some point, but I want something to get people over the "now what?" point of using ChatCraft.

tarasglek:
we should land this

humphd:
Added command info to response message for `/import` and `/summary`:

Screenshot 2023-08-25 at 9 09 06 PM

I'll land when this goes green.

Issue #214, opened 2023-08-13
Clicked share, nothing happened.


iramykytyn:
Link share stopped working without any error in UI. When I went to the debug console, it said that the share limit was reached.

<img width="1497" alt="Screenshot 2023-08-13 at 20 29 50" src="https://github.com/tarasglek/chatcraft.org/assets/6651168/fde68944-8d7b-41ee-b9fe-99954f3c9aae">

humphd:
Congrats on being the first person to hit this limit!  We have them coded in the function here:

https://github.com/tarasglek/chatcraft.org/blob/main/functions/api/share/%5B%5Buser_id%5D%5D.ts#L55-L69

Currently it's 10 shares in a 24 hour period, but this might be too low.

Also, we must have a bug (should be showing some error info in the UI, but clearly we aren't):

https://github.com/tarasglek/chatcraft.org/blob/26e4b153f652751f344f7089aec3f01e17e90e79/src/components/ShareModal.tsx#L48-L62

tarasglek:
i think we have a lot of cases where we should errors in Toasts that blink quickly and disappear...we have a similiar situation if openai aborts a message mid-stream

humphd:
> i think we have a lot of cases where we show errors in Toasts that blink quickly and disappear...we have a similar situation if openai aborts a message mid-stream

Could have error toasts not auto-close?  Might be annoying, not sure.  We could also dump errors into the chat, but I think that's going to be annoying in a different way (forcing you to clean up errors from your chat).

tarasglek:
I think having them show up as a moral equivalent of alert() would be fine..errors are rare. Having in them in chat log is probably ok too.

PR #215, opened 2023-08-13
Refactor Toasts into use-alert.ts hook

humphd:
From https://github.com/tarasglek/chatcraft.org/issues/214#issuecomment-1676429372, this change updates how we do info and error alerts.  The info alerts disappear after 3 seconds, while errors have to be closed manually (i.e. to give the user time to read them).

I've moved all the calls to the Chakra-UI `Toast` component/hook into a new `useAlert()` hook that manages the settings we use.


humphd:
If you want to test, try doing something like this:

```
Add 10 and 50 using @fn-url:https://google.com/function
```

This should error because it can't load/parse a function at that URL.  The error will stick around until you close it manually.

However, if you click the Copy button on a message, you'll get an info box that will vanish after 3 seconds.

tarasglek:
thank you for the quick fix!

PR #217, opened 2023-08-14
Allow server-side functions to work in dev environments

humphd:
Work to make it possible to use the CloudFlare functions locally or in preview URLs.

I had to add new environment variables to preview builds, and also created a new R2 bucket (`dev-chatcraft-org`), which I've bound to the preview environment.

Locally, you can use `pnpm dev-functions` to run this (login works, but I must need to do more to host an R2 bucket locally with wrangler...todo).  I've added this to the docs.

I want to test this in a preview build.


humphd:
OK, I have R2 buckets working locally too!  Everything is doable locally, in previews, and hopefully prod (can't test until we land).

A few notes about how this all works:

- Locally, we use `.dev.vars` to define the necessary environment variables we'll need when running the functions.  In preview builds and production, this is done via the CloudFlare console
- Locally, the `wrangler` CLI runs a local version of Pages Functions and R2 Buckets when you use `pnpm dev-functions` (we could make this the default for `dev`, not sure which is best)
- The API functions now use a `TokenProvider` which deals with cookies/tokens differently depending on the `ENVIRONMENT` value.  In `development` we don't use secure/host cookies and allow the origin to change; in `production` we lock everything down and force the use of `https://chatcraft.org`
- With `/api/login` in `development`, we skip GitHub and log the user in automatically with a fake account.
- The front-end now works with a secure cookie in prod, and less secure in dev/previews (since the origin changes, it can't be locked down as easily)
- The rest is mostly unit tests.

I think this is ready to land, and will make testing other stuff way easier.

tarasglek:
> > Confirmed: I can login and share on preview builds now!
> > https://0b6a7ad3.console-overthinker-dev.pages.dev/c/chatcraft_dev/EqMM4wSDdBRcL2V4HjS49
> 
> NOTE: I deleted this while testing. You can try sharing one yourself to see that it works.

yet it still works :)

functions/utils.ts
```diff
@@ -14,3 +16,20 @@ export function buildUrl(url: string, params: { [key: string]: string }) {
...
+  if (isDev) {
+    origin = new URL(requestUrl).origin;
+    appUrl = new URL("/", origin).href;
+  } else {
+    origin = "https://chatcraft.org";
+    appUrl = "https://chatcraft.org/";
+  }
+  const tokenProvider = new TokenProvider(environment, origin);
+
+  return { appUrl, isDev, tokenProvider };
```
tarasglek, line 34: Not a huge fan of unnamed tuples. please change to `return { appUrl: appUrl, ...`

tarasglek:
Rest is really good. With this change, this is now the best ci/cd i ever had for any web project

tarasglek (approved):

functions/utils.ts
tarasglek: ok, my mental type inference failed :) good to land

humphd:
It's possible this will break prod, so if you see things break there, file stuff.  I can't easily test on my current internet.

PR #220, opened 2023-08-17
Replace langchain with native openai library v.4.0.0

iramykytyn:
### Description
- Replaced langchain library with OpenAI native library v4.0.0.
- Implemented separate parsing for streaming and not-streaming options as opeai response handling.
- Reduced error handling logic since cancel is handled in .then(response) block now and there seems to be no need in such error parsing as it was with langchain. 

### How it was tested
Tested manually in the browser. Tested cancel, pause, and error handling as well - all seems to be working as expected.

### Known Issue
Openrouter is not working yet, it returns some CORS error, I'm investigating it:
Screenshot 2023-08-20 at 13 32 17

Screenshot 2023-08-20 at 13 33 20


humphd:
I see that `v4.0.0` shipped today, might be good to update to that now: https://github.com/openai/openai-node/releases/tag/v4.0.0

@iramykytyn let me know when you want reviews.  I can wait until CI is passing or start sooner, let me know.

iramykytyn:
> I see that `v4.0.0` shipped today, might be good to update to that now: https://github.com/openai/openai-node/releases/tag/v4.0.0
> 
> @iramykytyn let me know when you want reviews. I can wait until CI is passing or start sooner, let me know.

I need to do some refactoring to let non-streaming function to be working. I will let you know when its ready for review.
Regarding `v4.0.0`, sure I will change it to that version in my next commit.

src/lib/ChatCraftFunction.ts
```diff
@@ -314,11 +315,12 @@ export class ChatCraftFunction {
     };
   }
 
-  toLangChainFunction() {
+  toOpenAIFunction(): OpenAI.Chat.Completions.CompletionCreateParams.Function {
+    const openai_parameters: Record<string, unknown> = this.parameters as Record<string, unknown>;
```
humphd, line 319: prefer lowerCamelCase in JS/TS: `openAIParamaters`

src/lib/ChatCraftMessage.ts
```diff
@@ -98,25 +91,22 @@ export class ChatCraftMessage {
     };
   }
 
-  // Convert to a BaseMessage (or list of BaseMessages in the case of a function)
-  toLangChainMessage(): BaseMessage {
+  toOpenAiMessageJson(): OpenAI.Chat.Completions.CreateChatCompletionRequestMessage {
```
humphd, line 94: I would drop `Json` from the end of this, as you're returning an object literal.  `toOpenAiMessage()` is probably better.

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+  }
+
+  const openai = new OpenAI({
+    apiKey: apiKey,
+    baseURL: apiUrl,
+    defaultHeaders: headers,
+    dangerouslyAllowBrowser: true,
+  });
+
+  const chat_completion_params: OpenAI.Chat.CompletionCreateParams = {
```
humphd, line 156: `chatCompletionParams`

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+          : "auto"
+        : undefined,
+  };
+
+  const chat_completion_req_options = {
+    signal: controller.signal,
+  };
+
+  let response_promise;
+  if (streaming == true) {
```
humphd, line 195: `if (streaming)`

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+     */
+    function_call:
+      model.supportsFunctionCalling && functions
+        ? functionToCall?.name
+          ? { name: functionToCall.name }
+          : "auto"
+        : undefined,
+  };
+
+  const chat_completion_req_options = {
```
humphd, line 190: `chatCompletionReqOptions`

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+
+  let response_promise;
+  if (streaming == true) {
+    response_promise = openai.chat.completions
+      .create(
+        chat_completion_params as OpenAI.Chat.CompletionCreateParamsStreaming,
+        chat_completion_req_options
+      )
+      .then(async (stream_response) => {
+        let function_name: string = "";
```
humphd, line 202: Don't bother assigning a value to these, since you'll throw it away.  Also, `functionName` and `functionArgs`.

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+        throw new Error("unable to handle OpenAI response (not text or function)");
+      })
+      .catch((err) => {
+        return handleError(err);
+      })
+      .finally(() => {
+        removeEventListener("keydown", handleCancel);
+      });
+  } else {
+    response_promise = openai.chat.completions
```
humphd, line 245: `responsePromise`

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+      .create(
+        chat_completion_params as OpenAI.Chat.CompletionCreateParamsNonStreaming,
+        chat_completion_req_options
+      )
+      .then(async (response: OpenAI.Chat.ChatCompletion) => {
```
humphd, line 250: So the streaming/non-streaming cases only differ by this `then()` callback?  Can we move those two functions out into `handleStreamingResponse` and `handleNonStreamingResponse` functions or something and only have a single code path here, picking the right handler at runtime?

src/lib/ai.ts
```diff
@@ -163,105 +124,167 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+      .catch((err) => {
+        return handleError(err);
+      })
+      .finally(() => {
+        removeEventListener("keydown", handleCancel);
+      });
+  }
 
   return {
     // HACK: for some reason, TS thinks this could also be undefined, but I don't see how.
```
humphd, line 286: Does this still happen?  I suspect we don't need to manually type this anymore.

src/lib/db.ts
```diff
@@ -1,9 +1,10 @@
 import Dexie, { Table } from "dexie";
 import { ChatCraftChat, SerializedChatCraftChat } from "./ChatCraftChat";
 
-import type { MessageType } from "langchain/schema";
 import type { FunctionCallParams, FunctionCallResult } from "./ChatCraftMessage";
 
+export type MessageType = "human" | "ai" | "generic" | "system" | "function";
```
humphd, line 6: Can you put this in `ChatCraftMessage.ts` instead and `import` the type here (on line 4 with the other bits from that file)?

humphd (changes requested):
This is a nice code cleanup.  It's looking really good.  I've done a first pass through the code, but not tested it, and made some comments for things to look at.

src/lib/ai.ts
iramykytyn: I was thinking to do it in that way, and it actually was in my first version pushed here, but then I got build issues since it wasn't able to resolve types runtime. I'm not very confident in TypeScript yet, so maybe I just don't see all the ways to implement it. If you could look at the code and give me the hint I will refactor it 🤗

src/lib/db.ts
iramykytyn: done in [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
iramykytyn: Removed [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
iramykytyn: It gives errors if this is not assigned. It says that `Variable 'functionName' is used before being assigned.ts(2454)` 

Screenshot 2023-08-21 at 21 06 13

src/lib/ChatCraftMessage.ts
iramykytyn: done in [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
iramykytyn: done in [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
iramykytyn: done in [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
iramykytyn: done in [22dc6ea](https://github.com/tarasglek/chatcraft.org/pull/220/commits/22dc6ea8eb4cdf1d52b5e3c6ca5813d908dda9d2)

src/lib/ai.ts
humphd: Why don't we go with your way now, and you can file an issue for someone to refactor this in a follow-up.

humphd (changes requested):

src/lib/ai.ts
iramykytyn: I actually did some refactoring. Please take a look at this commit if it looks any better to you [d74c59f](https://github.com/tarasglek/chatcraft.org/pull/220/commits/d74c59fddc71eaf0efcdc72b32c4e3de10e22a19)

iramykytyn:
> Openrouter is not working yet, it returns some CORS error, I'm investigating it:

@humphd @tarasglek can I create a separate task to fix openrouter issue, since it didn't work as expected before this change, so I believe it's better to handle it in the separate task.

src/lib/ai.ts
iramykytyn: Issue created https://github.com/tarasglek/chatcraft.org/issues/227.

src/lib/ai.ts
```diff
@@ -39,45 +37,51 @@ export type ChatOptions = {
...
+function parseOpenAIResponse(response: OpenAI.Chat.ChatCompletion) {
+  let functionName: string = "";
+  let functionArgs: string = "";
+  let content: string = "";
+
+  if (response.choices[0]?.message?.content) {
+    content = response.choices[0]?.message?.content;
+  }
+
+  if (response.choices[0]?.message?.function_call) {
```
tarasglek, line 71: please reduce copy/paste here, put message into a variable..maybe .function_call too

src/lib/ai.ts
```diff
@@ -39,45 +37,51 @@ export type ChatOptions = {
...
-  model,
-}: Pick<ChatOptions, "temperature" | "model"> & { streaming: boolean }) => {
-  const { apiKey, apiUrl } = getSettings();
-  if (!apiKey) {
-    throw new Error("Missing API Key");
+function parseOpenAIChunkResponse(chunk: OpenAI.Chat.ChatCompletionChunk) {
+  let functionName: string = "";
+  let functionArgs: string = "";
+  let token: string = "";
+  if (chunk.choices[0]?.delta?.content) {
```
tarasglek, line 44: same here

src/lib/ai.ts
tarasglek: fixing in followup is ok

src/lib/ai.ts
iramykytyn: done 
617788c

src/lib/ai.ts
iramykytyn: done 
617788c

tarasglek:
This is a big change, solid step forward. Lets land this and fix open router in a followup.
@humphd you ok with that?

tarasglek:
Found a bug where function calls that throw exceptions don't display errors anymore: working example https://chatcraft.org/c/tarasglek/rQou4E50JmPF_usnFwNAI

iramykytyn:
> Found a bug where function calls that throw exceptions don't display errors anymore: working example https://chatcraft.org/c/tarasglek/rQou4E50JmPF_usnFwNAI

This is an interesting case because there actually was no function call, it's just GPT generated text response in the same format as our function call.
Double-checked exception handling  and it works as expected, so no issue here.

tarasglek:
Funny result of my investigation, @humphd would like your ok to land

humphd:
I'm not going to have time to review this (or anything else) for a while, so if you want to wait that's cool; or if you'd rather go ahead and fix in follow-ups, that's fine with me too.

The last time I read this code, my one concern was about how function calls/results were being converted between OpenAI style and our internal representation.  It's possible that everything is good, but I needed more time to dig into the code, SDK, and OpenAI implementation.

If you do land, please squash this so it's easy to back-out or test changes with bisect later on.

PR #222, opened 2023-08-19
I wanna be able to run ES6 modules and TypeScript

tarasglek:
Now I can https://taras-typescript.console-overthinker-dev.pages.dev/c/chatcraft_dev/0AjZxclU4CWxM_n2dG2bK (having share on dev branches rules!)

Needed some esbuild handholding with chatcraft to get here.

As a fun followup would be fun to add a python wasm runtime via something like https://www.npmjs.com/package/python-wasm

### DIFF
--- a/src/lib/run-code.ts
+++ b/src/lib/run-code.ts
@@ -39,7 +39,8 @@ async function runJavascript(code: string) {
           "Chatcraft: Evaling code in a module context, must `export default <your val>` at end to return a value"
         );
       }
-      const module = await import("data:text/javascript;base64," + btoa(code) /* @vite-ignore */);
+      const blob = new Blob([code], { type: "text/javascript" });
+      const module = await import(URL.createObjectURL(blob) /* @vite-ignore */);
       return module.default;
     } else {
       throw error;
--- a/src/lib/run-code.ts
+++ b/src/lib/run-code.ts
@@ -9,10 +9,40 @@ export function isRunnable(language: string) {
   return supportedLanguages.includes(language);
 }
 
-export async function runCode(code: string, language: string) {
-  if (isJS(language)) {
+/**
+ * Run JavaScript code in eval() context, to support returning values from simple expressions `1+1`
+ * and also support `import * as esbuild from 'https://cdn.skypack.dev/esbuild-wasm@0.19.2'` via ES6 modules fallback
+ */
+async function runJS(code: string) {
+  try {
     const fn = new Function(`return eval(${JSON.stringify(code)});`);
     return fn();
+  } catch (error: any) {
+    const msgLower = error.message.toLowerCase();
+    const maybeES6Module =
+      error instanceof SyntaxError &&
+      // Cannot use import statement outside a module
+      // import declarations may only appear at top level of a module
+      msgLower.includes("module") &&
+      msgLower.includes("import");
+    if (maybeES6Module) {
+      // check if code has export.*default regexp
+      if (!/export\s+default\s+/.test(code)) {
+        console.warn(
+          "Chatcraft: Evaling code in a module context, must `export default =` at end to return a value"
+        );
+      }
+      const module = await import("data:text/javascript;base64," + btoa(code));
+      return module.default;
+    } else {
+      throw error;
+    }
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/lib/run-code.ts
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      throw error;
+    }
+  }
+}
+
+async function compileWithEsbuild(tsCode: string) {
+  try {
+    await esbuild.initialize({
+      // no idea how to load the bundled wasm file from esbuild-wasm in vite
```
humphd, line 53: Vite can give you a URL to the `.wasm`, see https://vitejs.dev/guide/assets.html#explicit-url-imports:

```ts
import esbuildWasmUrl from "esbuild-wasm/esbuild.wasm?url";

// ...

wasmUrl: esbuildWasmUrl,
```

src/lib/run-code.ts
```diff
@@ -0,0 +1,78 @@
...
+      wasmURL:
+        "../../node_modules/.pnpm/esbuild-wasm@0.19.2/node_modules/esbuild-wasm/esbuild.wasm",
+    });
+  } catch (error: any) {
```
humphd, line 57: Instead of silently failing on re-init, why not add a boolean variable to the module, and set it to `true` on success the first time?

```ts
let initialized = false;

// ...

if (!initialzied) {
  await esbuild.initialize(...);
  initialized = true;
}
```

src/lib/run-code.ts
```diff
@@ -0,0 +1,78 @@
...
+      msgLower.includes("module") &&
+      msgLower.includes("import");
+    if (maybeES6Module) {
+      // check if code has export.*default regexp
+      if (!/export\s+default\s+/.test(code)) {
+        console.warn(
+          "Chatcraft: Evaling code in a module context, must `export default <your val>` at end to return a value"
+        );
+      }
+      const module = await import("data:text/javascript;base64," + btoa(code) /* @vite-ignore */);
```
humphd, line 42: Data URIs are too limited (can't be more than the size of a URL).  Use a Blob URL, like I do here:

https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ChatCraftFunction.ts#L143-L147

humphd (changes requested):
Some mad science going on here, love it!

src/lib/run-code.ts
tarasglek: I started that way, that still occasionally failed

humphd: Race condition?  Maybe we need a value of `'uninitialized' | 'initializing' | 'initialized' | 'error'` instead of a boolean here.

src/lib/run-code.ts
tarasglek: I don't understand this syntax
```text
Error: Failed to download "http://localhost:5173/c/esbuild-wasm/esbuild.wasm?url"
Error: Failed to download "http://localhost:5173/c/esbuild-wasm/esbuild.wasm?url"
    at blob:http://localhost:5173/6500aae8-5933-47e6-9d39-cd7926378d24:652:19
    at Generator.next (<anonymous>)
    at fulfilled (blob:http://localhost:5173/6500aae8-5933-47e6-9d39-cd7926378d24:9:30)
```

humphd: Vite *should* give you a URL to the file in the bundle. This how people load various static assets from within node_modules (e.g., shaders, worker scripts, etc).

I pulled your branch locally and did this and it worked for me.  Want to push your code so I can see?

src/lib/run-code.ts
tarasglek: interesting theory, just need to stick the promise there

tarasglek: no dice, i even tried to disable esbuild worker. For some reason the loaded wasm persists across page reloads whereas my module-global variable does not

tarasglek:
I can't get the other 2 fixes to work. Do you wanna try?

humphd:
@tarasglek I made a few changes, pushed a commit.  Seems to work for me.  Can you try?

tarasglek:
Wow, that's some next level node-foo. I don't recognize either of these fixes.

PR #223, opened 2023-08-19
WIP: parse typescript modules

tarasglek:
I really want us to use proper typescript with jsdoc comments and generate openai spec from that. Looks like it's a couple of hours of hacking to supplement esbuild with a typescript parser :)


tarasglek:
So one interesting workflow would be:

1. Write me a typescript function histogram to produce a word histogram
2. ai replies with code
3. Show me histogram of this convo using histogram function
4. We run the function..which we parsed metadata from using a typescript parser

tarasglek:
Though what i really wanna do is load multi-function modules without the work of writing out metadata for all of them beyond that of writing jsdoc

src/lib/run-code.ts
```diff
@@ -0,0 +1,80 @@
+import * as esbuild from "esbuild-wasm";
+import * as acorn from "acorn";
+import tsPlugin from "acorn-typescript";
```
humphd, line 3: What are you doing with acorn?

tarasglek: parsing me some ts

tarasglek:
Works now.

Made a hacky little lib with unit tests: https://github.com/tarasglek/typescript2openai

Pretty happy with calling raw typescript as openai functions, sample prompt:
"list functions with @fn-url:https://gist.github.com/tarasglek/40f3ffbeded5375a5618796ad237625f"

tarasglek:
I have no idea how to write and publish node.js packages or setup github actions for unit tests. Open to suggestions on what to do next.

tarasglek:
@humphd can I land this?
This switches our functions over to typescript. Here is a sample:
image

Only thing I can't figure out, is where that `javascript` header is coming from.

package.json
```diff
@@ -39,7 +39,8 @@
     "react-textarea-autosize": "^8.5.2",
     "react-use": "^17.4.0",
     "remark-gfm": "^3.0.1",
-    "tiktoken": "^1.0.10"
+    "tiktoken": "^1.0.10",
+    "typescript2openai": "github:tarasglek/typescript2openai"
```
humphd, line 43: Can you pin this to a commit, like `"typescript2openai": "tarasglek/typescript2openai#gitsha..."` so we control the update and you can land things there without breaking prod?

See https://docs.npmjs.com/cli/v9/configuring-npm/package-json#github-urls

src/lib/ChatCraftFunction.ts
```diff
@@ -139,30 +140,42 @@ export default async function (data) {
...
-  const blob = new Blob([code], { type: "text/javascript" });
+/**
+ *  Use esbuild to parse the code and return a module we can import.
+ */
+const parseModule = async (ts_code: string) => {
+  // strip typescript
+  const js = await toJavaScript(ts_code);
+
+  // pull out function declarations
+  const functionDeclarations = parseTypeScript(ts_code);
```
humphd, line 151: nit: `tsCode`

humphd (changes requested):
Made a few comments.

The 'javascript' title comes from https://github.com/tarasglek/chatcraft.org/blob/main/src/Function/FunctionEditor.tsx#L38.  You should also update the syntax highlighting lib we import at https://github.com/tarasglek/chatcraft.org/blob/main/src/Function/FunctionEditor.tsx#L3 to be typescript and use it at https://github.com/tarasglek/chatcraft.org/blob/main/src/Function/FunctionEditor.tsx#L45

humphd:
@tarasglek can you also update https://github.com/tarasglek/chatcraft.org/blob/df02ffad3380c69ae06722781e845b5bc63a4ae3/src/lib/ChatCraftFunction.ts#L98-L140 so it matches what a function should look like now?  This is what we should when you go to `/f/new`

package.json
tarasglek: it's pinned in pnpm lockfile

humphd (approved):
Epic

tarasglek:
sorry your blog is invalid now :(

humphd:
I'll write another. Are you gonna tweet/blog? I'll wait for you if so

tarasglek:
I wont have time, but i'm hitting the fake function call bug a lot, i think we need to fix that first https://github.com/tarasglek/chatcraft.org/issues/209

humphd:
Is it the "python" function or another?

tarasglek:
Sending our rendered function-calls to openai like
image

causes it to hallucinate our rendered calls in replies..bypassing the actual function call

humphd:
This is what I was discussing in https://github.com/tarasglek/chatcraft.org/pull/220#issuecomment-1694706762: we don't serialize the call/results properly to what OpenAI expects.  I'll file an issue.

PR #225, opened 2023-08-20
Don't recreate toast() in useAlert() on every render

### LINKED ISSUE #224: [UI/UX] Bad user experience when not correct API key entered.
When I entered incorrect API-key (one symbol was missed) I saw a lot of error messages popping up on my screen which eventually hid the API-key input area, so I was not able to correct it without page reloading.

I think that one error message would be enough to be displayed to the user in such a case, I'm not sure why so many identical messages were thrown.

https://github.com/tarasglek/chatcraft.org/assets/6651168/82f738a6-44d3-46d6-8934-f5e92d670c1b

humphd:
Fixes #224 

Thanks for filing this @iramykytyn.  @tarasglek mentioned it, but I couldn't reproduce until I saw your video.

This change fixes it for me locally, can you test and see if it covers your case?

### DIFF
--- a/src/hooks/use-alert.ts
+++ b/src/hooks/use-alert.ts
@@ -1,3 +1,4 @@
+import { useCallback } from "react";
 import { useToast } from "@chakra-ui/react";
 
 type AlertArguments = {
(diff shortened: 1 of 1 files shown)


iramykytyn (approved):
Fixed!

PR #234, opened 2023-08-29
custom endpoints

tarasglek:
We should add a custom openai provider and option to disable streaming.

https://www.anyscale.com/blog/llama-2-is-about-as-factually-accurate-as-gpt-4-for-summaries-and-is-30x-cheaper


tarasglek:
@humphd I thought this code was gonna be impossible, but gpt heled me sort out the ui logic to add a custom entry:

image

What do you think of this?

Need to add an option to disable streaming for 3rd party providers cos it seems to have bugs with almost every non-upstream api

tarasglek:
Yeah i think it would be really nice to point at a free model like fireworks so people could play

tarasglek:
I like the idea of grouping stuff into providers. I think we need to associate keys with a particular provider too. I'll think about how to simplify this weekend if you don't beat me to it

tarasglek:
Ok, so how about a spreadsheet-style ui. Where we have a table of providers


|     | Provider   | Key       |
|:---:|------------|-----------|
|     | OpenAI     |           |
|     | OpenRouter |           |
| ✅ | Anyscale   |           |


only when we set the checkbox, do we update the settings for apiURL/key

tarasglek:
closing this for now as don't have time to do this right

PR #240, opened 2023-09-04
Rework New > Clear to reuse current tab, not lose data

humphd:
This updates how the "Clear" logic works.  We've had users complain of losing work, and the way that "Clear" used to work was part of it.  In the past, "Clear" would delete the messages for the current chat from the db, which also clears them from the UI.  This is obvious if you understand what it does, but not what many expect.

Therefore, we're going to try something different:

- `Clear Chat` - create a new chat in the current tab (i.e., re-use this window).  No data is deleted, previous chat is available in db as before
- `New Blank Chat...` - create a new chat in a new tab

So they do the same thing, but how the tabs work is different.

I haven't removed the `/clear` command, which still deletes the data.  We also have `/new`, so I'm not sure if I need to?


tarasglek:
Seems to work as expected. Lets leave other commands for now as power user features

PR #241, opened 2023-09-06
Audio recording and speech recogntion

### LINKED ISSUE #237: Experiment with voice input
Spin-off from https://github.com/tarasglek/chatcraft.org/issues/194#issuecomment-1660385933. We'd like to explore integrating automatic speech recognition (ASR) into ChatCraft.

I've read the code for many of the existing React components that can do this, and they are all demo quality and largely unmaintained.  Someone might be able to find one that's actually being developed, but if not, we should build our own.

We could use a cloud-based model via API, but we could also consider doing this entirely in the browser using an ONNX or WASM version of Whisper, for example https://github.com/microsoft/onnxruntime-inference-examples/tree/main/js/ort-whisper or https://whisper.ggerganov.com/ (there are others).

If we run a model in the browser, we'd need to figure out some UX for letting the user opt-in to downloading it (i.e., don't do it on load), and giving mic access.

humphd:
Fixes #237 

This is an initial piece of the puzzle for audio recording and speech recognition.  Specifically, it does the following:

- Adds a Mic icon to the inside-right of the prompt text area
- Touching/clicking the Mic icon triggers the logic for audio recording, releasing the icon stops it
- While recording, if you drag to the left, eventually it goes red (indicating you can cancel the in-progress recording).  Releasing when the icon is dragged to the left cancels the current recording.  I've implemented a "spring" animation to make this work.  See what you think (I'm not a pro at animations).
- Adds a new class, `SpeechRecognition` which has hooks for starting/stopping/aborting a recording.

Screenshot 2023-09-06 at 12 46 24 PM

@tarasglek, [over to you!](https://github.com/tarasglek/chatcraft.org/issues/237#issuecomment-1704346905).

Once we know how were doing audio+recognition, we will need to add UI for loading models and/or permissions on first-use.

We shouldn't land this until we have a good first-run story figured out.

### DIFF
--- a/src/components/MicIcon.tsx
+++ b/src/components/MicIcon.tsx
@@ -1,24 +1,24 @@
 import { useState } from "react";
 import { IconButton } from "@chakra-ui/react";
 import { TbMicrophone } from "react-icons/tb";
-import { motion, useSpring } from "framer-motion";
-
-// import { useSettings } from "../hooks/use-settings";
+import { motion, useMotionValue } from "framer-motion";
 
 type MicIconProps = {
   onRecordingStart: () => void;
   onRecordingStop: () => void;
   onRecordingCancel: () => void;
+  isDisabled: boolean;
 };
 
 export default function MicIcon({
   onRecordingStart,
   onRecordingStop,
   onRecordingCancel,
+  isDisabled = false,
 }: MicIconProps) {
-  const spring = useSpring(0, { stiffness: 530, damping: 25 });
   const [colorScheme, setColorScheme] = useState<"blue" | "red">("blue");
   const [isRecording, setIsRecording] = useState(false);
+  const x = useMotionValue(0);
 
   const handleRecordingStart = () => {
     setIsRecording(true);
@@ -55,11 +55,13 @@ export default function MicIcon({
           handleRecordingCancel();
         }
         setColorScheme("blue");
+        x.set(0);
       }}
-      style={{ x: spring }}
+      style={{ x }}
     >
       <IconButton
         isRound
+        isDisabled={isDisabled}
         colorScheme={colorScheme}
         icon={<TbMicrophone />}
         variant={isRecording ? "solid" : "ghost"}
--- a/src/components/MicIcon.tsx
+++ b/src/components/MicIcon.tsx
@@ -1,0 +1,72 @@
+import { useState } from "react";
+import { IconButton } from "@chakra-ui/react";
+import { TbMicrophone } from "react-icons/tb";
+import { motion, useSpring } from "framer-motion";
+
+// import { useSettings } from "../hooks/use-settings";
+
+type MicIconProps = {
+  onRecordingStart: () => void;
+  onRecordingStop: () => void;
+  onRecordingCancel: () => void;
+};
+
+export default function MicIcon({
+  onRecordingStart,
+  onRecordingStop,
+  onRecordingCancel,
+}: MicIconProps) {
+  const spring = useSpring(0, { stiffness: 530, damping: 25 });
+  const [colorScheme, setColorScheme] = useState<"blue" | "red">("blue");
+  const [isRecording, setIsRecording] = useState(false);
+
+  const handleRecordingStart = () => {
+    setIsRecording(true);
+    onRecordingStart();
+  };
+
+  const handleRecordingStop = () => {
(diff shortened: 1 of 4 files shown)


tarasglek:
thanks will take it over in a few days! Exciting

humphd:
https://github.com/xenova/whisper-web

tarasglek:
This works. I would like land it asap, then refine later. It enables really really nice mobile UX, where I now have a very convenient assistant. With functions it should turn into a very nice remote control :)

Currently there are issues:

1. For some reason errors are only reported to console. I have no idea how to hook this up to your toast. I found some seemingly relevant code, but nothing uses it. It seems dead https://github.com/tarasglek/chatcraft.org/blob/issue-237/src/hooks/use-alert.ts#L11 . I think this is the only blocker.
2. Ideally we would only show mic icon if using openai
3. First time it's invoked, it fails as I have to click through permissions. Ideally we would query current mediaRecorder permissions and show one icon for "ask for permissions", "permission to record is granted" and "permission to record denied"
4. Would be really nice to show progress , eg 
   a) show some animation instead of prompt while recording
   b) once recording is being processed show how many bytes and what audio format is being sent to openai

tarasglek:
Demo:
https://github.com/tarasglek/chatcraft.org/assets/857083/5995e3bb-f1cc-4e20-9d86-ceaa1f6dc481

tarasglek:
For some safari reason, code doesn't do anything on iOS safari. I have no idea how to access browser logs there. Works on desktop chrome/firefox/safari and android

tarasglek:
> For some safari reason, code doesn't do anything on iOS safari. I have no idea how to access browser logs there. Works on desktop chrome/firefox/safari and android

Can't repro desktop safari working...Seems that the problem is in that safari only supports audio/mp4..which it encodes in a way that confuses openai, so it can only transcribe first few seconds of audio out of it

tarasglek:
I hate safari and openai..the problem is that the .mp4 file that safari generates is processed wrong by openai..if i pull the aac stream out with ffmpeg, openai processes it file

humphd:
@tarasglek amazing, nice work!  I've tried this here and it works really well.

I spent a bit of time reviewing and fixing your code, then pushed a new commit with changes.  It does the following:

1. Simplifies the Promise logic you had in `SpeechRecognition.start()` and cleans up that file in general
2. Deals with the provider not being OpenAI, hiding the mic icon
3. Fixes a bunch of small paper-cuts in the UX with how the prompt form and mic work
4. Fixes a bug with iOS Safari requiring blob sizes of 1 second
5. Fixes how errors are reported using Toasts vs only in console

I haven't done any other fancy UI stuff, which should probably happen in follow ups.

Test this and see what you think.

tarasglek:
I spent hours on that ios bug, how did you fix it. Can we land?

humphd:
> I spent hours on that ios bug, how did you fix it.

I would never have figured it out without your code + comments and notes.  Based on what you had, I found https://community.openai.com/t/whisper-problem-with-audio-mp4-blobs-from-safari/322252, which led me to https://webkit.org/blog/11353/mediarecorder-api/ and a comment in their example code.

> Can we land? Works great.

If you're happy with it, go ahead.  I haven't tested very much, but I think we can fix whatever breaks in follow-ups.

As an aside, I was interested that you decided to do it all via OpenAI API vs. using an in-browser option.

humphd:
Tested on iOS Safari via preview URL.  One weird thing that happens on first-run is that I have to take my finger off the button when I accept the permission prompt, but then it seems to still be recording?  This first-run experience probably needs some work in a follow-up.

After that it works great.

humphd:
Also, I think the audio recorder/stream might be staying open longer than we expect.  My iOS device shows an orange pill in the top-left, indicating that the page is accessing the mic.

tarasglek:
> Also, I think the audio recorder/stream might be staying open longer than we expect.  My iOS device shows an orange pill in the top-left, indicating that the page is accessing the mic.

Yeah I see that too. At least some of these are browser bugs.

tarasglek:
> 
> As an aside, I was interested that you decided to do it all via OpenAI API vs. using an in-browser option.

Openai is a quick way to ship. Whisper in browser has a problem because openai uses the maximum size model. ..browser would have to use a smaller one which reduces recognition quality

PR #242, opened 2023-09-07
Update deps for Sep 7, 2023

humphd:
Updates to deps, nothing seems problematic in my (light) testing.

Prettier now wants to do the DOCTYPE differently, so I've fixed that, and one of the types for OpenAI changed names.


tarasglek:
feel free to land this

humphd:
I've rebased and update again for Sep 18.  All seems good except for https://github.com/remarkjs/remark-gfm/issues/57, which I've left where it is for now, until the rest of the related packages are updated.

humphd:
Something seems broken with the LLM responses in this.  I'm going to wait on this until I can debug more.  Let's not land this yet.

humphd:
https://github.com/tarasglek/chatcraft.org/pull/251 fixed the regression in LLM responses, I think this is good to go.

PR #247, opened 2023-09-15
Improve first-run experience for mic

### LINKED ISSUE #245: Audio first-run state issue
I think I've understood more about what's going wrong with the first-run experience for audio.

We are trying to do a press-and-hold UX on the mic icon.  However, the user is forced to take their finger off the icon when they are shown the browser's permission UI to allow/deny the microphone access.

Because this happens outside the DOM (e.g., browser chrome), I don't think it's triggering an event to indicate that we've released the icon.

We need to do the following:

1. User presses the mic icon
2. We check for, or try to acquire permission to the mic
3. We succeed or fail
    i. If we fail, we need to cancel the recording action
    ii. if we succeed, we need to check if the active element in the DOM is still the mic icon button.  If it is, we continue (user is still pressing it, which happens when this isn't the first run); if not, we need to cancel it

## Note from GPT-4 on ChatCraft:

Yes, you can use the `document.activeElement` property to check if the microphone button is currently being pressed. This property returns the element in the document that currently has focus. If the microphone button is being pressed, it should have focus.

Here's how you can modify your `onRecordingStart` function to check if the microphone button is still being pressed after requesting microphone permissions:

```tsx
const onRecordingStart = async () => {
  try {
    // Request microphone permissions
    await navigator.mediaDevices.getUserMedia({ audio: true });

    // Check if the microphone button is still being pressed
    if (document.activeElement !== microphoneButtonRef.current) {
      // The microphone button is no longer being pressed
      // You may want to return here to avoid starting the recording
      return;
    }

    // Start recording
    speechRecognitionRef.current = new SpeechRecognition();
    await speechRecognitionRef.current.start();
    setIsListening(true);
  } catch (err: any) {
    console.error(err);
    error({
      title: "Speech Recognition Error",
      message: `Unable to start audio recording: ${err.message}`,
    });
  }
};
```

In this code, `microphoneButtonRef` is a React ref that refers to the microphone button. You can create this ref and attach it to the microphone button like this:

```tsx
const microphoneButtonRef = useRef(null);

// ...

<IconButton
  // ...
  ref={microphoneButtonRef}
/>
```

Please note that this code assumes that the microphone button will have focus when it is being pressed. This should be the case if the button is a focusable element (like a button or a link), but it may not be the case for other types of elements.

humphd:
Fixes #245

This reworks the state machine for using the mic.  The main change is that I've separated the part where you request access to the audio stream from the start of recording.  This allows for extra checks to happen in the UI when the user takes their finger off the mic icon in order to press the "allow" button in the permissions UI.

I've tested in every browser I have on my Mac, and it's working pretty well.

I don't think I know how to make this perfect, but I think this is about as good as I can get it.  See if you agree.

### DIFF
--- a/src/components/MicIcon.tsx
+++ b/src/components/MicIcon.tsx
@@ -1,38 +1,103 @@
-import { useState } from "react";
+import { useState, useRef } from "react";
 import { IconButton } from "@chakra-ui/react";
 import { TbMicrophone } from "react-icons/tb";
 import { motion, useMotionValue } from "framer-motion";
 
+import { SpeechRecognition } from "../lib/speech-recognition";
+import { useAlert } from "../hooks/use-alert";
+
 type MicIconProps = {
-  onRecordingStart: () => void;
-  onRecordingStop: () => void;
-  onRecordingCancel: () => void;
+  onRecording: () => void;
+  onTranscribing: () => void;
+  onCancel: () => void;
+  onTranscriptionAvailable: (transcription: string) => void;
   isDisabled: boolean;
 };
 
 export default function MicIcon({
-  onRecordingStart,
-  onRecordingStop,
-  onRecordingCancel,
+  onRecording,
+  onTranscribing,
+  onCancel,
+  onTranscriptionAvailable,
   isDisabled = false,
 }: MicIconProps) {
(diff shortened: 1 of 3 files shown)


tarasglek:
Sorry been busy here, completely forgot about github for a few days

PR #248, opened 2023-09-15
Allow re-sending final message if human

humphd:
A pattern I get myself into a lot:

- Get into a long chat
- Go back and delete a bunch of things, mostly responses
- Update something in the final human prompt
- Want to retry this final prompt

The way I do it now is to copy the prompt, delete it, then paste into the prompt form and send.

This change allows clicking the "Ask" button with nothing in the prompt.  If the final message is a human message, it will send the chat as-is vs. adding anything new.  This is perfect for cases where you have edited and want to retry.

If the final message isn't a human message, there's no point re-sending without adding context, since the AI will usually just repeat itself.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -182,17 +182,25 @@ function ChatBase({ chat }: ChatBaseProps) {
       }
 
       // Not a slash command, so pass this prompt to LLM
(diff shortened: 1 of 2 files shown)


tarasglek:
it's a hack, but a useful one :)

PR #250, opened 2023-09-17
Experiment with summary methods

humphd:
This reworks the way the summaries are generated to use a "Term Frequency-Inverse Document Frequency" algorithm. I'm also stripping the markdown and using some NLP deps for better tokenization. I added a few tests to make development easier.

This works better than what we have now.


tarasglek:
I'm ok with this, but why not use gpt to summarize convo?

humphd:
> I'm ok with this, but why not use gpt to summarize convo?

Two reasons:

1. I need it to happen synchronously (e.g., sidebar needs something to show immediately)
2. It happens every time the chat gets updated, which is frequent and would cost a lot

When a user shares a chat, they can use gpt to summarize if they want.  This is just a placeholder.

PR #253, opened 2023-09-20
Don't call stop() if MediaRecorder state isn't inactive

### LINKED ISSUE #252: Can't record on android

humphd:
Fixes #252

I've added some extra checks around when we call `.stop()` on a `MediaRecorder`, since [it will error if the `state` is already `inactive`](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/stop#errors).

I guess some browsers don't send the error and Android must.


tarasglek:
my android device is away for a day, will test this once it's back

PR #255, opened 2023-09-21
Rework PromptForm into Desktop and Mobile versions

### LINKED ISSUE #50: Don't hide content on mobile
From https://github.com/tarasglek/chatcraft.org/pull/46#issuecomment-1541592561:

> We have a general ui problem where on mobile devices with kb open we can't see content at all

We might need to rework the prompt area for small screens.  Maybe switch away from `textarea` to `input`, move buttons inline vs below, drop the hint area at the top.

humphd:
I'm starting to experiment with a better prompt form, and first up is the mobile experience.  I've been looking at how various mobile apps do this (Messages on iOS, Telegram, WhatsApp) and have tried to create something more appropriate for use on a small screen with your fingers:

Screenshot 2023-09-20 at 9 15 01 PM

A few things to call out:

- The form now takes up almost no height (fixes #50).  The input element has been greatly reduced in size.  It will expand vertically if the text requires it, but I think audio is the ideal way to enter a prompt on mobile.
- Single Message Mode has been removed
- Input expansion and keyboard hints have been removed
- The token cost for the session has been moved to the header for the chat (I'm not sure about this, there there is no good place for it):

Screenshot 2023-09-21 at 7 49 09 AM

- The "New +" button has been changed to "+" and moved to the left of the input.  Clicking it brings up the same menu of options:

Screenshot 2023-09-21 at 7 50 05 AM

7. The "Ask" button/model chooser has been separated into two buttons.  The up-arrow lets you choose a model, and the "Ask" button has been shortened to a "send" arrow.

Screenshot 2023-09-21 at 7 50 33 AM

9. The Mic button has been moved out of the input element so that attempting to press it doesn't accidentally activate/focus the input box on mobile.  It works the same as before (hold to record, slide-left to cancel).
10. While recording/transcribing, the input element is replaced with a status indicator showing the progress of the recording or transcription:

Screenshot 2023-09-21 at 7 45 17 AM

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -28,8 +28,6 @@ function ChatBase({ chat }: ChatBaseProps) {
   const { error: apiError } = useModels();
   // When chatting with OpenAI, a streaming message is returned during loading
   const { streamingMessage, callChatApi, cancel, paused, resume, togglePause } = useChatOpenAI();
-  // Whether to include the whole message chat history or just the last response
-  const [singleMessageMode, setSingleMessageMode] = useState(false);
   const { isOpen: isExpanded, onToggle: toggleExpanded } = useDisclosure();
   const { settings, setSettings } = useSettings();
   const { isOpen: isSidebarVisible, onToggle: toggleSidebarVisible } = useDisclosure({
@@ -200,11 +198,6 @@ function ChatBase({ chat }: ChatBaseProps) {
           }
         }
 
-        // In single-message-mode, trim messages to last few. Otherwise send all.
-        // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
-        const messages = chat.messages({ includeAppMessages: false });
-        const messagesToSend = singleMessageMode ? [...messages].slice(-2) : messages;
-
         // If there's any problem loading referenced functions, show an error
         const onError = (err: Error) => {
           error({
@@ -226,7 +219,9 @@ function ChatBase({ chat }: ChatBaseProps) {
           }
         }
 
-        const response = await callChatApi(messagesToSend, {
+        // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
+        const messages = chat.messages({ includeAppMessages: false });
+        const response = await callChatApi(messages, {
           functions,
           functionToCall,
         });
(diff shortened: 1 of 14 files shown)


humphd:
The "send" and "model chooser" could also get combined into something like Slack uses vs. splitting them out into two separate buttons:

Screenshot 2023-09-21 at 10 12 13 AM

The down side of this is the round, round, square look.  However, I could also "ghost" the + and mic buttons.

Just thinking out loud.

tarasglek:
Sexy

tarasglek:
I would think this ui would be ok for desktop too. Not sure why we have to fork em

tarasglek:
I would just keep text on buttons on desktop ver view responsive layout

humphd:
OK, new updates on Mobile and a total redesign of Desktop.

I've done a ton of small tweaks to things, and I think I'm pretty happy with both of these.

Some notable things you can't see unless you play with it:

- On desktop, I got rid of the extra space above the input element where you could expand and various hints were placed.  The expansion is done on-demand as you press enter vs. doing it via a button.  The hints have moved to either be inline (recording) or down near the send button
- When recording, we now hide the input element and show a "Recording 00:01..." message that updates in realtime.
- Single Message Mode is gone from both desktop and mobile
- Cost info is moved to the chat header for both desktop and mobile

## Desktop

Screenshot 2023-09-26 at 2 42 25 PM

## Mobile

Screenshot 2023-09-26 at 2 42 39 PM

tarasglek:
Good-bye single-msg mode! This feels good.

This is good thought we lost multi-line mode, not sure if it matters.

Only remaining mobile issue is that pause/cancels buttons are not big enough for mobile..might make sense to replace input box with them to conserve vertical space?

humphd:
> This is good thought we lost multi-line mode, not sure if it matters.

What does "multi-line mode" mean?  If you type and press enter, it will expand to multiple lines.  Or do you mean the chevron you could click to expand it full-height without typing?

> Only remaining mobile issue is that pause/cancels buttons are not big enough for mobile..might make sense to replace input box with them to conserve vertical space?

By play/cancel, you mean the mic button?  How much bigger should it be?

tarasglek:
> > This is good thought we lost multi-line mode, not sure if it matters.
> 
> What does "multi-line mode" mean? If you type and press enter, it will expand to multiple lines. Or do you mean the chevron you could click to expand it full-height without typing?

sorry, i meant single-message mode. Just got off a long flight, bit mentally weak :)
> 
> > Only remaining mobile issue is that pause/cancels buttons are not big enough for mobile..might make sense to replace input box with them to conserve vertical space?
> 
> By play/cancel, you mean the mic button? How much bigger should it be?

image
these buttons

PR #256, opened 2023-09-27
Fix DesktopPromptForm text area resize bug

humphd:
One of my daughters has already found a bug in my new DesktopPromptForm :( , where multi-line content pushes the content down vs. up:

Screenshot 2023-09-26 at 8 47 02 PM

This fixes it so the prompt container flexes upward instead, when the textarea resizes.


humphd:
r+=my daughter :)

PR #258, opened 2023-09-27
Reduce bundle size ~50% (esbuild, dexie-import-export, lodash, Function, ts parsing)

humphd:
This change improves our main bundle size, dropping it almost 1M.  It moves dependencies we don't need in all cases to use dynamic imports, which Vite can then move outside the main bundle.  These include:

- esbuild
- lodash
- dexie-export-import

Everything works the same, but the browser loads this on-demand at runtime instead of up-front.


humphd:
Actually, I can drop the main bundle size by half!  I've added two more improvements:

- Lazily load the `<Function />` component, which isn't used very often, and has some big deps for CodeMirror
- dynamically import your typescript2openai, which isn't used often and has large deps too.

tarasglek:
Nice, do we keep stats on bundle size anywhere?

Would be fun to graph this over time.

humphd:
When we do `pnpm build`, it shows sizes for each of the bundles it creates, and the one that matters the most is the main `index.js` bundle that contains the core app.  In this PR, I added a new vite plugin, [rollup-plugin-visualizer](https://github.com/btd/rollup-plugin-visualizer#rollup-plugin-visualizer), which lets you get a visualization of what is in each bundle.  So now a new `stats.html` gets created in the root, and you can use that to inspect things.

Most of the time I don't think we'll care, but every once in a while, after we add new deps, it's useful to do some work to optimize.  What I did today took 2M off of our initial page load, which is pretty amazing.

PR #261, opened 2023-09-29
Add margin to top of HTML preview, fix #203

### LINKED ISSUE #203: Fix z-index/overlap in iframe HTML examples
When an example HTML page includes content in the top-right, it can cause issues with the "open in fullscreen" icon button:

Screenshot 2023-08-10 at 11 26 38 AM

It would be good to find a way to add enough padding around this icon button so that it doesn't get lost.

humphd:
Added a margin to the top of the web content so it doesn't interfere.

Screenshot 2023-09-29 at 12 03 05 PM


humphd:
Fixes #203

PR #262, opened 2023-09-29
Update deps for Sept 29, 2023, fix breaking changes in react-markdown

humphd:
Updates all our deps.  I had to do some surgery to get the `react-markdown` update to work in our code.  They've changed their API, removed a bunch of properties and forced the use of some plugins.

Things seem OK in my testing now (Mermaid diagrams, HTML previews, inline code rendering, GFM, etc).


tarasglek:
ok to land?

humphd:
Let's try.  If something breaks, we'll know who to suspect...

PR #265, opened 2023-10-04
Set prompt form focus when navigating

### LINKED ISSUE #264: Set prompt form focus when creating new chat
Currently if you create a new Chat (e.g., from the `new` button in the sidebar), the newly created chat does NOT put the focus on the prompt form.

humphd:
Fixes #264.

When the user navigates (e.g., creates a new chat), this will re-focus the prompt input form.  I'm only doing this on desktop, since we don't want to clutter the viewport on mobile.


humphd:
r+=daughter

Issue #205, opened 2023-08-10
Size preview iframes to fit contents


humphd:
When we create preview iframes, we always use the same height.  It would be cool if we could tell the iframe to resize to fit its contents (to a maximum threshold).  There are some libraries to do this, which require you to inject some JS into the document within the iframe and then send a message back to the host.  Not trivial, but doable.

https://chatcraft.org/c/humphd/AsI25XEORVC57AKsPI-rF is a good example of where I want it.

humphd:
https://github.com/davidjbradshaw/iframe-resizer and https://github.com/davidjbradshaw/iframe-resizer-react are interesting.

humphd:
Fixed in #267

PR #267, opened 2023-10-20
Add - Size preview iframes to fit contents

mingming-ma:
Hi, I would like to add this feature that addressed in #205 that supports size preview iframes to fit contents. Let me know about any changes. Thanks a lot!

## Change description:

For the host page settings. I use the `iframe-resizer-react`'s `IframeResizer` component generates `iframe` and give the prop `src` by the children html content's url.
For the iframe page settings, because need to include the ([iframeResizer.contentWindow.min.js](https://raw.github.com/davidjbradshaw/iframe-resizer/master/js/iframeResizer.contentWindow.min.js)) from [iframe-resizer](https://github.com/davidjbradshaw/iframe-resizer) as [documented](https://github.com/davidjbradshaw/iframe-resizer-react#usage), I use `DOMParser` to add `script` tag with the `cdnjs` version `iframeResizer.contentWindow.min.js` to achieve the requirements.
- [x] upload test screenshots: https://github.com/tarasglek/chatcraft.org/assets/133393905/49bb9503-69dd-4b59-a24a-249ecb4e19b7


src/components/HtmlPreview.tsx
```diff
@@ -29,7 +39,13 @@ const HtmlPreview = ({ children }: HtmlPreviewProps) => {
         variant="ghost"
       />
       <CardBody mt={6} p={2}>
-        <chakra.iframe w="100%" minHeight="200px" frameBorder="0" src={url}></chakra.iframe>
+        <IframeResizer
+          maxHeight={1200}
```
humphd, line 43: `1200` seems arbitrary.  Is there a reason you're capping it vs. letting it expand to infinity?  I'm trying to imagine a situation where the LLM writes HTML that produces content too long to be usable.

src/components/HtmlPreview.tsx
```diff
@@ -29,7 +39,13 @@ const HtmlPreview = ({ children }: HtmlPreviewProps) => {
...
+          checkOrigin={false}
```
humphd, line 44: Why is this disabled?

src/components/HtmlPreview.tsx
```diff
@@ -29,7 +39,13 @@ const HtmlPreview = ({ children }: HtmlPreviewProps) => {
...
+          src={url}
+          width="100%"
```
humphd, line 46: Why not use their recommended `width` setup https://www.npmjs.com/package/iframe-resizer-react#typical-setup?

humphd (changes requested):
@mingming-ma this is amazing, thank you!  I've wanted this feature for so long, but never got around to implementing it.  I'm glad you took up the challenge.

I've tested it locally, and it's working well in my limited tests.

I've left some feedback and questions.  Welcome to the ChatCraft dev team :)

src/components/HtmlPreview.tsx
mingming-ma: Yeah, the [minHeight](https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/parent_page/options.md#maxheight--maxwidth) not say but by my test it seems use px, and I guess 1200 should be good for most situation, and if more than that, it will goes to scroll mode.

src/components/HtmlPreview.tsx
mingming-ma: This is due to the cdn js, so it will shows warning if true

src/components/HtmlPreview.tsx
mingming-ma: Oh, I didn't notice the [example](https://github.com/davidjbradshaw/iframe-resizer-react/blob/master/example/src/app.jsx) uses the different settings. I guess it will works also by the recommended `width` setup. Let me try that.

src/components/HtmlPreview.tsx
humphd: This shouldn't be necessary now without CDN, correct?  Let's remove.

humphd (changes requested):

mingming-ma:
@humphd I've made some changes:

- [x] use local js file rather than cdn
- [x] rename variable to `html`
- [x] remove height capping
- [ ] `checkOrigin={false}`
  - This is needed to keep even not using cdn js. After checking I guess the reason is our `iframe`' `src` will starts with` blob:http://localhost`, while the `<script> src` will starts with`http://localhost`
(screenshots: https://github.com/tarasglek/chatcraft.org/assets/133393905/4a8e7dba-37d9-4f7b-a2c4-8959b293d642)
- [x] use recommended width setup

Let me know if there are any other changes needed. Thanks a lot!

src/components/HtmlPreview.tsx
mingming-ma: After trying I think we still need this 😢

humphd: OK, your explanation makes sense.

src/components/HtmlPreview.tsx
```diff
@@ -29,7 +38,12 @@ const HtmlPreview = ({ children }: HtmlPreviewProps) => {
         variant="ghost"
       />
       <CardBody mt={6} p={2}>
-        <chakra.iframe w="100%" minHeight="200px" frameBorder="0" src={url}></chakra.iframe>
+        <IframeResizer
+          checkOrigin={false}
+          src={url}
+          style={{ width: "1px", minWidth: "100%" }}
+          scrolling={true}
```
humphd, line 45: Is this necessary if we don't have a height set?

humphd (approved):

src/components/HtmlPreview.tsx
mingming-ma: Yeah, should be ok to the content height 😄

humphd:
Really nice work @mingming-ma, great job!  Can't wait to use this.  I've wanted it for a long time.

humphd:
Live and working on https://chatcraft.org!:

Screenshot 2023-10-20 at 9 07 24 PM

@mingming-ma if you'd like to fix another bug, we should disable updates during `loading`, because it flickers so much and jumps the page.

mingming-ma:
@humphd  Thank you so much for your encouragement! I will definitely try to fix that!

tarasglek:
@mingming-ma congrats on getting your first contribution in, thank you!

mingming-ma:
@tarasglek Thank you! More to come!

PR #269, opened 2023-10-22
Lazily load the IframeResizer only when needed

humphd:
@mingming-ma I realized that we can save some download size in the bundle by not loading this until needed (we don't show previews that often), using React's `lazy` function, see https://react.dev/reference/react/lazy.

Can you test this and make sure it still works OK for you?


mingming-ma:
@humphd I tested and it still works on my side! 👍

humphd:
Thanks for testing, I'll land this.

PR #272, opened 2023-10-24
Refactor: Add an isLoading check guard for optimization.

mingming-ma:
## Description

This change updates the logic of handling `children` in the `url` function of `HtmlPreview`, when `isLoading` is `true`, skip the parse process, which improves performance.

### DIFF
--- a/src/components/HtmlPreview.tsx
+++ b/src/components/HtmlPreview.tsx
@@ -6,20 +6,25 @@ const IframeResizer = lazy(() => import("iframe-resizer-react"));
 
 type HtmlPreviewProps = {
   children: ReactNode & ReactNode[];
+  isLoading: boolean;
 };
 
 const toUrl = (html: string) => URL.createObjectURL(new Blob([html], { type: "text/html" }));
 
-const HtmlPreview = ({ children }: HtmlPreviewProps) => {
+const HtmlPreview = ({ children, isLoading = false }: HtmlPreviewProps) => {
   const url = useMemo(() => {
-    const parser = new DOMParser();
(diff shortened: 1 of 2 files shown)


src/components/HtmlPreview.tsx
```diff
@@ -6,20 +6,25 @@ const IframeResizer = lazy(() => import("iframe-resizer-react"));
...
-    const parser = new DOMParser();
-    const doc = parser.parseFromString(String(children), "text/html");
-    const scriptElement = document.createElement("script");
-    scriptElement.src = new URL("/js/iframeResizer.contentWindow.min.js", import.meta.url).href;
-    doc.body.appendChild(scriptElement);
-    const html = `<!DOCTYPE html>${doc.documentElement.innerHTML}`;
-    return toUrl(html);
-  }, [children]);
+    if (isLoading) {
+      return "about:blank";
```
humphd, line 17: Clever

src/components/HtmlPreview.tsx
```diff
@@ -6,20 +6,25 @@ const IframeResizer = lazy(() => import("iframe-resizer-react"));
...
+    } else {
```
humphd, line 18: Don't use `else` after a `return.  Do this:

```ts
if (isLoading) {
  return "about:blank";
}

// rest of code here...
```

humphd (changes requested):
Looks good.  One nit to fix.

humphd (approved):

humphd:
@mingming-ma you're on a roll, great work on this and thank you!

mingming-ma:
Thanks! I'm very glad to contribute 😄

PR #278, opened 2023-11-08
Add 'Save & Submit' button

### LINKED ISSUE #277: Ability to change and resubmit previous prompts
When I change the last prompt and save, the prompt is not resubmitted. I kind of expect the button to resubmit the prompt and continue the conversation from there, like chatgpt.

Is this the intended behaviour or can we have a 'Save & Submit' button?

DukeManh:
Fixes #277

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -31,6 +31,8 @@ import {
   Textarea,
   VStack,
   useClipboard,
+  Kbd,
+  Spacer,
 } from "@chakra-ui/react";
 import ResizeTextarea from "react-textarea-autosize";
 import { TbDots, TbTrash } from "react-icons/tb";
@@ -349,25 +351,34 @@ function MessageBase({
                       defaultValue={text}
                       autoFocus={true}
                     />
-                    <ButtonGroup>
-                      <Button size="sm" variant="outline" onClick={() => onEditingChange(false)}>
+                    <Flex width="100%" alignItems="center" alignContent="end" gap={2}>
+                      <Spacer />
+                      <Text fontSize="sm" color="gray">
+                        <span>
+                          <Kbd>Ctrl</Kbd> + <Kbd>Enter</Kbd>
+                          <span> to save</span>
+                        </span>
+                      </Text>
+                      <Button size="sm" variant="ghost" onClick={() => onEditingChange(false)}>
                         Cancel
                       </Button>
-                      {message instanceof ChatCraftHumanMessage && (
-                        <Button
-                          size="sm"
-                          type="submit"
-                          name="resubmit"
-                          colorScheme="cyan"
-                          onClick={handleClick}
-                        >
-                          Save & Submit
+                      <ButtonGroup>
+                        {message instanceof ChatCraftHumanMessage && (
+                          <Button
+                            size="sm"
+                            variant="outline"
+                            type="submit"
+                            name="resubmit"
+                            onClick={handleClick}
+                          >
+                            Re-Ask
+                          </Button>
+                        )}
+                        <Button size="sm" type="submit" name="save" onClick={handleClick}>
+                          Save
                         </Button>
-                      )}
-                      <Button size="sm" type="submit" name="save" onClick={handleClick}>
-                        Save
-                      </Button>
-                    </ButtonGroup>
+                      </ButtonGroup>
+                    </Flex>
                   </VStack>
                 </form>
               ) : (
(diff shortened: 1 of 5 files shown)


humphd:
@DukeManh let me know when you want reviews on this.

DukeManh:
@humphd, can you review it please.

src/components/MessagesView.tsx
```diff
@@ -97,6 +97,10 @@ function MessagesView({
           message={message}
           chatId={chatId}
           isLoading={isLoading}
+          onResubmitClick={async () => {
+            await deleteMessages(message.id, "after");
```
humphd, line 101: This is surprising.  Why does editing a message and then resubmitting delete anything?

humphd (changes requested):
This is looking good.  I have one concern about data loss.

Also, this UI should probably use different colours for Save and Save & Submit:

Screenshot 2023-11-09 at 9 19 53 AM

I think right-most should be CTA and be the primary colour, next one to left should be secondary, and cancel can stay outline or become ghost.

src/components/MessagesView.tsx
DukeManh: So I thought how it works is when you edit a message, the conversation starts again from the editted message similar to chatgpt. Do you have a different idea of how it should work?

DukeManh:
I changed the 'Save & submit' color, let me know if it looks fine. Since 'Save' is the default action, I moved it to right most.

humphd (changes requested):
I think this is great.  I have a few things I'd like to see before we land it:

1. Let's rename "Save & Resubmit" to something that better communicates what's going to happen.  The reality is that we're not "Saving" this any more, but we're sort of forking the chat from this point and throwing away the messages that follow.  What about "Re-Ask" or "Save & Restart Chat" or "Save & Continue"?  None of these is perfect, but i want to communicate to the user that this will cause a new thing to happen beyond simply updating the text.
2. Would you mind borrowing the code from the prompt form to advertise that `CTRL + Enter` will save (could happen in follow-up)?  To make room for it, what if we put `Cancel` all the way on the left, and have the two action buttons on the right.

Screenshot 2023-11-10 at 7 45 52 AM

3. Let's make `Cancel` use the `ghost` variant, the new button be `outline` and the primary button be `Save`

tarasglek:
@DukeManh I invited you to the repo for better CI/CD

DukeManh:
> What about "Re-Ask" or "Save & Restart Chat" or "Save & Continue"?

I like 'Re-Ask', it sounds more precise.

For the the placement of the keyboard hint, I think having the 'Cancel' button on the far left is a bit inconvenient. What about putting it besides button group?

Current
Screenshot 2023-11-12 at 8 28 39 PM
Bottom
Screenshot 2023-11-12 at 8 21 47 PM
Middle
Screenshot 2023-11-12 at 8 20 23 PM

src/components/Message/MessageBase.tsx
```diff
@@ -116,11 +132,29 @@ function MessageBase({
...
+    messageForm.current?.setAttribute("data-action", e.currentTarget.name);
+  }, []);
+
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent) => {
+      if (e.ctrlKey && (e.key === "Enter" || e.key === "NumpadEnter")) {
+        const submitEvent = new Event("submit", { cancelable: true, bubbles: true });
+        messageForm.current?.dispatchEvent(submitEvent);
+      }
+      if (e.key === "Escape") {
```
humphd, line 145: This is great to have too, nice work.

src/components/Message/MessageBase.tsx
```diff
@@ -307,14 +351,34 @@ function MessageBase({
                       defaultValue={text}
                       autoFocus={true}
                     />
-                    <ButtonGroup>
-                      <Button size="sm" variant="outline" onClick={() => onEditingChange(false)}>
+                    <Flex width="100%" alignItems="center" alignContent="end" gap={2}>
+                      <Spacer />
+                      <Text fontSize="sm" color="gray">
+                        <span>
+                          <Kbd>Ctrl</Kbd> + <Kbd>Enter</Kbd>
```
humphd, line 358: This will be CMD on Mac.  I think I solved this already in the other part of the code that does this, maybe we can extract a reusable component (could be done in follow-up)?

humphd (approved):
I agree with your current placement.  I made one comment, but I think this is good.

Amazing work, Duke!

src/components/Message/MessageBase.tsx
DukeManh: I will do it in this PR or we might forget about it.
So because we are editing a prompt textarea, I think the default behaviour of Enter is new line and Ctr/cmd + enter is to submit. I don't think it shares the same behaviours with the new prompt such that that we can extract a resusable component.

humphd (approved):
Excellent work.  Please squash-and-merge when you're ready.

Issue #279, opened 2023-11-08
Figure out ChatCraft use cases for new OpenAI API features


humphd:
OpenAI just had a DevDay, and announced a bunch of cool new features, see https://openai.com/blog/new-models-and-developer-products-announced-at-devday.

- larger context window for GPT-4 (128K).  This would let you stuff a bunch of files into the prompt
- Improvements to Function Calling (multiple functions at once, JSON mode)
- Visual Input support for GPT-4.  Could use this to allow talking about screenshots or visual output and turning that into code (Make React components that look like this...)
- Image generation

Not sure if there are people who want to try hacking on some of these, but I thought I'd call attention to it.  I probably won't have time to do it any time soon, but could review.

Here's the email I got yesterday:

-----------

We're excited to share major new features and updates that were announced at our first conference, OpenAI  DevDay. You can read the full details on our blog, watch the keynote recording, or check out the new @OpenAIDevs Twitter, but here a brief summary:
New GPT-4 Turbo:
We announced GPT-4 Turbo, our most advanced model. It offers a 128K context window and knowledge of world events up to April 2023. 
We’ve reduced pricing for GPT-4 Turbo considerably: input tokens are now priced at $0.01/1K and output tokens at $0.03/1K, making it 3x and 2x cheaper respectively compared to the previous GPT-4 pricing. 
We’ve improved function calling, including the ability to call multiple functions in a single message, to always return valid functions with JSON mode, and improved accuracy on returning the right function parameters.
Model outputs are more deterministic with our new reproducible outputs beta feature.
You can access GPT-4 Turbo by passing gpt-4-1106-preview in the API, with a stable production-ready model release planned later this year.

Updated GPT-3.5 Turbo:
The new gpt-3.5-turbo-1106 supports 16K context by default and that 4x longer context is available at lower prices: $0.001/1K input, $0.002/1K output. Fine-tuning of this 16K model is available. 
Fine-tuned GPT-3.5 is much cheaper to use: with input token prices decreasing by 75% to $0.003/1K and output token prices by 62% to $0.006/1K.
gpt-3.5-turbo-1106 joins GPT-4 Turbo with improved function calling and reproducible outputs.

Assistants API:
We're excited to introduce the beta of our new Assistants API, designed to help you build agent-like experiences in your applications effortlessly. Use cases range from a natural language-based data analysis app, a coding assistant, an AI-powered vacation planner, a voice-controlled DJ, a smart visual canvas—the list goes on
[...]

PR #281, opened 2023-11-09
Add - Support visual input for the gpt-4-vision-preview

mingming-ma:
Hi, as #279 discuses, the `gpt-4-vision-preview` now allow us to ask questions related images.  I would like to help to update the code support visual Input. Because it related to many components, I was trying to make minimum possible updates to show the workable and share my process now. Hope my approach is on the right way.

## Tasks

- [ ] Fix response cutoff bug
- [ ] Update the model based on file type
- [ ] Attach file instead of drag file
- [ ] Better preview if attach file is image

## Status
Now user can drag the image to text area and ask questions, like below:
<details>

Before submit:
image

After submit:
image
</details>

### response cutoff bug
image

### type mismatch
New payload format not meet the interface type, trying to fix
image

Feel free to share your feedback! Thanks!

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -141,7 +141,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string, image_url?: string[]) => {
+    async (prompt?: string, image?: string[]) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
@@ -184,7 +184,7 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, image_url, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
           await chat.addMessage(promptMessage);
         } else {
           // If there isn't any prompt text, see if the final message in the chat was a human
@@ -220,7 +220,6 @@ function ChatBase({ chat }: ChatBaseProps) {
 
         // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
         const messages = chat.messages({ includeAppMessages: false });
-        console.log(messages);
         const response = await callChatApi(messages, {
           functions,
           functionToCall,
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -141,7 +141,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string) => {
+    async (prompt?: string, image_url?: string[]) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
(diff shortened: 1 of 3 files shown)


humphd:
Alright, this is amazing.  @mingming-ma thank you for jumping on this.

I tried this out today, and here are things I noticed and thoughts on things we should do (some in this PR, or as follow-ups).

1. When we drag-and-drop an image into the prompt, we should automatically switch the current model to be a model that supports vision.  To do that we should probably add something like `supportsImages()` similar to https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ChatCraftModel.ts#L51.  I'm not sure what we should do when the user goes back to using text only.  We might need a way to keep track of more than one model in `localStorage`, so we can remember what you were using before we switched you to the new model.  This could get done in a follow-up.
2. We need a way to make it clear you can "add" or "attach" files to the current prompt.  I've wanted to do this for a while, actually.  It should work for files (here's the source code I'm talking about...), PDFs (pull the text out and let's discuss...), and images.  At first, this could be done only for text files and it could get added to the prompt internally when we send, along with whatever you write.  To make this work, I think we should add a Plus Sign icon or Paper Clip icon beside/near the Microphone icon.  Drag-and-Drop is nice, but it's not easily discoverable.  This could get done in a follow-up.
3. The files in https://github.com/tarasglek/chatcraft.org/tree/main/src/components/Message need to get updated to be able to show an image.  When I write a message with an image, I'd expect the image to be present in the chat history display.
4. We need to figure out how to deal with images in the database.  We could add another table to store `Blob`s (see https://dexie.org/docs/API-Reference#storing-binary-data) for the uploaded files and reference those in the chat somehow.  When the chat is sent, we can serialize each entry from the files table by reading the blob and parsing it appropriately (e.g., text can be added as-is, a PDF would get parsed into text first, an image would get base64 encoded, etc).
5. For logged in ChatCraft users, it probably makes sense to upload images to CloudFlare and store in R2, then pass a URL vs. a base64 encoded message (see https://platform.openai.com/docs/guides/vision).  That's an optimization that can happen later.  

I tried using your current PR to review the work that @DukeManh is doing in #278:

Screenshot 2023-11-10 at 7 19 16 AM

The first issue I ran into is that the `width: 100px`, `height: 100px` makes it impossible to see what I'm talking about.  I understand why you did it, but we should find a better way to constrain the image, but also show it:

Screenshot 2023-11-10 at 7 19 07 AM

Maybe we need a way to show the while image in smaller form (maintain aspect ratio), or maybe we show an icon and you can click to see it (I don't love this, but maybe on mobile)?

Next problem: every time I send the prompt, I get back a similar length response (usually 10-15 words) and then it's cut off.  The text varies, but it's always the first bit only:

Screenshot 2023-11-10 at 7 21 14 AM

This is probably due to issues on the OpenAI side, but it's buggy enough that it makes the experience not very usable.

Reading the code, I think that `ChatBase.tsx` should probably get updated to have a more complex type for `onPrompt`, since we now need to be able to send more than just a `string`.  We should probably create a `Prompt` type that looks something like:

```ts
type Prompt = {
  prompt?: string;
  files?: File[];
};
```

Then we can delay processing the `File` objects until later, and outside the UI code, when we're going to send to the LLM.  It also makes it easier to store the `File`s as `Blob`s in the DB, or process them.  

Does this give you enough to keep iterating?

I'm really excited to see this happening.  Don't get overwhelmed by my requests above, much of it is me thinking out loud, and we can do this in a series of PRs.  Also, you can get other people to help you write pieces of it.

mingming-ma:
@humphd Very comprehensive feedback. Thanks a lot! I also noticed the length response cutoff bug. Let me try to solve this to make it works. Then I will solve the following in this PR:

- [ ] Update the model based on file type
- [ ] Attach file instead of drag file
- [ ] Better preview if attach file is image

I believe can be completed in the current PR and keep track of others in different issues about Message history shows image, DB and upload images for ChatCraft users.

tarasglek:
@mingming-ma I invited you to the repo, so you can benefit from CI/CD more.

mingming-ma:
@tarasglek Thank you so much!

humphd:
@mingming-ma you can push branches to this repo vs. your fork, and it will automatically deploy for people to test on CloudFlare.  If you want to close this and do it again as a branch here, that's fine.

mingming-ma:
@humphd That's great. I'll create an issue and create a branch to do this PR again

PR #284, opened 2023-11-10
Feature/system prompts table kosty

kosty:
Overall idea is to move starred system messages to a dedicated table, while keeping other "messages" as they are. Handling "starred" attribute seem difficult in this case. Likely it should become a dynamic check on specific wording being present in "starred" table.

### DIFF
--- a/src/components/Message/SystemMessage.tsx
+++ b/src/components/Message/SystemMessage.tsx
@@ -60,7 +60,7 @@ function SystemPromptVersionsMenu({
       ChatCraftStarredSystemPrompt.exists(promptMessage.text).catch((err) => {
         console.warn("Unable to query 'starred' table for PK", err);
         error({
-          title: "Error while checking for presense of Starred System Prompt in db",
+          title: "Error while checking for presence of starred System Prompt in db",
           message: err.message,
         });
         return false;
--- a/src/components/Message/SystemMessage.tsx
+++ b/src/components/Message/SystemMessage.tsx
@@ -20,7 +20,7 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
 import { createSystemPromptSummary, defaultSystemPrompt } from "../../lib/system-prompt";
 import db from "../../lib/db";
 import { ChatCraftSystemMessage } from "../../lib/ChatCraftMessage";
-import { ChatCraftStarredText } from "../../lib/ChatCraftStarredText";
+import { ChatCraftStarredSystemPrompt } from "../../lib/ChatCraftStarredSystemPrompt";
 import { useAlert } from "../../hooks/use-alert";
 
 function SystemPromptVersionsMenu({
@@ -55,17 +55,17 @@ function SystemPromptVersionsMenu({
     []
   );
 
-  const isStarredMessage = useLiveQuery<boolean>(async () => {
-    return ChatCraftStarredText.check(promptMessage.text);
+  const isStarredSystemPrompt = useLiveQuery<boolean>(async () => {
+    return ChatCraftStarredSystemPrompt.exists(promptMessage.text);
   }, [promptMessage]);
 
   const handleStarredChanged = () => {
-    const starredText = new ChatCraftStarredText({ text: promptMessage.text });
-    if (!isStarredMessage) {
+    const starredText = new ChatCraftStarredSystemPrompt({ text: promptMessage.text });
+    if (!isStarredSystemPrompt) {
       starredText.save().catch((err) => {
         console.warn("Unable to update system prompt", err);
         error({
-          title: `Error changing starred for System Prompt`,
+          title: `Error while saving Starred System Prompt to db`,
           message: err.message,
         });
       });
@@ -73,14 +73,14 @@ function SystemPromptVersionsMenu({
       starredText.remove().catch((err) => {
         console.warn("Unable to update system prompt", err);
         error({
-          title: `Error changing starred for System Prompt`,
+          title: `Error while removing Starred System Prompt from db`,
           message: err.message,
         });
       });
     }
   };
 
-  const title = isStarredMessage
+  const title = isStarredSystemPrompt
     ? "Unstar System Prompt to forget it"
     : "Star System Prompt to save for future use";
 
@@ -90,7 +90,7 @@ function SystemPromptVersionsMenu({
         size="sm"
         aria-label={title}
         title={title}
-        icon={isStarredMessage ? <TbStarFilled /> : <TbStar />}
+        icon={isStarredSystemPrompt ? <TbStarFilled /> : <TbStar />}
         variant="ghost"
         onClick={handleStarredChanged}
       />
--- a/src/components/Message/SystemMessage.tsx
+++ b/src/components/Message/SystemMessage.tsx
@@ -46,34 +46,43 @@ function SystemPromptVersionsMenu({
           console.error("Failed to query the starred table:", error);
         });
       if (!records) {
-        return Promise.resolve([defaultSystemPrompt()]);
+        return [defaultSystemPrompt()];
       }
 
-      return Promise.resolve([defaultSystemPrompt(), ...records]);
+      return [defaultSystemPrompt(), ...records];
     },
     [],
     []
   );
 
-  const isStarredSystemPrompt = useLiveQuery<boolean>(async () => {
-    return ChatCraftStarredSystemPrompt.exists(promptMessage.text);
-  }, [promptMessage]);
+  const isStarredSystemPrompt = useLiveQuery<boolean>(
+    () =>
+      ChatCraftStarredSystemPrompt.exists(promptMessage.text).catch((err) => {
+        console.warn("Unable to query 'starred' table for PK", err);
+        error({
+          title: "Error while checking for presense of Starred System Prompt in db",
+          message: err.message,
+        });
+        return false;
+      }),
+    [promptMessage]
+  );
 
   const handleStarredChanged = () => {
     const starredText = new ChatCraftStarredSystemPrompt({ text: promptMessage.text });
     if (!isStarredSystemPrompt) {
       starredText.save().catch((err) => {
-        console.warn("Unable to update system prompt", err);
+        console.warn("Unable to save text to 'starred' table", err);
         error({
-          title: `Error while saving Starred System Prompt to db`,
+          title: "Error while saving Starred System Prompt to db",
           message: err.message,
         });
       });
     } else {
       starredText.remove().catch((err) => {
-        console.warn("Unable to update system prompt", err);
+        console.warn("Unable to remove text from 'starred' table", err);
         error({
-          title: `Error while removing Starred System Prompt from db`,
+          title: "Error while removing Starred System Prompt from db",
           message: err.message,
         });
       });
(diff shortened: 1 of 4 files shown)


src/lib/ChatCraftStarredText.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  async save() {
+    const promptText = this.text;
+    return db.starred
+      .where("text")
+      .equalsIgnoreCase(promptText)
+      .count()
+      .then((count) => {
+        if (count == 0) {
```
humphd, line 34: This makes me think that we should use the `text` itself as the primary key and drop the `id`.  We don't want duplicates, so let's simply remove the possibility of it happening.

src/lib/db.ts
```diff
@@ -107,11 +114,34 @@ class ChatCraftDatabase extends Dexie {
...
+          .toArray()
+          .then((entries) => {
+            // Prepare the entries for the systemPrompts table by removing the starred attribute
+            const promptsEntries = entries
+              .filter((entry) => entry.starred === true)
+              .map(({ id, date, text }) => {
+                return { id: id, date: date, text: text };
+              });
+            // Add the entries to the systemPrompts table
+            return tx.table("starred").bulkAdd(promptsEntries, { allKeys: true });
```
humphd, line 136: We need to also `.modify()` the `message` records to remove the `starred` property once it's been copied over.

src/lib/ChatCraftStarredText.ts
```diff
@@ -0,0 +1,97 @@
+import { nanoid } from "nanoid";
+import db, { type ChatCraftStarredTextTable } from "./db";
+
+export type SerializedChatCraftStarredText = {
+  id: string;
+  date: string;
+  text: string;
+};
+
+export class ChatCraftStarredText {
```
humphd, line 10: Let's rename this to `ChatCraftSavedSystemPrompt` or `ChatCraftStarredSystemPrompt` to be more clear what it is.

src/lib/ChatCraftStarredText.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  toDB(): ChatCraftStarredTextTable {
+    return {
+      id: this.id,
+      date: this.date,
+      text: this.text,
+    };
+  }
+
+  // static fromDB(message: ChatCraftStarredTextTable) {
```
humphd, line 90: What's happening with these commented out functions?

src/components/Message/SystemMessage.tsx
```diff
@@ -20,53 +20,79 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
...
-
-      return Promise.resolve([defaultSystemPrompt(), ...uniq]);
+      return Promise.resolve([defaultSystemPrompt(), ...records]);
     },
     [],
     []
   );
 
-  const title = isStarred ? "Unstar System Prompt" : "Star System Prompt";
+  const isStarredMessage = useLiveQuery<boolean>(async () => {
```
humphd, line 58: `isStarredSystemPrompt`

src/lib/ChatCraftStarredText.ts
```diff
@@ -0,0 +1,97 @@
...
+        entries.forEach((entry) => {
+          return db.starred.delete(entry.id);
+        });
+      })
+      .catch((error) => {
+        console.error("Failed to query the systemPrompts table:", error);
+      });
+  }
+
+  static async check(text: string): Promise<boolean> {
```
humphd, line 61: Let's name this `exists`

src/lib/ChatCraftStarredText.ts
```diff
@@ -0,0 +1,97 @@
...
+    return db.starred
+      .where("text")
+      .equalsIgnoreCase(text)
+      .count()
+      .then((count) => {
+        if (count > 0) {
+          return Promise.resolve(true);
```
humphd, line 68: You can avoid the manual `Promise`s here.  Because you're already returning a `Promise` via the `then()` promise chain, you can simply return your final result, and the containing `Promise` will resolve it.

humphd (changes requested):
This is looking really good @kosty, great work.  I've added some comments of things to fix.  I'll test it locally on some live migrations after you do that.

Appreciate you jumping into this, it's been bugging us for a while.

src/components/Message/SystemMessage.tsx
```diff
@@ -20,53 +20,79 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
...
+    return ChatCraftStarredText.check(promptMessage.text);
+  }, [promptMessage]);
+
+  const handleStarredChanged = () => {
+    const starredText = new ChatCraftStarredText({ text: promptMessage.text });
+    if (!isStarredMessage) {
+      starredText.save().catch((err) => {
+        console.warn("Unable to update system prompt", err);
+        error({
+          title: `Error changing starred for System Prompt`,
```
tarasglek, line 68: grammar

tarasglek: Refusing to change starred prompt?

src/components/Message/SystemMessage.tsx
```diff
@@ -20,53 +20,79 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
...
+        .then((entries) => {
+          // Map the array of entries to get an array of 'text' attributes
+          return entries.map((entry) => entry.text);
+        })
+        .catch((error) => {
+          console.error("Failed to query the starred table:", error);
+        });
       if (!records) {
-        return [];
+        return Promise.resolve([defaultSystemPrompt()]);
```
humphd, line 49: Because you're in an `async` function, all `return` and `throw` statements are automatically wrapped in a `Promise`, you can remove `Promise.resolve()` here and in line 52.

src/components/Message/SystemMessage.tsx
```diff
@@ -20,53 +20,79 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
...
-
-      return Promise.resolve([defaultSystemPrompt(), ...uniq]);
+      return Promise.resolve([defaultSystemPrompt(), ...records]);
     },
     [],
     []
   );
 
-  const title = isStarred ? "Unstar System Prompt" : "Star System Prompt";
+  const isStarredSystemPrompt = useLiveQuery<boolean>(async () => {
```
humphd, line 58: nit: in an arrow function that only returns a value, drop the `async`, `{...}` body, and `return` statement:

```ts
async () => { return foo() }

// becomes


() => foo()
```

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,78 @@
...
+  text: string;
+};
+
+export class ChatCraftStarredSystemPrompt {
+  id: string;
+  date: Date;
+  text: string;
+
+  constructor({ id, date, text }: { id?: string; date?: Date; text: string }) {
+    this.id = id ?? nanoid();
```
humphd, line 16: Remove `id`, use `text` as pk.

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,78 @@
...
+        entries.forEach((entry) => {
+          return db.starred.delete(entry.id);
+        });
+      })
+      .catch((error) => {
+        console.error("Failed to query the systemPrompts table:", error);
+      });
+  }
+
+  static async exists(text: string): Promise<boolean> {
```
humphd, line 61: Drop `async` on this, you are already returning the `Promise` implicitly.

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+          console.warn(`${promptText} was already found in DB. Ignoring`);
+        }
+      })
+      .catch((error) => {
+        console.error("Failed to query the systemPrompts table:", error);
+      });
+  }
+
+  async remove() {
```
humphd, line 45: Drop `async` on this, you are already returning the `Promise` implicitly.

humphd (changes requested):
Looks like I reviewed again before you had a chance to fix everything, sorry about that.  I've added more comments.

This is looking better and better.

tarasglek:
I am not a fan of this UX. At the moment prompts magically disappear when you switch away from them if you don't star them.

I would prefer to have prompts persist by default + delete button to get rid of em explicitly

kosty:
I like what @tarasglek  is saying as it has fewer UI gotchas. Currently one has to somehow figure out what "star" means. Either by hovering or just by trial/error. 

One trouble though is that there is difference between saving a message (and it affects any/all messages) and how starred prompts are handled. To get to what @taras is saying we could move all the storage logic to `ChatCraftSystemMessage` yet one would have to come up with eloquent way to present those additional UI butons for "delete". Namely, distinguis between "delete message" and "delete starred system prompt" (aka remove star).

kosty:
@humphd @tarasglek  added fixes addressing David's comments yet lets keep talking about UI. Would be great if we could navigate to a place that makes ChatCraft and exciting experience

One observation I have is that first message in chat seems to have several different jobs to do
- manage starred prompts
- provide context for rest of the chat

Maybe we could keep star as a way to add texts to stored system prompts and have a dedicated view for management of those, probably somewere in settings, along with export / import.

This would play nicely into next feature - exporting/importing starred text to/from current user's github repo. WDYT?

kosty:
@tarasglek another naïve idea - what if we just change star UI element to a button that say "Store this promt" / "Forget this pompt" ?

src/components/Message/SystemMessage.tsx
```diff
@@ -20,53 +20,88 @@ import MessageBase, { type MessageBaseProps } from "./MessageBase";
...
     []
   );
 
-  const title = isStarred ? "Unstar System Prompt" : "Star System Prompt";
+  const isStarredSystemPrompt = useLiveQuery<boolean>(
+    () =>
+      ChatCraftStarredSystemPrompt.exists(promptMessage.text).catch((err) => {
+        console.warn("Unable to query 'starred' table for PK", err);
+        error({
+          title: "Error while checking for presense of Starred System Prompt in db",
```
humphd, line 63:
```suggestion
          title: "Error while checking for presence of starred System Prompt in db",
```

humphd (changes requested):

tarasglek:
> @tarasglek another naïve idea - what if we just change star UI element to a button that say "Store this promt" / "Forget this pompt" ?

I would prefer to not lose data by default. Just rank prompts by usage if we get too many. It's a good problem to have.

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  clone() {
+    return new ChatCraftStarredSystemPrompt({
+      text: this.text,
+    });
+  }
+
+  async save() {
+    return db.starred.get(this.text).then((starred) => {
```
humphd, line 24: Question: if the text is our PK, why not just write it again vs. bothering with a `get`?  No matter what, we want to end up with the data in the db, so either write or overwrite.

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,46 @@
...
+      return this.text;
+    });
+  }
+
+  remove() {
+    return db.starred.delete(this.text);
+  }
+
+  static exists(text: string): Promise<boolean> {
+    return db.starred.get(text).then((entry) => entry !== undefined);
```
humphd, line 37:
```suggestion
    return db.starred.get(text).then(Boolean);
```

src/lib/db.ts
humphd: This is still undone, an you add it?

Also, we should be able to drop `starred?: boolean;` from `ChatCraftMessageTable`, etc.

humphd (changes requested):

PR #286, opened 2023-11-10
Allow using images as input with gpt-4 vision

### LINKED ISSUE #285: Add - Allow using images as input with gpt-4 vision
This issue is used for tracking the branch which continue the develop from PR https://github.com/tarasglek/chatcraft.org/pull/281

## Feedback
https://github.com/tarasglek/chatcraft.org/pull/281#issuecomment-1805657667

### response cutoff bug (still needs to set max_tokens when updated to `openai@4.19.0`)
image

### <s>type mismatch</s> (Fixed by using `ChatCompletionContentPart` of `openai@4.19.0`)
<details>
Before:
New payload format not meet the interface type, trying to fix
image

After:
image
</details>

mingming-ma:
Fix #285 

[Result](https://issue-285.console-overthinker-dev.pages.dev/)

## Tasks
<details>
-  Fix response cutoff bug
    - [x] found a workaround by setting `max_tokens`. Tested on latest version openai@v4.20.0, without max_tokens response still cutoff
- [x] Update the model based on file type
-  Attach file method
    - [x] Clip button
    - [x] Directly Paste   
- [x] Better preview if attach file is image
    - [x] full width of small version
    - [x] Click on the small image to show a larger version with a "close" button in the middle of the window.
- [x] Keep images in chat history
- [x] Images in messages should use width 100%, and be clickable to see full version.
- [x] Support adding more than 1 image at a time in the file picker
- [x] Add an index number to the top left corner of each image
- [x] Re-focus after closing the full image
- [x] Fix no text no respond [bug](https://github.com/tarasglek/chatcraft.org/pull/286#pullrequestreview-1844494192)
- [x] Fix number button shape: oval -> round
- [x] Fix number/cross button margin too close
- [x] Fix cursor ->  `cursor: pointer`
</details>

## Issues

- Unresolved task: Resize the image in the browser within the max size of the image that OpenAI will process
    - Found  `Uncaught ReferenceError: process is not defined` when trying to do `sharp` convert/resize images, guessing it needs server side running so cause this error.
    -> #395 
    
- cutoff bug workaround [reminder](https://github.com/tarasglek/chatcraft.org/pull/286#discussion_r1483799736)

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -200,7 +200,7 @@ function ChatBase({ chat }: ChatBaseProps) {
           // Add this prompt message to the chat
           promptMessage = new ChatCraftHumanMessage({ text: prompt, imageUrls, user });
           await chat.addMessage(promptMessage);
-        } else if (imageUrls && imageUrls.length) {
+        } else if (imageUrls?.length) {
           // Add only image to the chat
           promptMessage = new ChatCraftHumanMessage({ text: "", imageUrls, user });
           await chat.addMessage(promptMessage);
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -146,7 +146,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string, image?: string[]) => {
+    async (prompt?: string, imageUrls?: string[]) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
@@ -198,11 +198,11 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, imageUrls, user });
           await chat.addMessage(promptMessage);
-        } else if (image && image.length) {
+        } else if (imageUrls && imageUrls.length) {
           // Add only image to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: "", image, user });
+          promptMessage = new ChatCraftHumanMessage({ text: "", imageUrls, user });
           await chat.addMessage(promptMessage);
         } else {
           // If there isn't any prompt text, see if the final message in the chat was a human
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -200,7 +200,7 @@ function ChatBase({ chat }: ChatBaseProps) {
           // Add this prompt message to the chat
           promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
           await chat.addMessage(promptMessage);
-        } else if (image) {
+        } else if (image && image.length) {
           // Add only image to the chat
           promptMessage = new ChatCraftHumanMessage({ text: "", image, user });
           await chat.addMessage(promptMessage);
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -146,7 +146,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string) => {
+    async (prompt?: string, image_url?: string[]) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
@@ -198,7 +198,7 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, image_url, user });
           await chat.addMessage(promptMessage);
         } else {
           // If there isn't any prompt text, see if the final message in the chat was a human
@@ -239,6 +239,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
         // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
         const messages = chat.messages({ includeAppMessages: false });
+        console.log(messages);
         const response = await callChatApi(messages, {
           functions,
           functionToCall,
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -200,6 +200,10 @@ function ChatBase({ chat }: ChatBaseProps) {
           // Add this prompt message to the chat
           promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
           await chat.addMessage(promptMessage);
+        } else if (image) {
+          // Add only image to the chat
+          promptMessage = new ChatCraftHumanMessage({ text: "", image, user });
+          await chat.addMessage(promptMessage);
         } else {
           // If there isn't any prompt text, see if the final message in the chat was a human
           // message or a function response. If it was either, we'll allow sending that through
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -146,7 +146,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string, image_url?: string[]) => {
+    async (prompt?: string, image?: string[]) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
@@ -198,7 +198,7 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, image_url, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
           await chat.addMessage(promptMessage);
         } else {
           // If there isn't any prompt text, see if the final message in the chat was a human
@@ -239,7 +239,6 @@ function ChatBase({ chat }: ChatBaseProps) {
 
         // NOTE: we strip out the ChatCraft App messages before sending to OpenAI.
         const messages = chat.messages({ includeAppMessages: false });
-        console.log(messages);
         const response = await callChatApi(messages, {
           functions,
           functionToCall,
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -6,7 +6,6 @@ import {
   ModalCloseButton,
   ModalBody,
   Image,
-  Box,
   Flex,
 } from "@chakra-ui/react";
 
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -6,6 +6,8 @@ import {
   ModalCloseButton,
   ModalBody,
   Image,
+  Box,
+  Flex,
 } from "@chakra-ui/react";
 
 interface ImageModalProps {
@@ -17,17 +19,20 @@ interface ImageModalProps {
 const ImageModal: React.FC<ImageModalProps> = ({ isOpen, onClose, imageSrc }) => (
   <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
     <ModalOverlay />
-    <ModalContent maxW="90vw" maxH="90vh">
+    <ModalContent maxW="90vw" height="90vh">
       <ModalCloseButton />
       <ModalBody>
-        <Image
-          src={imageSrc}
-          alt="Selected Image"
-          m="auto"
-          width="100%"
-          maxH="85vh"
-          objectFit="contain"
-        />
+        <Flex height={"100%"} justifyContent={"center"} alignItems={"center"}>
+          <Image
+            maxWidth="100%"
+            maxHeight="70vh"
+            overflow={"auto"}
+            src={imageSrc}
+            alt="Selected Image"
+            m="auto"
+            objectFit="contain"
+          />
+        </Flex>
       </ModalBody>
     </ModalContent>
   </Modal>
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -15,12 +15,19 @@ interface ImageModalProps {
 }
 
 const ImageModal: React.FC<ImageModalProps> = ({ isOpen, onClose, imageSrc }) => (
-  <Modal isOpen={isOpen} onClose={onClose} size="xl" isCentered>
+  <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
     <ModalOverlay />
-    <ModalContent>
+    <ModalContent maxW="90vw" maxH="90vh">
       <ModalCloseButton />
       <ModalBody>
-        <Image src={imageSrc} alt="Selected Image" maxW="100%" maxH="100vh" m="auto" />
+        <Image
+          src={imageSrc}
+          alt="Selected Image"
+          m="auto"
+          width="100%"
+          maxH="85vh"
+          objectFit="contain"
+        />
       </ModalBody>
     </ModalContent>
   </Modal>
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import {
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalCloseButton,
+  ModalBody,
+  Image,
+} from "@chakra-ui/react";
+
+interface ImageModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+  imageSrc: string;
+}
+
+const ImageModal: React.FC<ImageModalProps> = ({ isOpen, onClose, imageSrc }) => (
+  <Modal isOpen={isOpen} onClose={onClose} size="xl" isCentered>
+    <ModalOverlay />
+    <ModalContent>
+      <ModalCloseButton />
+      <ModalBody>
+        <Image src={imageSrc} alt="Selected Image" maxW="100%" maxH="100vh" m="auto" />
+      </ModalBody>
+    </ModalContent>
+  </Modal>
+);
+
+export default ImageModal;
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -18,7 +18,7 @@ interface ImageModalProps {
 const ImageModal: React.FC<ImageModalProps> = ({ isOpen, onClose, imageSrc }) => (
   <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
     <ModalOverlay />
-    <ModalContent maxW="90vw" height="90vh">
+    <ModalContent maxW="90vw" maxHeight="90vh">
       <ModalCloseButton />
       <ModalBody>
         <Flex height={"100%"} justifyContent={"center"} alignItems={"center"}>
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -54,6 +54,7 @@ import { ChatCraftModel } from "../../lib/ChatCraftModel";
 import { useModels } from "../../hooks/use-models";
 import { useSettings } from "../../hooks/use-settings";
 import { useAlert } from "../../hooks/use-alert";
+import ImageModal from "../ImageModal";
 
 // Styles for the message text are defined in CSS vs. Chakra-UI
 import "./Message.css";
@@ -115,6 +116,9 @@ function MessageBase({
   const messageForm = useRef<HTMLFormElement>(null);
   const messageContent = useRef<HTMLDivElement>(null);
   const meta = useMemo(getMetaKey, []);
+  const [imageModalOpen, setImageModalOpen] = useState<boolean>(false);
+  const [selectedImage, setSelectedImage] = useState<string>("");
+
   const { isOpen, onToggle: originalOnToggle } = useDisclosure();
   const isLongMessage = text.length > 5000;
   const displaySummaryText = !isOpen && (summaryText || isLongMessage);
@@ -269,6 +273,12 @@ function MessageBase({
     },
     [message, onResubmitClick, chatId, error, onEditingChange]
   );
+
+  const openModalWithImage = (imageSrc: string) => {
+    setSelectedImage(imageSrc);
+    setImageModalOpen(true);
+  };
+  const closeModal = () => setImageModalOpen(false);
 
   return (
     <Box
@@ -487,7 +497,8 @@ function MessageBase({
                       key={`${id}-${index}`}
                       src={image}
                       alt={`Images# ${index}`}
-                      style={{ width: "100%" }}
+                      width="100%"
+                      onClick={() => openModalWithImage(image)}
                     />
                   ))}
                   <Markdown
@@ -506,6 +517,7 @@ function MessageBase({
                 </Box>
               )}
             </Box>
+            <ImageModal isOpen={imageModalOpen} onClose={closeModal} imageSrc={selectedImage} />
           </Flex>
         </CardBody>
         {footer && <CardFooter py={2}>{footer}</CardFooter>}
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -54,7 +54,6 @@ import { ChatCraftModel } from "../../lib/ChatCraftModel";
 import { useModels } from "../../hooks/use-models";
 import { useSettings } from "../../hooks/use-settings";
 import { useAlert } from "../../hooks/use-alert";
-import ImageModal from "../ImageModal";
 
 // Styles for the message text are defined in CSS vs. Chakra-UI
 import "./Message.css";
@@ -105,7 +104,7 @@ function MessageBase({
   disableFork,
   disableEdit,
 }: MessageBaseProps) {
-  const { id, date, text, image } = message;
+  const { id, date, text } = message;
   const { models } = useModels();
   const { onCopy } = useClipboard(text);
   const { info, error } = useAlert();
@@ -116,9 +115,6 @@ function MessageBase({
   const messageForm = useRef<HTMLFormElement>(null);
   const messageContent = useRef<HTMLDivElement>(null);
   const meta = useMemo(getMetaKey, []);
-  const [imageModalOpen, setImageModalOpen] = useState<boolean>(false);
-  const [selectedImage, setSelectedImage] = useState<string>("");
-
   const { isOpen, onToggle: originalOnToggle } = useDisclosure();
   const isLongMessage = text.length > 5000;
   const displaySummaryText = !isOpen && (summaryText || isLongMessage);
@@ -273,12 +269,6 @@ function MessageBase({
     },
     [message, onResubmitClick, chatId, error, onEditingChange]
   );
-
-  const openModalWithImage = (imageSrc: string) => {
-    setSelectedImage(imageSrc);
-    setImageModalOpen(true);
-  };
-  const closeModal = () => setImageModalOpen(false);
 
   return (
     <Box
@@ -492,15 +482,6 @@ function MessageBase({
                   // Add a single pixel of offset for rendering to canvas (offset handled above with m=-1)
                   p={1}
                 >
-                  {image.map((image, index) => (
-                    <Image
-                      key={`${id}-${index}`}
-                      src={image}
-                      alt={`Images# ${index}`}
-                      width="100%"
-                      onClick={() => openModalWithImage(image)}
-                    />
-                  ))}
                   <Markdown
                     previewCode={!hidePreviews && !displaySummaryText}
                     isLoading={isLoading}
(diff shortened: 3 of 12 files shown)


mingming-ma:
## Response cutoff bug investigate
### Results from python scripts for local images vs. chatcraft


image


image

humphd:
This might be worth filing upstream with the https://github.com/openai/openai-node repo itself, since it doesn't seem to be an OpenAI issue.

mingming-ma:
> This might be worth filing upstream with the https://github.com/openai/openai-node repo itself, since it doesn't seem to be an OpenAI issue.

Yeah, I think the `openai-node` will be updating in the following days, and found another thing, after setting the `max_tokens` in `chatCompletionParams`, seems no cutoff. 

```tsx
    max_tokens: 4096,
```

Result:
image

I found someone said `max_tokens` default to max value of current model, and by looking into `openai-node` code: `ChatCompletionCreateParamsBase` currently only write
```tsx
  model:
    | (string & {})
    | 'gpt-4'
    | 'gpt-4-0314'
    | 'gpt-4-0613'
    | 'gpt-4-32k'
    | 'gpt-4-32k-0314'
    | 'gpt-4-32k-0613'
    | 'gpt-3.5-turbo'
    | 'gpt-3.5-turbo-16k'
    | 'gpt-3.5-turbo-0301'
    | 'gpt-3.5-turbo-0613'
    | 'gpt-3.5-turbo-16k-0613';
```

maybe these given models impact the default max_tokens of `gpt4-vision-preview`, I think I can leave this for now to develop the other tasks.

humphd:
@mingming-ma this is coming along really well, I'm impressed.  A few things I notice testing this today:

Screenshot 2023-11-23 at 8 48 18 AM

1. I wonder if the close X circle could go over top of the image (maybe top-right corner?) instead of below it. Could happen in follow-up/someone else could do.
2. I want to be able to click on the small version of the image in the prompt form and have it go big so I can see it better (like a light-box that shows in the centre of the screen and you can close?).  Could happen in follow-up/someone else could do.
3. I love how the model changes to when I attach an image, that's smart
4. When I send the image + prompt, I lose the image:

Screenshot 2023-11-23 at 8 51 13 AM

We should probably alter the message component so that it can show the image that was attached.

Keep going, you're doing great work here!

mingming-ma:
@humphd Thanks a lot for the feedback! I just tested store images in db, now we can keep images after submit. Note that since I changed the db schema, better try this in the Incognito Window so that not impact existing IndexedDB. I also tested in my normal Window and seems working as before.

humphd:
@mingming-ma do you want to fix the conflicts you have in this branch?

mingming-ma:
@humphd Sure, seems no conflicts on my side for now, let me know I'll fix them.

humphd:
For some reason, this is showing:

> This branch cannot be rebased due to conflicts

At any rate, some more feedback.

First, I **love** this.  I showed it to some other people and they were blown away too.  Some comments they made:

1. Should be able to paste an image and have it get attached to the current prompt (can happen in follow-up)
2. Images in messages should use width 100%, and be clickable to see full version.

Another optimization we should make: what is the max size of the image that OpenAI will process?  We should resize the image in the browser so we don't waste bandwidth when sending. Can happen in follow-up.

mingming-ma:
@humphd Oh, It must be that I merged the main branch midway. Thanks a lot for the feed back! I was thinking about implementing the paste function yesterday, and we had the same thought! I'll update the width and check image size for the optimization.

mingming-ma:
@humphd I've done a rebase. Are there still any conflicts?

humphd:
No, looks good.  I want to do a pass over the code too, I haven't read it yet.  I'll try to do that this week.

mingming-ma:
That's great! This PR involves many components, so no rush on the review. Honestly, the code implementation is still quite rough. Can’t wait to see your thoughts on it!

mingming-ma:
Nice catch 👍  just fixed that

src/components/Message/MessageBase.tsx
```diff
@@ -381,9 +381,19 @@ function MessageBase({
                   </VStack>
                 </form>
               ) : (
-                <Markdown previewCode={!hidePreviews} isLoading={isLoading} onPrompt={onPrompt}>
-                  {summaryText || text}
-                </Markdown>
+                <>
+                  {image.map((image, index) => (
+                    <img
```
humphd, line 386: Do you want to use https://chakra-ui.com/docs/components/image/usage here, and have it fill the width better?

src/components/PromptForm/ClipIcon.tsx
```diff
@@ -0,0 +1,52 @@
+import { useState, useRef } from "react";
```
humphd, line 1: Can we rename `Clip` everywhere, including the filename, to be `Attach` or something that better describes what this is for?

src/components/PromptForm/ClipIcon.tsx
```diff
@@ -0,0 +1,52 @@
...
+import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
+
+type ClipIconProps = {
+  isDisabled: boolean;
+  onFileSelected: (base64: string) => void;
+};
+
+export default function ClipIcon({ isDisabled = false, onFileSelected }: ClipIconProps) {
+  const isMobile = useMobileBreakpoint();
+  const [colorScheme /*, setColorScheme */] = useState<"blue" | "red">("blue");
```
humphd, line 14: if you don't need set, please remove

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -105,12 +135,39 @@ function DesktopPromptForm({
     };
   }, [isRecording, recordingSeconds]);
 
+  // Update model to the supported model when inputImages is not empty
+  useEffect(() => {
+    if (inputImages && inputImages.length > 0) {
+      const modelSupportsImages = models.find((model) => model.supportsImages);
```
humphd, line 141: Move this logic to the Model class, similar to other checks we do for OpenAI specific things (e.g., function calling).

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -105,12 +135,39 @@ function DesktopPromptForm({
...
+      }
+    }
+    // Assuming models and setSettings are stable and don't need to be in the dependencies array
+    // eslint-disable-next-line
+  }, [inputImages]);
+
+  // Attach paste event listener to the textarea
+  useEffect(() => {
+    const textAreaElement = inputPromptRef.current;
+    if (textAreaElement) {
```
humphd, line 153: What if I paste an image?

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -163,9 +220,55 @@ function DesktopPromptForm({
...
+  };
+
+  const handlePasteImage = (e: ClipboardEvent) => {
+    const clipboardData = e.clipboardData;
+    const items = Array.from(clipboardData?.items || []);
+    const imageFiles = items
+      .map((item) => item.getAsFile())
+      .filter((file): file is File => file != null && file.type.startsWith("image/"));
+    if (imageFiles.length > 0) {
+      Promise.all(imageFiles.map((file) => getBase64FromFile(file))).then((base64Strings) => {
```
humphd, line 268: Add a `.catch()` here so we at least show an error popup, like in other parts of the code that use `alerts`.

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -174,45 +277,89 @@ function DesktopPromptForm({
...
+                bg="white"
+                _dark={{ bg: "gray.700" }}
+                onDragOver={(e) => e.preventDefault()}
+                onDrop={handleDropImage}
+              >
+                <Flex w="100%" h="100%" direction="column">
+                  <Flex flexWrap="wrap">
+                    {inputImages.map((image, index) => (
+                      <Box key={index} position="relative" height="100px" m={2}>
+                        <Image
```
humphd, line 291: Exactly, do it like this above.

src/lib/ChatCraftMessage.ts
```diff
@@ -93,13 +98,25 @@ export class ChatCraftMessage {
     };
   }
 
-  toOpenAiMessage(): OpenAI.Chat.Completions.CreateChatCompletionRequestMessage {
+  toOpenAiMessage(): OpenAI.Chat.Completions.ChatCompletionMessageParam {
     const text = this.text;
+
+    const textAndImage: OpenAI.Chat.Completions.ChatCompletionContentPart[] = [];
```
humphd, line 104: This is oddly named, since it might not have an image in it.  I'd call it `content` or something more generic.

src/lib/ChatCraftMessage.ts
```diff
@@ -93,13 +98,25 @@ export class ChatCraftMessage {
...
+    textAndImage.push({ type: "text", text: this.text });
+    if (this.image && this.image.length > 0) {
```
humphd, line 106: Why would `this.image` ever be undefined here?

src/lib/ChatCraftModel.ts
```diff
@@ -52,6 +52,10 @@ export class ChatCraftModel {
     return this.name.startsWith("gpt-3.5-turbo") || this.name.startsWith("gpt-4");
   }
 
+  get supportsImages() {
```
humphd, line 55: Call this in your `useEffect` above.

src/lib/ai.ts
```diff
@@ -290,6 +290,14 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
           : "auto"
         : undefined,
+
+    /*
+     Got this value from error message of OpenAI's gpt-4-vision-preview max_tokens
+     by setting a very large number,
+     After setting this value, seems no cutoff
+     Tested on latest version openai@v4.20.0, without max_tokens response still cutoff
+     */
+    max_tokens: model.supportsImages ? 4096 : undefined,
```
humphd, line 300: This worries me.  Is this going to mean we can't go above 4096?  We should confirm this.

src/lib/db.ts
```diff
@@ -19,7 +19,8 @@ export type ChatCraftMessageTable = {
   user?: User;
   func?: FunctionCallParams | FunctionCallResult;
   text: string;
-  versions?: { id: string; date: Date; model: string; text: string }[];
+  image: string[];
+  versions?: { id: string; date: Date; model: string; text: string; image: string[] }[];
```
humphd, line 23: I think `image: string[]` has to be optional, since no existing data has it.  If you want to include it like you have here, you need to run a migration step to add `[]` to all old messages in the db.

humphd (changes requested):

mingming-ma:
@humphd Absolutely! I didn't have much free time as the end of the semester, but now that it's behind me, I'm committed to making the necessary changes. Thanks a lot for your feedback, and if you can give me another two to three days, I'll do my best to get it done.

src/components/Message/MessageBase.tsx
mingming-ma: Fixed at  [cb915e3](https://github.com/tarasglek/chatcraft.org/pull/286/commits/cb915e357fe1a4cc42b34b35b0907e4af2ecca37)

src/components/PromptForm/ClipIcon.tsx
mingming-ma: Fixed at [c65d529](https://github.com/tarasglek/chatcraft.org/pull/286/commits/c65d529c4ba77ff2fb2b3e688f5f236e8851abe6)

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: Yeah, currently supports both image and text copy paste

src/lib/ai.ts
mingming-ma: From this doc https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo, it says "Returns a maximum of 4,096 output tokens." not 100% percent sure for this setting, most likely the reason 
image

My personal test: 
image

src/lib/db.ts
mingming-ma: Fixed at [5d3143b](https://github.com/tarasglek/chatcraft.org/pull/286/commits/5d3143b4c8fd7fc5381569a1045ed807b3b6e2d1)

src/components/PromptForm/ClipIcon.tsx
mingming-ma: Fixed at [3dea32b](https://github.com/tarasglek/chatcraft.org/pull/286/commits/3dea32b7a73e5ad6afa13fc7cd390ad574c10fc9)

src/lib/ChatCraftMessage.ts
mingming-ma: Fixed at [21c96ef](https://github.com/tarasglek/chatcraft.org/pull/286/commits/21c96effb4828767df96ded959e1e872a0caea0c)

src/lib/ChatCraftMessage.ts
mingming-ma: Fixed at [fba7bec](https://github.com/tarasglek/chatcraft.org/pull/286/commits/fba7bec930c0a4222b2f4cca8287dbce32e6fc96)

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: Fixed at [589c95a](https://github.com/tarasglek/chatcraft.org/pull/286/commits/589c95ab3c03674aaf84ce0e009854e24c1f1e17)

mingming-ma:
@humphd I think I have fixed most except the 
> [Move this logic to the Model class, similar to other checks we do for OpenAI specific things (e.g., function calling).](https://github.com/tarasglek/chatcraft.org/pull/286#discussion_r1416244517 )
> [Call this in your useEffect above.](https://github.com/tarasglek/chatcraft.org/pull/286#discussion_r1416249489)

I tried to do factor but haven't found a good approach yet. Can you give me a few more hints, or should we consider addressing this in a future pull request?

humphd:
Two more requests:

1. Support adding more than 1 image at a time in the file picker.  You can add multiple images, but not at once.
2. When there are multiple images, put a number on each one, so it's possible to refer to them.  The LLM seems to understand the order of the images you send in the array, so as long as it matched that order, it might be OK:

Screenshot 2024-01-13 at 11 43 46 AM

In the example above, I have 2 screenshots, but no good way to describe them in my prompt.  If there was a "1" and a "2" over top-left corder, similar to your current "X"

rjwignar:
I've been playing around with this feature for a bit and it's really neat!
One thing I noticed is that when clicking a small image in the input area and then closing the larger version (whether by pressing "Esc" or clicking the Close Button), the input textarea is no longer focused.

Should the prompt be re-focused after closing the full image?

kliu57 (commented):
@mingming-ma
I've tested the following functionalities in my browser and they look good:

- attach multiple images with paperclip icon
- attach multiple images by pasting from clipboard
- remove images in list of attached images
- ask gpt-4-vision-preview to describe images
- ask gpt-4-vision-preview to translate text in image to text
- ask gpt-4-vision-preview to compare two images
- images attached are numbered (index number on top left) and can be removed
- images attached are clickable to see full version
- images stay in chat history
- images sent to chat are sized at width 100% and are clickable

Some questions I have:

- I understand the need to click on an attached image to get a popup with a bigger version of it, but what is the point of being able to click on the image already sent to gpt-4? All you would get is a smaller or equal size version of it.

question1

- When I have attached images but no message written, and click Ask gpt-4-vision-preview, gpt-4 does not answer and the images are wiped. this is the behaviour we want?

question2

mingming-ma:
> I've been playing around with this feature for a bit and it's really neat! One thing I noticed is that when clicking a small image in the input area and then closing the larger version (whether by pressing "Esc" or clicking the Close Button), the input textarea is no longer focused.
> 
> Should the prompt be re-focused after closing the full image?

@rjwignar Thanks for the suggestion. I remember I have did some re-focus in this PR, let me see if this can be added also

mingming-ma:
@kliu57 Thanks for the testing. 
For the question 1, I think the requirement is coming from some point when the image in history does not have full width of the page.
We need to discuss how to optimize this, by either shows smaller version rather than full width image in history or abandon this popup feature.

For the question 2. I think better send the image and get the response rather than wipe out. Actually I remember I have did this before, let me see why it's not working.

humphd:
Using this today, I noticed a few things:

First, I love it, and want to see us ship this ASAP!

Second, here are some reflections on the image overlays

image

The `1` should be round, not an oval.

Both the image number and close button need some margin at the top and left so they aren't so tight.

The close button is clickable, so the cursor pointer needs to change to reflect that.

mingming-ma:
@humphd Thanks for the feedback! Currently the small picture and close button both clickable, I think I used same `pointer` image for now. Maybe we can change to use`zoom-in` image for the small picture? Any suggestions about the [cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) ?

humphd:
yes, `cursor: pointer` when something is user-clickable.

mingming-ma:
@humphd like this? 
image

humphd:
Exactly.

mingming-ma:
@humphd 1.2 sounds good and I will fix the issues asap

kliu57 (commented):
**Some questions:**

- Q1) I noticed the "attach image" icon is present even when we did not select gpt-4-vision-preview. And when I attach an image we get automatically switched over to gpt-4-vision-preview. Is this the behaviour we want?

If this is the behaviour we want, there is another issue where if you manually change it back to gpt-4 and send the image, we will get an error:

```
Response Error
OpenAI API Returned Error: 400 Invalid content type. image_url is only supported by certain models.
```

**[UPDATE]** As discussed in meeting, the current behaviour is okay.

- Q2) I noticed you changed the image widths in the past messages to not width 100% anymore and when we click on it it shows the image sized at window width 100%. Is this what we want? If it is maybe you can update the PR description with the changed expected behaviour. **[UPDATE]** As discussed in meeting, we want the image to be either centered or 100% width, and when we click on the image, we want the image to go no bigger than its original size.

- Q3) I noticed a past comment by Dave saying we should "Support adding more than 1 image at a time in the file picker. You can add multiple images, but not at once." Right now we are allowing multiple images at once. Was this requirement changed? **[UPDATE]** As discussed in meeting, we DO want to allow attaching multiple images at once.

**Issues:**

- Issue 1) Right now the code in your PR allows users to send blank messages, whereas the code in production does not allow the user to send messages that are blank or only has spaces. This affects not only gpt-4-vision-preview but other models. **[UPDATE]** As discussed in meeting, this issue needs to be fixed. We do not want the user to be able to send blank messages (current behaviour in production)

Everything else looks good to me.

Rachit1313:
Follow Ups (for Release 1.2):
- Not being able to send empty messages.
- Don't include images in the body of requests to the model for which images are not supported
Subsequent PRs:
- Image width (either make it 100% or responsive or centered)
- Preview size of the image while clicking whether limit it to the size of original image

kliu57 (commented):
Tested the following vision models on OpenRouter:

- gpt-4-vision-preview (works)
- gemini-pro-vision-vision (works)
- nous-hermes-2-vision-7b (does not work, see ai response below)

```
{"detail":""}{"reason":"error"}
```

Just FYI, if a user wants to use `gemini-pro-vision-vision`, then they would have to upload their images (causing the model to auto-change to `gpt-4-vision-preview`, and then they have to manually change the model to `gemini-pro-vision-vision`. If they change the model to `gemini-pro-vision-vision` first and upload the image it won't work cause we get automatically switched to `gpt-4-vision-preview`

Might be more of an edge case though since I assume not many people would use it.

tarasglek:
We should definitely not magically switch away from a model that works.

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -84,6 +108,14 @@ function DesktopPromptForm({
...
     }
   }, [isLoading, inputPromptRef]);
+  // Also focus when the attached images changes
+  useEffect(() => {
+    inputPromptRef.current?.focus();
+  }, [inputImages, inputPromptRef]);
+  // Also focus when the closing the modal image
+  useEffect(() => {
+    inputPromptRef.current?.focus();
+  }, [imageModalOpen, inputPromptRef]);
```
humphd, line 118: Could these dependencies be added to the previous `useEffect`, since they all do the same thing?

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -163,9 +222,72 @@ function DesktopPromptForm({
...
   const handleTranscriptionAvailable = (transcription: string) => {
     // Use this transcript as our prompt
-    onSendClick(transcription);
+    onSendClick(transcription, inputImages);
     setIsRecording(false);
     setIsTranscribing(false);
+    setInputImages([]);
+  };
+
+  const getBase64FromFile = (file: File): Promise<string> => {
```
humphd, line 231: We should extract this out into a separate class, maybe a `ChatCraftAttachment` class or something.

Can happen in a follow-up.

cc @rjwignar

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -163,9 +222,72 @@ function DesktopPromptForm({
...
+  };
+
+  const handleClickImage = (image: string) => {
+    setSelectedImage(image);
+    setImageModalOpen(true);
+  };
+  const closeModal = () => setImageModalOpen(false);
+
+  const handlePasteImage = (e: ClipboardEvent) => {
+    const clipboardData = e.clipboardData;
```
humphd, line 273: Should we check the `DataTransfer.types` and only do this for images?

src/lib/db.ts
```diff
@@ -19,7 +19,8 @@ export type ChatCraftMessageTable = {
   user?: User;
   func?: FunctionCallParams | FunctionCallResult;
   text: string;
-  versions?: { id: string; date: Date; model: string; text: string }[];
+  image?: string[];
```
humphd, line 22: I think this should be made a `Blob` in a follow-up, and renamed `attachments` or `files` or something, to indicate it can be various things (cc @rjwignar)

For now, let's rename this to `imageUrls` so it's clear, and do that throughout this PR.

humphd (changes requested):
@mingming-ma I've gone through the code.  It's really good, you've done a great job managing a very complex change that affects nearly the entire source tree.  Well done.

I have some changes that I want made before we land this on Friday, and some for follow-ups.

To do the rename from `image` to `imageUrls` in lots of places, I suggest you use TypeScript/VSCode and [Rename symbol](https://code.visualstudio.com/docs/editor/refactoring#_rename-symbol).

src/components/PromptForm/Attach.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const handleFileChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const files = event.target.files;
+    if (files) {
+      for (let i = 0; i < files.length; i++) {
+        const reader = new FileReader();
+        reader.onload = (e) => {
+          onFileSelected(e.target?.result as string);
+        };
+        reader.readAsDataURL(files[i]);
```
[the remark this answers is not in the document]
humphd: Filed #426

src/components/PromptForm/DesktopPromptForm.tsx
humphd: See #426 for later.

src/Chat/ChatBase.tsx
```diff
@@ -188,7 +188,11 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, image, user });
+          await chat.addMessage(promptMessage);
+        } else if (image && image.length) {
+          // Add only image to the chat
```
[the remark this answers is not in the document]
mingming-ma: seems good, maybe just for fun 😄 
image

src/components/ImageModal.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import {
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalCloseButton,
+  ModalBody,
+  Image,
+  Box,
```
Amnish04, line 9:
```suggestion
```

Forgot to remove this

Amnish04, line 9:
```suggestion
```

Amnish04 (commented):
Oops, forgot to remove the unused import

src/components/ImageModal.tsx
mingming-ma: No worries. Thanks for the help.

humphd:
This needs a rebase for `MessageBase.tsx`.  @mingming-ma please ping us when this is ready for more reviews, let's prioritize getting this merged.

mingming-ma:
@humphd Got it! Rebased done, just `DataTransfer.types` issue left, will soon to find out.

humphd:
> @humphd Got it! Rebased done, just `DataTransfer.types` issue left, will soon to find out.

Excited!

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: Have tried that using `DataTransfer.types` and it gives me string `Files`, so we still need some filter on each file like the following
```
      .filter((file): file is File => file != null && file.type.startsWith("image/"));
```

src/Chat/ChatBase.tsx
```diff
@@ -198,7 +198,11 @@ function ChatBase({ chat }: ChatBaseProps) {
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
-          promptMessage = new ChatCraftHumanMessage({ text: prompt, user });
+          promptMessage = new ChatCraftHumanMessage({ text: prompt, imageUrls, user });
+          await chat.addMessage(promptMessage);
+        } else if (imageUrls && imageUrls.length) {
```
humphd, line 203: nit: `imageUrls?.length`

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -105,12 +133,39 @@ function DesktopPromptForm({
     };
   }, [isRecording, recordingSeconds]);
 
+  // Update model to the supported model when inputImages is not empty
+  useEffect(() => {
+    if (inputImageUrls && inputImageUrls.length > 0) {
```
humphd, line 138: `inputImageUrls?.length > 0`

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -105,12 +133,39 @@ function DesktopPromptForm({
...
+      const modelSupportsImages = models.find((model) => model.supportsImages);
+      if (modelSupportsImages) {
+        setSettings({ ...settings, model: modelSupportsImages });
+      }
+    }
+    // Assuming models and setSettings are stable and don't need to be in the dependencies array
```
humphd, line 144: This may become an issue with the work that @kliu57 is doing to allow switching providers.  Let's add them, it won't cause a lot of re-renders I don't think.

src/lib/ai.ts
```diff
@@ -293,6 +293,14 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
           : "auto"
         : undefined,
+
+    /*
+     Got this value from error message of OpenAI's gpt-4-vision-preview max_tokens
+     by setting a very large number,
+     After setting this value, seems no cutoff
+     Tested on version openai@4.24.7, without max_tokens response still cutoff
+     */
+    max_tokens: model.supportsImages ? 4096 : undefined,
```
humphd, line 303: I feel like this may come back to burn us.  Let's try to remember that we did this...

humphd (changes requested):
Amazing.  A couple small things, and this is good to go from my perspective.

I'm so happy to see this finished, @mingming-ma!  Thank you for following through on what you started in the fall.  This is an epic feature.  People are going to be blown away.

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: Ohh, yeah, this is the last issue to fix I believe 😄 . When using openrouter, mode changes. 

I updated a little bit to this function
```
  get supportsImages() {
    return this.name.includes("vision");
  }
```
It may still goes to the chatgpt4-vision mode. Do you mean we add all the support mode in `supportsImages`?

src/lib/ai.ts
mingming-ma: True. Will check later the newest version frequently.

mingming-ma:
@humphd  Thanks a bunch for all the feedback! I'm also very excited to tackle this and it's my biggest PR yet! 😄 Luckily, it will be landed soon!

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: OHHH, understand, let me update the dependencies

humphd: I don't understand why this is resolved, it doesn't appear that you fixed this?  Please don't resolve if things aren't done.

humphd (changes requested):

mingming-ma:
@humphd It's strange, I've commit a new one in [f679e4e] so I made that solved. Can you check that again?(https://github.com/tarasglek/chatcraft.org/pull/286/commits/f679e4e8ddc122a24abd71bc59009f6be8107451)

image

humphd:
@mingming-ma bizarre, I wonder if the GitHub UI was lagging.  It wasn't there when I reviewed.  I apologize.  Thanks for drawing my attention to this.

humphd (approved):
Ship this 🚢 🚢 🚢!

mingming-ma:
No worries! Let's make sure all good to land!

humphd:
@mingming-ma when you're happy, please merge so we can avoid rebase issues.

PR #287, opened 2023-11-13
System Prompt UX updates

kosty:
Forked another branch from existing backend changes branch. With a proposed changes to UX. Here's the complete list 

- Changes to DB table structure
- Data migration (with backward compatibility)
- Changes to System Prompt storage
- Changes to UX
   - Allow selecting from starred System Prompts ONLY in edit mode
   - Allow to star system prompt ONLY in non-edit mode


kosty:
@tarasglek how about this UX "change" it's really minor, yet it kinda separates editing text from storing prompts

tarasglek:
I tried this, I can't edit system prompt anymore

Issue #266, opened 2023-10-19
Configuration/Settings model parameters


hypzz:
I have a feature request: It would be useful for me to be able to configure the temperature and presence/frequency penalties sent to the model when generating. These parameters are described in detail in the following link: [Parameter Details](https://platform.openai.com/docs/guides/gpt/parameter-details)

I would love to have the ability to tweak these parameters on a per-chat basis, but even having them globally configurable in the user settings would be acceptable.

[By the way, I am really enjoying ChatCraft so far... thank you!]

tarasglek:
Hi, thanks for feedback. Would you be interested in contributing this?

humphd:
@hypzz happy to give guidance as needed.  But also, don't be afraid to use ChatCraft for help; that's what we do ;)

PR #288, opened 2023-11-14
Temperature setting

hypzz:
Adds a global setting that allows control over temperature sent to the model. Addresses #266

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -2,6 +2,7 @@ import { ChangeEvent, RefObject, useCallback, useEffect, useRef, useState } from
 import { useCopyToClipboard, useDebounce } from "react-use";
 import {
   Button,
+  Box,
   FormControl,
   FormLabel,
   FormHelperText,
@@ -18,6 +19,10 @@ import {
   VStack,
   Stack,
   Select,
+  Slider,
+  SliderTrack,
+  SliderFilledTrack,
+  SliderThumb,
   Kbd,
   Checkbox,
   Link,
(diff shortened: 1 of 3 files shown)


src/components/PreferencesModal.tsx
```diff
@@ -61,6 +61,8 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
   const inputRef = useRef<HTMLInputElement>(null);
   const provider = settings.apiUrl === OPENAI_API_URL ? "OpenAI" : "OpenRouter.ai";
   const [isApiKeyInvalid, setIsApiKeyInvalid] = useState(false);
+  const [temperature, setTemperature] = useState(settings.temperature || 0);
+  const [isTemperatureValid, setIsTemperatureValid] = useState(true);
```
humphd, line 65: Let's not do this, and instead enforce valid limits when we update `settings`.

src/components/PreferencesModal.tsx
```diff
@@ -61,6 +61,8 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
   const inputRef = useRef<HTMLInputElement>(null);
   const provider = settings.apiUrl === OPENAI_API_URL ? "OpenAI" : "OpenRouter.ai";
   const [isApiKeyInvalid, setIsApiKeyInvalid] = useState(false);
+  const [temperature, setTemperature] = useState(settings.temperature || 0);
```
humphd, line 64: Let's put this in `settings` and not bother with extra state.  The defaults for `settings` can return `0`.

src/components/PreferencesModal.tsx
```diff
@@ -290,6 +306,22 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               </FormHelperText>
             </FormControl>
 
+            <FormControl isInvalid={!isTemperatureValid}>
+              <FormLabel>Temperature</FormLabel>
+              <Input
```
humphd, line 311: What do you think about using a `<Slider />` for this? https://chakra-ui.com/docs/components/slider

humphd (changes requested):
Thanks for doing this! I've made some comments, but overall I think you're heading in the right direction.  Let's reduce the state complexity and make invalid `temperature` values impossible.

humphd (approved):
This looks great! I've always thought we should add this, thanks for suggesting and implementing it, @hypzz.  Great to have you on the team.

humphd:
Shipped on chatcraft.org

PR #290, opened 2023-11-15
starred System Prompts table

kosty:
DB changes to support starred table

Minimal UI updates to make use of the new table

### DIFF
--- a/src/lib/ChatCraftStarredSystemPrompt.ts
+++ b/src/lib/ChatCraftStarredSystemPrompt.ts
@@ -1,0 +1,57 @@
+import db, { type ChatCraftStarredSystemPromptTable } from "./db";
+
+export class ChatCraftStarredSystemPrompt {
+  text: string;
+  title: string;
+  date: Date;
+  usage: number;
+
+  constructor({
+    text,
+    title,
+    date,
+    usage,
+  }: {
+    text: string;
+    title?: string;
+    date?: Date;
+    usage?: number;
+  }) {
+    this.text = text;
+    this.title = title ?? text.split("\n")[0].substring(0, 42);
+    this.date = date ?? new Date();
+    this.usage = usage ?? 1;
+  }
+
+  clone() {
+    return new ChatCraftStarredSystemPrompt({
+      text: this.text,
+    });
+  }
+
+  async save() {
+    return db.starred.get(this.text).then((starred) => {
+      if (starred === undefined) {
+        return db.starred.put(this.toDB());
+      }
+      return starred.text;
+    });
+  }
+
+  remove() {
+    return db.starred.delete(this.text);
+  }
+
+  static exists(text: string): Promise<boolean> {
+    return db.starred.get(text).then((entry) => entry !== undefined);
+  }
+
+  toDB(): ChatCraftStarredSystemPromptTable {
+    return {
+      text: this.text,
+      title: this.title,
+      date: this.date,
+      usage: this.usage,
+    };
+  }
+}
--- a/src/lib/ChatCraftStarredSystemPrompt.ts
+++ b/src/lib/ChatCraftStarredSystemPrompt.ts
@@ -11,7 +11,7 @@ export class ChatCraftStarredSystemPrompt {
     this.usage = usage ?? 1;
   }
 
-  title() {
+  get title() {
     return this.text.split("\n")[0].substring(0, 42);
   }
 
@@ -26,12 +26,12 @@ export class ChatCraftStarredSystemPrompt {
   }
 
   async save() {
-    return db.starred.get(this.text).then((starred) => {
-      if (starred === undefined) {
-        return db.starred.put(this.toDB());
-      }
-      return starred.text;
-    });
+    const starred = await db.starred.get(this.text);
+    if (!starred) {
+      return db.starred.put(this.toDB());
+    }
+
+    return starred.text;
   }
 
   static async delete(text: string) {
@@ -43,8 +43,8 @@ export class ChatCraftStarredSystemPrompt {
     return db.starred.delete(text);
   }
 
-  static exists(text: string): Promise<boolean> {
-    return ChatCraftStarredSystemPrompt.find(text).then((entry) => entry !== undefined);
+  static exists(text: string) {
+    return ChatCraftStarredSystemPrompt.find(text).then(Boolean);
   }
 
   toDB(): ChatCraftStarredSystemPromptTable {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,6 +30,7 @@
     "jose": "^4.15.4",
     "lodash-es": "^4.17.21",
     "mermaid": "^10.5.1",
+    "murmurhash-js": "^1.0.0",
```
humphd, line 33: This hasn't been updated in 10 years.  Can we use the built-in stuff that ships in browsers? https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

src/lib/db.ts
```diff
@@ -40,11 +41,20 @@ export type ChatCraftFunctionTable = {
...
+  title: string;
+  usage: number;
+};
+
 class ChatCraftDatabase extends Dexie {
   chats: Table<ChatCraftChatTable, string>;
   messages: Table<ChatCraftMessageTable, string>;
   shared: Table<SharedChatCraftChatTable, string>;
   functions: Table<ChatCraftFunctionTable, string>;
+  starred: Table<ChatCraftStarredSystemPromptTable, string>;
```
humphd, line 57: Type issue here

humphd (changes requested):

package.json
kosty: Thanks for pointing this out! Love how sha-256 is available out of the box and chatcraft codebase is **not** involved with picking seed.

kosty: @humphd async nature of `crypto.subtle.digest` turned out to much for Dexie's transactions. Eventually had to go back to `text` as PK

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,57 @@
...
+  async save() {
+    return db.starred.get(this.text).then((starred) => {
+      if (starred === undefined) {
+        return db.starred.put(this.toDB());
+      }
+      return starred.text;
+    });
+  }
+
+  remove() {
```
humphd, line 41: Make this `static`, rename `delete` to mach how we do it everywhere else in the tree.

src/lib/db.ts
```diff
@@ -107,11 +120,40 @@ class ChatCraftDatabase extends Dexie {
     this.version(7).stores({
       messages: "id, date, chatId, type, model, user, text, versions, starred",
     });
+    this.version(8)
+      .stores({
+        messages: "id, date, chatId, type, text, starred", // Keep the old table and data for now
```
humphd, line 125: Can we remove from the old table?  Why keep two copies of this?

humphd (changes requested):

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,66 @@
...
+    date?: Date;
+    usage?: number;
+  }) {
+    this.text = text;
+    this.title = title ?? text.split("\n")[0].substring(0, 42);
+    this.date = date ?? new Date();
+    this.usage = usage ?? 1;
+  }
+
+  // Find in db or load via URL
```
humphd, line 26: Remove the `load via URL` from this comment.

src/lib/db.ts
```diff
@@ -107,11 +120,48 @@ class ChatCraftDatabase extends Dexie {
...
+              date: record.date,
+              title: record.text.split("\n")[0].substring(0, 61),
+              usage: 1,
+            });
+          });
+        await tx
+          .table("messages")
+          .where({ type: "system" })
+          .filter((message) => message.starred)
+          .delete();
```
humphd, line 153: I don't think we want to delete them (that will break old chats).  I think we want to remove the `.starred` property from them, and pull that out of the type.

src/lib/db.ts
```diff
@@ -107,11 +120,48 @@ class ChatCraftDatabase extends Dexie {
...
+      });
+    // Version 9 Migration - removes .starred from messages table
+    this.version(9).stores({
+      messages: "id, date, chatId, type, text",
```
humphd, line 157: NOTE: this is only removing the index.  I think we can do this in migration 8 above as part of the upgrade.

humphd (changes requested):

src/lib/ChatCraftStarredSystemPrompt.ts
```diff
@@ -0,0 +1,66 @@
...
+    date,
+    usage,
+  }: {
+    text: string;
+    title?: string;
+    date?: Date;
+    usage?: number;
+  }) {
+    this.text = text;
+    this.title = title ?? text.split("\n")[0].substring(0, 42);
```
humphd, line 21: Because of the work you want to do in #292, let's make this dynamic (i.e., `get title() {...}`) so we can easily adjust the number use for `42` here, in case you want more/less once you do the UX.

humphd (changes requested):

src/lib/ChatCraftStarredSystemPrompt.ts
kosty: great point! was wondering how to go about it, and love your take on it.

src/lib/db.ts
kosty: Not sure how do we go about this. `version(8)` assumes the index on `starred` is there, to perform the migration. So in `version(8)` I could use `.modify({starred: undefined})` or something similar, which will "unset" `starred` attribute. Yet it looks like we still would need `version(9)` to remove an index to an non existing field, if my reading of Dexie docs is right.

src/lib/db.ts
kosty: I did some trial-testing, and it seemed to work without those starred messages, but I agree it's much safer with them in place. Let's just keep them.

src/lib/db.ts
kosty: Not sure if latter [comment ](https://github.com/tarasglek/chatcraft.org/pull/290#discussion_r1395610048) counters this point or I am missing the point. I like the idea of keeping messages and removing attributes only - safer for user data. Not loosing data is a good thing 😄

src/lib/db.ts
kosty: @humphd submitted my proposed approach, looking forward to your feedback

humphd:
@kosty I've made some fixes and cleaned up the types a bit.  See new commit.

kosty:
looks great to me, learned quite a bit JS/TS while doing this. Neat trick with `get title` and `find(text).then(Boolean)`

PR #293, opened 2023-11-16
Fix type error while saving edits to System Prompt

### LINKED ISSUE #291: Type error while saving edits to System Prompt
Error seems to be pointing to this line 

https://github.com/tarasglek/chatcraft.org/blob/0b0ee516006a825ae18606b2d2390e4b55e5ddb3/src/components/Message/MessageBase.tsx#L207

`TypeError: Cannot read properties of null (reading 'removeAttribute')` is thrown like so

Screenshot 2023-11-15 at 9 04 56 AM

DukeManh:
Fix #291 and add keyboard event handler hook to reuse keyboard event check logic

### DIFF
--- a/src/components/PromptForm/MobilePromptForm.tsx
+++ b/src/components/PromptForm/MobilePromptForm.tsx
@@ -3,12 +3,12 @@ import { Box, chakra, Flex } from "@chakra-ui/react";
 import AutoResizingTextarea from "../AutoResizingTextarea";
 
 import { useSettings } from "../../hooks/use-settings";
-import { isMac, isWindows } from "../../lib/utils";
 import NewButton from "../NewButton";
 import MicIcon from "./MicIcon";
 import { isTranscriptionSupported } from "../../lib/speech-recognition";
 import PromptSendButton from "./PromptSendButton";
 import AudioStatus from "./AudioStatus";
+import { useKeyDownHandler } from "../../hooks/use-key-down-handler";
 
 type MobilePromptFormProps = {
   forkUrl: string;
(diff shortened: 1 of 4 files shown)


DukeManh:
I noticed 'Cmd + Enter' on message edit form wasn't working on MAC due to my previous PR. I fixed that and added a hook to extract the keyboard event check logic.

src/components/PromptForm/MobilePromptForm.tsx
```diff
@@ -3,12 +3,12 @@ import { Box, chakra, Flex } from "@chakra-ui/react";
 import AutoResizingTextarea from "../AutoResizingTextarea";
 
 import { useSettings } from "../../hooks/use-settings";
-import { isMac, isWindows } from "../../lib/utils";
 import NewButton from "../NewButton";
 import MicIcon from "./MicIcon";
 import { isTranscriptionSupported } from "../../lib/speech-recognition";
 import PromptSendButton from "./PromptSendButton";
 import AudioStatus from "./AudioStatus";
+import { useKeyDownHandler } from "../../hooks/use-key-down-handler";
```
humphd, line 11: Do we need this in mobile?  I guess people could use a keyboard with their small screen.

src/hooks/use-key-down-handler.ts
```diff
@@ -0,0 +1,25 @@
+import { useCallback, KeyboardEvent } from "react";
```
humphd, line 1: This is great, thanks for pulling this out.

humphd (approved):
Looks great, land it when you're ready

src/components/PromptForm/MobilePromptForm.tsx
DukeManh: Yes I think it's for mobiles and small or resized screens.

Issue #294, opened 2023-11-16
run button doesn't like ES6 modules


tarasglek:
https://chatcraft.org/c/tarasglek/d6Lre07Go-Y2jvZ53f9x4

Error:
```text
TypeError: Failed to fetch dynamically imported module: https://chatcraft.org/assets/browser-e187e7f8.js
TypeError: Failed to fetch dynamically imported module: https://chatcraft.org/assets/browser-e187e7f8.js
```

humphd:
This is not a regression, it's `eval()` not liking module syntax:

```js
eval("export default function f() {} \n f();");
```

```
node test.js
undefined:1
export default function f() {
^^^^^^

SyntaxError: Unexpected token 'export'
    at file:///Users/humphd/repos/chatcraft.org/test.js:1:1
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)

Node.js v20.9.0
```

We could make it work by using a `Blob` to create a "file" and then get a URL we can dynamically import; but we wouldn't know how to call it (e.g., what's the export we use?).

I could make this particular case work, but I'm not sure how to fix this in a way that will always work for arbitrary module code.

PR #295, opened 2023-11-16
logging intercept and more careful eval of es6

tarasglek:
this fixes one of the problems i had in #294.

- better fallback to ES6
- better code for checking missing export default
- handles weird edge case of objects that are not serializable into JSON, yet toString()able like functions
- lack of log intercept bothered me a lot now:

image

### DIFF
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -84,8 +84,7 @@ function CodeHeader({
       ret = result;
     } catch (error: any) {
       ret = formatAsCodeBlock(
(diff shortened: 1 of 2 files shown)


tarasglek:
Wow, this is a huge improvement for writing code.

https://taras-code.console-overthinker-dev.pages.dev/c/chatcraft_dev/twn11pdCbkYNvZOKH6-86

tarasglek:
landing cos it's too painful to live without this now, can address any issues in followup

humphd:
Epic hackery.

PR #303, opened 2023-12-07
nuke "Just show me the code"

tarasglek:
Proper way to do this is to edit prompt. We didn't have that before.

This feature resulted in me having 100s of saved prompts with multiple copies of just show me the code


tarasglek:
@humphd is this safe to remove a preference like this? Will this mess up IDB?

humphd (approved):
This is safe, doesn't affect db.

This is a good fix.

PR #304, opened 2023-12-11
Update ai.ts

### LINKED ISSUE #227: Refactor OpenAI's response handling to pick the right handler at runtime
The issue was discussed here: https://github.com/tarasglek/chatcraft.org/pull/220#discussion_r1299371774

Amir-Helali:
Updated the OpenAI's response handling to pick the right handler based on the `streaming` flag. This is done through an interface function which has 2 separate implementations: 1 for handling streaming responses and 1 for handling non-streaming responses. And the `responseHandler` which is in charge of assigning the appropriate function implmenetation based on the value of the `streaming` flag.

Fixes #227

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -10,6 +10,13 @@ import { getReferer } from "./utils";
 import { getSettings, OPENAI_API_URL } from "./settings";
 
 import type { Tiktoken } from "tiktoken/lite";
+import {
+  ChatCompletion,
+  ChatCompletionChunk,
+  ChatCompletionCreateParamsStreaming,
+  CreateChatCompletionRequestMessage,
+} from "openai/resources";
+import { Stream } from "openai/streaming";
 
 export const usingOfficialOpenAI = () => getSettings().apiUrl === OPENAI_API_URL;
 
@@ -121,6 +128,10 @@ export const transcribe = async (audio: File) => {
   });
   return transcription.text;
 };
+
+interface ResponseHandlerStrategy {
+  (response: any): Promise<any>;
+}
 
 export const chatWithLLM = (messages: ChatCraftMessage[], options: ChatOptions = {}) => {
   const {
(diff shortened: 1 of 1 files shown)


src/lib/ai.ts
```diff
@@ -297,49 +306,45 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+    const content = buffer.join("");
+    return handleOpenAIResponse(content, functionName, functionArgs);
+  };
+
+  const handleNonStreamingResponse: ResponseHandlerStrategy = async (response: ChatCompletion) => {
+    const { content, functionName, functionArgs } = parseOpenAIResponse(response);
+    return handleOpenAIResponse(content, functionName, functionArgs);
+  };
+
+  const responseHandler: ResponseHandlerStrategy = streaming
```
humphd, line 330: Rename this to a verb (it's active, it will "handle the response" vs. passive "responseHandler").  Also, infer the type:

```suggestion
  const handleResponse = streaming
```

src/lib/ai.ts
```diff
@@ -10,6 +10,11 @@ import { getReferer } from "./utils";
 import { getSettings, OPENAI_API_URL } from "./settings";
 
 import type { Tiktoken } from "tiktoken/lite";
+import {
```
humphd, line 13: This formatting looks wrong (run `pnpm prettier` to fix).

Also, these are types, so you should import them as such:

`import type { ChatCompletion, ChatCompletionChunk } from "openai/streaming";`

Move your type imports down below the others and group together:

```ts
import foo from "foo"

import type { bar } from "bar"
```

src/lib/ai.ts
```diff
@@ -10,6 +10,11 @@ import { getReferer } from "./utils";
...
+  ChatCompletion,
+  ChatCompletionChunk,
+} from "openai/resources";
+import { Stream } from "openai/streaming";
```
humphd, line 17: Move this up with the other imports that are not `type`

src/lib/ai.ts
```diff
@@ -122,6 +127,10 @@ export const transcribe = async (audio: File) => {
   return transcription.text;
 };
 
+interface ResponseHandlerStrategy {
```
humphd, line 130: A few comments about this:

1. Try to avoid naming new types in TypeScript when they can be inferred.  In this case, you are creating an interface for two functions that already return the same type; that is, both functions you are wrapping return `handleOpenAIResponse()` which is of type `Promise<ChatCraftAiMessage | ChatCraftFunctionCallMessage>`.
2. Whenever you see `any` in TypeScript, it's often a sign that you need to do more work.  Here you are losing the type information I just mentioned above by widening the type to `any` from `ChatCraftAiMessage | ChatCraftFunctionCallMessage`.  Keep it narrow.

If you avoid defining this interface at all, you skip these problems.  I'll show you how below.

src/lib/ai.ts
```diff
@@ -297,49 +306,45 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
-        return handleOpenAIResponse(content, functionName, functionArgs);
-      })
-      .catch((err) => {
-        return handleError(err);
-      })
-      .finally(() => {
-        removeEventListener("keydown", handleCancel);
-      });
-  }
+  const handleStreamingResponse: ResponseHandlerStrategy = async (
```
humphd, line 309: You can avoid the return type here and let TS infer it from `return handleOpenAIResponse(content, functionName, functionArgs)`.

src/lib/ai.ts
```diff
@@ -297,49 +306,45 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+        parsedData.functionName,
+        parsedData.functionArgs
+      );
+    }
+
+    const content = buffer.join("");
+    return handleOpenAIResponse(content, functionName, functionArgs);
+  };
+
+  const handleNonStreamingResponse: ResponseHandlerStrategy = async (response: ChatCompletion) => {
```
humphd, line 325: Same here, drop the return type and infer.  You will end up with the same type for both.

src/lib/ai.ts
```diff
@@ -297,49 +306,45 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+  const responseHandler: ResponseHandlerStrategy = streaming
+    ? handleStreamingResponse
+    : handleNonStreamingResponse;
+
+  const responsePromise = openai.chat.completions
+    .create(
+      chatCompletionParams as OpenAI.Chat.ChatCompletionCreateParamsStreaming,
+      chatCompletionReqOptions
+    )
+    .then(async (response) => {
```
humphd, line 339: Simplify this. Both functions take a `response` and return a `Promise`, so you can simply give the function name to use:

```ts
.then(handleResponse)
```

src/lib/ai.ts
```diff
@@ -297,49 +306,45 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
...
+      return responseHandler(response);
+    })
+    .catch((err) => {
```
humphd, line 342: Same deal here:

```ts
.catch(handleError)
```

humphd (changes requested):

Amir-Helali:
@humphd Thanks for the feedback. Sorry I've been a bit busy with all the end of the semester work, I will address all the points you provided in your feedback once I am done with the school work.

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: Done

src/lib/ai.ts
Amir-Helali: In this case I used parameter type: `any` for `handleNonStreamingResponse()` because `parseOpenAiResponse()` takes a `ChatCompletion`, and if I had used parameter type `ChatCompletion` for `handleNonStreamingResponse()` then I'd got error in `.then(handleResponse)` because the response would be of type `Stream<ChatCompletionChunk>`. Let me know if this needs changing to make the typing narrower.

src/lib/ai.ts
Amir-Helali: Done

humphd (approved):
Looks good, nice simplification.

PR #306, opened 2023-12-13
Added validation for OpenRouter API key

### LINKED ISSUE #230: Use OpenRouter User Limits API to check key
When a user enters an API key for OpenRouter, we need a way to check that it's good.  It looks like we could use https://openrouter.ai/docs#limits

kliu57:
Closes #230 

Added validation for OpenRouter API key

**Current code (before PR):**
- No validation for OpenRouter API key

**PR Code changes:**
- In `src/lib/ai.ts`, created new function `validateApiKey()` which validates both OpenAI and OpenRouter keys, by calling `validateOpenAiApiKey()` or `validateOpenRouterApiKey()`
- In `src/lib/ai.ts`, create new function `validateOpenRouterApiKey()`, which uses response from https://openrouter.ai/docs#limits to check if OpenRouter API key is valid
- In `src/components/Message/AppMessage/Instructions.tsx`, replaced call of `validateOpenAiApiKey()` with `validateApiKey()`
- In `src/components/PreferencesModal.tsx`, replaced call of `validateOpenAiApiKey()` with `validateApiKey()`

**Testing:**

Invalid OpenRouter API key:

invalidOpenRouter1

invalidOpenRouter2

invalidOpenRouter3

Valid OpenRouter API key:

validOpenRouter

**Regression Testing:**

Invalid OpenAI API key:

invalidOpenAI1

invalidOpenAI2

invalidOpenAI3

Valid OpenAI API key:

validOpenAI

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -387,20 +387,20 @@ export async function validateOpenRouterApiKey(apiKey: string) {
     throw new Error(`${res.status} ${await res.text()}`);
   }
 
-  return res;
+  return true;
 }
 
 export async function validateApiKey(apiKey: string) {
-  const usingOpenAI = usingOfficialOpenAI();
-  const usingOpenRouter = usingOfficialOpenRouter();
-
-  if (usingOpenAI) {
+  if (usingOfficialOpenAI()) {
     return validateOpenAiApiKey(apiKey);
-  } else if (usingOpenRouter) {
-    return !!(await validateOpenRouterApiKey(apiKey));
-  } else {
-    return false;
   }
+
+  if (usingOfficialOpenRouter()) {
+    return validateOpenRouterApiKey(apiKey);
+  }
+
+  // If not either of those providers, something is wrong
+  throw new Error("unexpected provider, not able to validate api key");
 }
 
 // Cache this instance on first use
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -7,11 +7,12 @@ import {
 import { ChatCraftModel } from "./ChatCraftModel";
 import { ChatCraftFunction } from "./ChatCraftFunction";
 import { getReferer } from "./utils";
-import { getSettings, OPENAI_API_URL } from "./settings";
+import { getSettings, OPENAI_API_URL, OPENROUTER_API_URL } from "./settings";
 
 import type { Tiktoken } from "tiktoken/lite";
 
 export const usingOfficialOpenAI = () => getSettings().apiUrl === OPENAI_API_URL;
+export const usingOfficialOpenRouterAI = () => getSettings().apiUrl === OPENROUTER_API_URL;
 
 const createClient = (apiKey: string, apiUrl?: string) => {
   // If we're using OpenRouter, add extra headers
@@ -353,16 +354,11 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
 export async function queryModels(apiKey: string) {
   const { apiUrl } = getSettings();
   const usingOpenAI = usingOfficialOpenAI();
+  const usingOpenRouterAI = usingOfficialOpenRouterAI();
   const { openai } = createClient(apiKey, apiUrl);
 
   try {
-    const models = [];
-
-    if (usingOpenAI) {
-      for await (const page of openai.models.list()) {
-        models.push(page);
-      }
-    } else {
+    if (usingOpenRouterAI) {
       // Use response from https://openrouter.ai/docs#limits to check if API key is valid
       const res = await fetch(`https://openrouter.ai/api/v1/auth/key`, {
         method: "GET",
@@ -374,6 +370,11 @@ export async function queryModels(apiKey: string) {
       if (!res.ok) {
         throw new Error(`${res.status} ${await res.text()}`);
       }
+    }
+
+    const models = [];
+    for await (const page of openai.models.list()) {
+      models.push(page);
     }
 
     return models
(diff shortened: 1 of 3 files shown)


src/lib/ai.ts
```diff
@@ -357,8 +357,23 @@ export async function queryModels(apiKey: string) {
...
-    for await (const page of openai.models.list()) {
-      models.push(page);
+
+    if (usingOpenAI) {
+      for await (const page of openai.models.list()) {
+        models.push(page);
+      }
+    } else {
+      // Use response from https://openrouter.ai/docs#limits to check if API key is valid
+      const res = await fetch(`https://openrouter.ai/api/v1/auth/key`, {
```
tarasglek, line 367: Note, openrouter is a step toward generic openai compat. Would be good to check if the completion endpoint is openrouter.ai before requiring/initiating the verification

tarasglek: Example of another endpoint is https://docs.endpoints.anyscale.com/guides/models/

kliu57: I have moved the verification code to inside an if block that will only be triggered when apiUrl is for OpenRouter, please let me know if it is ok

tarasglek:
I like it, thank you for improving this!

tarasglek:
@kliu57 I added you as a collaborator. For all new work, please push as a branch to chatcraft.org repo, then we can benefit from cloudflare branch deployments.

kliu57:
> @kliu57 I invited you as a collaborator, I hope you accept. For all new work, please push as a branch to chatcraft.org repo, then we can benefit from cloudflare branch deployments.

Sure, I can do that for future PRs. Do you need to recreate my current PRs? (this one or PR#308)

tarasglek:
@humphd ok to accept this one as is?

humphd (changes requested):

src/lib/ai.ts
kliu57: I have created the generic `validateApiKey()` function and and replaced calls to `validateOpenAiApiKey()` with called to `validateApiKey()`. Please see the updated PR description for full code details. Please review again at your convenience.

Also, I noticed in [`src/hooks/use-models.tsx`](https://github.com/tarasglek/chatcraft.org/blob/main/src/hooks/use-models.tsx) there is a call to `queryModels()`. Now that my OpenRouter validation is not done inside of `queryModels()` anymore, this particular call to `queryModels()` will not be able to validate any OpenRouter keys. Is this something we need to fix?

src/lib/ai.ts
```diff
@@ -373,6 +374,35 @@ export async function validateOpenAiApiKey(apiKey: string) {
...
+}
+
+export async function validateApiKey(apiKey: string) {
+  const usingOpenAI = usingOfficialOpenAI();
+  const usingOpenRouter = usingOfficialOpenRouter();
+
+  if (usingOpenAI) {
+    return validateOpenAiApiKey(apiKey);
+  } else if (usingOpenRouter) {
+    return !!(await validateOpenRouterApiKey(apiKey));
```
humphd, line 400: Do this `boolean` cast in `validateOpenRouterApiKey` instead, so both functions return the same type, namely, `boolean`.  e.g., return `true` vs. `res` above:

```ts
  if (!res.ok) {
    throw new Error(`${res.status} ${await res.text()}`);
  }

  return true;
```

src/lib/ai.ts
```diff
@@ -373,6 +374,35 @@ export async function validateOpenAiApiKey(apiKey: string) {
...
+
+  if (!res.ok) {
+    throw new Error(`${res.status} ${await res.text()}`);
+  }
+
+  return res;
+}
+
+export async function validateApiKey(apiKey: string) {
+  const usingOpenAI = usingOfficialOpenAI();
```
humphd, line 394: Don't bother with variables, you can call the the functions in the `if` below

src/lib/ai.ts
```diff
@@ -373,6 +374,35 @@ export async function validateOpenAiApiKey(apiKey: string) {
...
+  const usingOpenRouter = usingOfficialOpenRouter();
+
+  if (usingOpenAI) {
+    return validateOpenAiApiKey(apiKey);
+  } else if (usingOpenRouter) {
```
humphd, line 399: No else after return (not needed, since you've already exited, and helps keep code left-aligned and easier to read):

```ts
if (usingOfficialOpenAI()) {
  return validateOpenAiApiKey(apiKey);
}

if (usingOfficialOpenRouter()) {
  return validateOpenRouterApiKey(apiKey);
}

// If not either of those providers, something is wrong
throw new Error('unexpected provider, not able to validate api key');
```

humphd (changes requested):
A few small things, this is looking good.

kliu57:
@humphd Made the three code fixes as suggested.

Also, I noticed in [`src/hooks/use-models.tsx`](https://github.com/tarasglek/chatcraft.org/blob/main/src/hooks/use-models.tsx) there is a call to `queryModels()`. Now that my OpenRouter validation is not done inside of `queryModels()` anymore, this particular call to `queryModels()` will not be able to validate any OpenRouter keys. Is this something of concern?

humphd:
@kliu57 This needs to be rebased to pick up the change we just landed in `src/lib/ai.ts`.

I think the call to `queryModels() in `src/hooks/use-models.tsx` doesn't need to be validated, since it will fail if the key is wrong anyway. We only really care about warning people and validating when we are setting the key during setup.

humphd (approved):

humphd:
Great work @kliu57.  I hope you'll keep working ChatCraft in the future.

PR #308, opened 2023-12-14
Clear stored api key when changing providers in user settings

### LINKED ISSUE #307: When changing API key in User Settings, do not send stored key
Sure, go for it.  Another thing you could work on that's related is that when the user is entering a new API key, let's say for OpenAI vs. OpenRouter, we don't want to send the stored key to the other provider.

There's a bunch of things we could improve in how this is all done.

_Originally posted by @humphd in https://github.com/tarasglek/chatcraft.org/issues/230#issuecomment-1852606536_

**Issue description** (Attempting to switch from OpenRouter to OpenAI):

When going to User Settings to change the API URL from the dropdown, do not send the stored key to the newly selected API URL provider the moment we select from the dropdown.

**Issue screenshots:**

sendIssue1
sendIssue2

**Expected behaviour:**

Please help me to clarify what should be the expected behaviour.

kliu57:
Closes #307

User Settings -> Change provider using API URL dropdown -> Clears the stored API key

**Current code (before PR):**
- In `src/components/PreferencesModal.tsx`, change event handler for the API URL dropdown does not clear the stored API key, thereby sending the old stored key to the newly selected provider

**PR Code changes:**
- In `src/components/PreferencesModal.tsx`, change event handler for the API URL dropdown clears the stored API key (just like the Remove button does)
- The resulting behaviour is the same as the Remove Button (the only difference is the provider is changed to the one we selected)

**Testing:**

Changing from OpenRouter -> OpenAI

test1
test2
test3

Changing from OpenAI -> OpenRouter

test3
test4

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -164,7 +164,9 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <FormLabel>API URL</FormLabel>
               <Select
                 value={settings.apiUrl}
-                onChange={(e) => setSettings({ ...settings, apiUrl: e.target.value })}
+                onChange={(e) =>
+                  setSettings({ ...settings, apiUrl: e.target.value, apiKey: undefined })
+                }
               >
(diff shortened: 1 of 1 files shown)


tarasglek:
lets try this

PR #329, opened 2024-01-13
Add rjwignar to .sops.yaml config, re-encrypt keys.enc.yaml

humphd:
Added @rjwignar's age key to the list of allowed users of `keys.enc.yaml`.  @rjwignar, can you test this for me?

```sh
git fetch origin
git checkout humphd/sops-rjwignar
sops -d sops/keys.enc.yaml
```

This should produce a decrypted key for you.

@tarasglek hopefully I've done this correctly.


rjwignar:
I ran the above commands in git bash from the project root directory (under humphd/sops-rjwignar) and get 8 blocks of the following text (1 for each AGE key in the YAML file):
image

In the same git bash session I tried initializing the SOPS_AGE_KEY env variable as per [README.sops.md](https://github.com/tarasglek/chatcraft.org/blob/main/README.sops.md#user-decrypting-secrets) and then running sops but I get this error instead (8 times, once for each AGE key):
 
image

humphd:
To confirm, you're doing the following in a Unix shell?

```sh
export SOPS_AGE_KEY=$(ssh-to-age -private-key < ~/.ssh/id_ed25519)
sops -d keys.enc.yaml
```

rjwignar:
I used
```bash
export SOPS_AGE_KEY=scripts/sops_age_key.sh
sops -d sops/keys.enc.yaml
```

I just tried the export command you sent, it didn't work at first as my key is passphrase-protected.
However, after running the following I successfully to decode my API Key.:
```bash
read -s SSH_TO_AGE_PASSPHRASE; export SSH_TO_AGE_PASSPHRASE
[enter passphrase then hit ENTER]
ssh-to-age -private-key -i $HOME/.ssh/id_ed25519 -o key.txt
export SOPS_AGE_KEY=$(ssh-to-age -private-key < ~/.ssh/id_ed25519)
sops -d sops/keys.enc.yaml
```

I don't know if the first `ssh-to-age` call was required, but I did it as suggested by the [ssh-to-age usage instructions](https://github.com/Mic92/ssh-to-age/blob/main/README.md#usage):
image

Thanks for your help in this!
Now that I have my API key, should I store it anywhere for easy access, or decode it each time I'm working with ChatCraft?

humphd:
We should update our docs, can you do a PR?

You can decrypt it on demand, store it securely in a password manager, use it in the browser with ChatCraft (it will get cached in LocalStorage)

rjwignar:
I'll have a PR ready tonight.
I'll be updating the User instructions for decrypting secrets in `README.sops.md`
image

In my experience, running the following
```bash
export SOPS_AGE_KEY=scripts/sops_age_key.sh

sops -d sops/keys.enc.yaml
```
failed to decrypt my API key.

But is that everyone's experience? I don't want to replace those instructions if they work for others (maybe it works if the user's SSH/AGE key is not passphrase-protected). For now I'll add the commands that allowed me to decrypt my key (assuming it's passphrase-protected).

humphd:
> But is that everyone's experience? I don't want to replace those instructions if they work for others (maybe it works if the user's SSH/AGE key is not passphrase-protected). For now I'll add the commands that allowed me to decrypt my key (assuming it's passphrase-protected).

Ask on Discord, and see their blogs.  I think others likely didn't use a passphrase.  My use case is also harder, since I keep my SSH Key in 1Password, and have to use the 1Password CLI `op` tool to extract it.

PR #338, opened 2024-01-14
Update README.sops.md with user instructions for encrypted key

rjwignar:
Summary
---
Based on discussion in https://github.com/tarasglek/chatcraft.org/pull/329#issuecomment-1890483618 , added instructions in README.sops.md for users decrypting secrets using an encrypted key.


humphd:
Nice. I love seeing problems turn into updated docs.  Well done, @rjwignar.

PR #344, opened 2024-01-16
Added logic for sorting the retry options

### LINKED ISSUE #316: make "retry with" a submenu
maybe even multiple levels of submenus with some fixed number of entries each. 
Here is what openrouter looks like, it's only gonna get worse:
image

Can probably get gpt to group entries for us via function call.

Rachit1313:
Fixes #316


tarasglek:
The bottom menu needs sorting too. 

image

Need to unify this sorting code somewhere. Maybe do it at the point of requesting list models endpoint?

tarasglek:
Thanks for breaking this up into separate pull req

humphd (changes requested):
Let's fix this at the source, so it works for every user of `models`, which we get from the `useModels()` hook:

https://github.com/tarasglek/chatcraft.org/blob/a24f51d0f008bb12ae04ecbeece76b696bdf927e/src/hooks/use-models.tsx#L57

If you change the way we do the sort there, it should mean we're good in all UI situations.

src/hooks/use-models.tsx
```diff
@@ -54,7 +54,8 @@ export const ModelsProvider: FC<{ children: ReactNode }> = ({ children }) => {
       isFetching.current = true;
       try {
         const models = await queryModels(apiKey).then((models) => {
-          models.sort();
+          // models.sort();
```
humphd, line 57: you can delete vs. comment-out code

humphd (changes requested):

tarasglek:
image

Still not sorting bottom menu

humphd:
Given that we ultimately want the pretty names to be sorted, let's do that in the hook (i.e., convert the models to ChatCraftModel instances, then sort on the `prettyName`) so that all consumers will have this list in the right order.

src/hooks/use-models.tsx
```diff
@@ -54,8 +54,7 @@ export const ModelsProvider: FC<{ children: ReactNode }> = ({ children }) => {
       isFetching.current = true;
       try {
         const models = await queryModels(apiKey).then((models) => {
-          // models.sort();
-          models.sort((a, b) => a.localeCompare(b));
+          models.sort();
```
humphd, line 57: Don't bother sorting here if you're going to sort again later.

src/hooks/use-models.tsx
```diff
@@ -74,7 +73,7 @@ export const ModelsProvider: FC<{ children: ReactNode }> = ({ children }) => {
   }, [settings.apiUrl, settings.apiKey]);
 
   const value = {
-    models: models || defaultModels,
+    models: models?.sort((a, b) => a.prettyModel.localeCompare(b.prettyModel)) || defaultModels,
```
humphd, line 76: Do this above on line 60 so you cache the result in state vs. doing it for every consumer:

```ts
models.sort((a, b) => a.prettyModel.localeCompare(b.prettyModel))
setModels(models);
```

humphd (changes requested):

src/hooks/use-models.tsx
Rachit1313: removed

src/hooks/use-models.tsx
Rachit1313: done thanks.

humphd (approved):
Code change looks good to me.  Get some of your colleagues to test the change at the UI level and/or put in some screenshots of it doing what you expect.

tarasglek (approved):
good to land. this will be very helpful

Rachit1313:
> Code change looks good to me. Get some of your colleagues to test the change at the UI level and/or put in some screenshots of it doing what you expect.
Screenshot 2024-01-18 at 12 27 22 PM
Screenshot 2024-01-18 at 12 27 29 PM

Rachit1313:
> good to land. this will be very helpful

Do you want me to squash the commits?

PR #346, opened 2024-01-17
working openai key again

tarasglek:
For some reason I replaced the openai key with a broken one when I added openrouter. Sorry about that. Also added fancyness so i can see decrypted sops in git log -p =D


tarasglek:
pretty cool having history of key changes in git =D

humphd:
Great thank you for fixing this

PR #348, opened 2024-01-17
Removed non-functioning code from README.sops.md

### LINKED ISSUE #347: Edit README.sops.md to remove non-functioning code
nonfunctioningcode

Confirmed with myself, @WangGithub0, @Amnish04, that the first set of code does not work, even if the user has an empty passphrase set.

Confirmed with @tarasglek that we should remove the first set of code. The second set of code works for everyone; those with an empty passphrase can simply type [Enter]

kliu57:
Fixes #347

### DIFF
--- a/README.sops.md
+++ b/README.sops.md
@@ -46,13 +46,8 @@ git diff sops/keys.enc.yaml
 
 ### User: Decrypting secrets
 
-```bash
-export SOPS_AGE_KEY=scripts/sops_age_key.sh
-
-sops -d sops/keys.enc.yaml
-```
-
 If you are using an encrypted key, you will have to first export the password in `SSH_TO_AGE_PASSPHRASE` and then export the AGE key in `SOPS_AGE_KEY`:
+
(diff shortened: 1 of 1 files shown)


WangGithub0:
LGTM

WangGithub0 (approved):

humphd:
@kliu57 feel free to "squash and merge" when you're ready to land.

Amnish04 (commented):
I agree with this.

The following command
`export SOPS_AGE_KEY=scripts/sops_age_key.sh`

was setting the environment variable as the path itself (scripts/sops_age_key.sh) instead of the script's result.

So I had to manually execute the script, take the result and set the env var before trying to decrypt. I have talked about it in this post.
https://dev.to/amnish04/joining-chatcraftorg-279f#decrypting-the-key

Can we rather replace it with the following steps instead of removing it altogether so people have an option to decrypt without necessarily having to use the external tool.
image

Issue #335, opened 2024-01-13
Formula rendering for dark mode


kosty:
Currently formulas are being rendered by https://latex.codecogs.com/svg.latex as black on transparent background making them extra hard to read in dark mode. 

Few solutions are possible:
* force white background at all the times (is there such a GET parameter?)
* force some special formula color clearly visible in either mode, vibrant yellow? ;) 
* switch rendering to white on transparent background for dark mode

PR #349, opened 2024-01-17
Add Latex Rendering with remark-math and Katex Plugins

### LINKED ISSUE #333: Fix positioning, padding, sizing of images
When rendering latex images in a message, the position, padding, and size could be improved:

Screenshot 2024-01-13 at 11 30 57 AM

Here's an example chat:

https://chatcraft.org/c/humphd/HWrYbl4QNB52j3RNkyT4k

This likely happens for all images vs. the ones I'm doing here.

mingming-ma:
Fix #333 #335 

Need to add "formula in LaTeX by enclosing it with $$ display style" or similar to mention the dollar symbol in the prompt.

[Examples](https://mingming-equation.console-overthinker-dev.pages.dev/c/chatcraft_dev/y-R7-655HWbsBgYYMpEBj)

### DIFF
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,7 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your Markdown by enclosing it with the dollar symbol rather than parentheses, for example: $$O(nlogn)$$
+- If responding with math, use KaTeX syntax in your Markdown by enclosing it with double-dollar signs rather than parentheses, for example: $$O(n\\log n)$$
 `;
 
 export const defaultSystemPrompt = () => {
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,7 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your Markdown by enclosing it in $$...$$
+- If responding with math, use KaTeX syntax in your Markdown by enclosing it with the dollar symbol rather than parentheses, for example: $$O(nlogn)$$
 `;
 
 export const defaultSystemPrompt = () => {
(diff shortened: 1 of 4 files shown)


humphd (changes requested):
We better update the default system prompt to include a line about using Latex/Katex for any math, so the model does it and we can render it.

src/lib/system-prompt.ts
```diff
@@ -10,6 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
+- formula in LaTeX by enclosing it with $$ display style
```
humphd, line 13:
```suggestion
- If responding with math, use KaTeX syntax in your Markdown by enclosing it in $$...$$
```

humphd (changes requested):

src/lib/system-prompt.ts
mingming-ma: image Seems a little bit confusing after rendering, let me try another approach

tarasglek:
nice

src/lib/system-prompt.ts
humphd: You could try these:

- If responding with math, use KaTeX syntax in your Markdown
- If responding with math, use KaTeX syntax in your Markdown by enclosing it in double-dollar signs

src/lib/system-prompt.ts
```diff
@@ -10,7 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your Markdown by enclosing it in $$...$$
+- If responding with math, use KaTeX syntax in your Markdown by enclosing it with the dollar symbol rather than parentheses, for example: $$O(nlogn)$$
```
humphd, line 13: '...with double-dollar signs...'

I like that you included an example, smart.

humphd (changes requested):

humphd (approved):

tarasglek:
This has a funny conflict with coding :(
image

image

mingming-ma:
👍 somewhere need to configure so only accept $$, let me fix that

humphd:
Or never render katex/math in code/pre blocks.

PR #351, opened 2024-01-17
Fix and add back the first sops decryption method

### LINKED ISSUE #350: Add back the fixed SOPS decryption method
PR #348 removed the first method for decrypting sops keys, as it did not work for anyone.
image

However, the reason that did not work was that the step where `SOPS_AGE_KEY` environment variable was set did not store the result of the script, but the path of the script file.
https://github.com/tarasglek/chatcraft.org/pull/348#pullrequestreview-1828027901

image

That does not mean this method doesn't work. We just need to fix the way we are setting the environment variable such that it has our private age key.

Amnish04:
This PR adds the fix for the first decryption method that was removed from [README.sops.md](https://github.com/tarasglek/chatcraft.org/blob/main/sops/README.sops.md) file.

We now make sure that the script is executed, and its **result** is stored in the `SOPS_AGE_KEY` environment variable.

```bash
export SOPS_AGE_KEY=$(bash ./scripts/sops_age_key.sh)
```

Instead of

```bash
export SOPS_AGE_KEY=./scripts/sops_age_key.sh
```
This fixes #350

### DIFF
--- a/sops/README.sops.md
+++ b/sops/README.sops.md
@@ -9,7 +9,7 @@ export SOPS_AGE_KEY=$(bash ./scripts/sops_age_key.sh)
 sops -d sops/keys.enc.yaml
 ```
 
-**Note:** Instead of storing the private age key in the environment variable, it is recommended to store the secrets in a `keys.txt` file as documented at
+**Note:** Alternatively, to not mess with the environment variables, you can also store the secrets in a `keys.txt` file as documented at
 https://github.com/getsops/sops/blob/0bceaf42b834f254cf2c6a6f61e7121de8eb9c52/README.rst#L205C1-L212C22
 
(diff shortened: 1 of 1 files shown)


tarasglek:
Add a note that instead of env var can keep a permanent age equiv of your private key as documented in https://github.com/getsops/sops/blob/0bceaf42b834f254cf2c6a6f61e7121de8eb9c52/README.rst#L205C1-L212C22

Amnish04:
@tarasglek Just pushed the changes.

tarasglek:
It's not instead of instead of and recommended, it's just an alternative to not mess with env vars :) almost there.

Amnish04:
@tarasglek Sorry about that, I've made the changes.

tarasglek:
lgtm!

PR #353, opened 2024-01-19
Fix math render conflict with coding 

mingming-ma:
Fix https://github.com/tarasglek/chatcraft.org/pull/349#issuecomment-1901090635

Now we can use $ as normal:
image

[link](https://mingming-equation-strict.console-overthinker-dev.pages.dev/c/chatcraft_dev/2GkxXdOTBG0Soha-IYGsr)

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -61,7 +61,7 @@ function Markdown({ previewCode, isLoading, onPrompt, children }: MarkdownProps)
     <ReactMarkdown
       className="message-text"
       children={children}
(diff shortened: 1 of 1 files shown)


mingming-ma:
@humphd Based on the test, it looks like the code block is all good and will not be rendered.

humphd (approved):
Nice!  Thanks for doing a fast-follow up on this bug.  Appreciated.

tarasglek:
lets see how this goes, thanks for quick fix

Issue #354, opened 2024-01-21
Add support for openai tts api


tarasglek:
As a warmup for #310 one can add a tts feature to chatcraft. 
Openai can do tts using really high quality voices: https://platform.openai.com/docs/guides/text-to-speech

We should add a speak menu entry to each message so it can be spoken by these voices. Once the message is spoken, it should also get a download speech menu so we can download the generated speech.

TTS could also allow chatcraft to respond in voice when we ask it questions using the voice feature.

This would be handy for recording voiceovers for demos on youtube, presentations, etc.

tarasglek:
We should also support using browser-local VTT and TTS ala https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API

Amnish04:
Sounds Interesting!

I tried to play with it a bit, but for some reason I can't get access the method from documentation
![image](https://github.com/tarasglek/chatcraft.org/assets/78865303/c35a3e71-2e62-4d59-bb2d-8213e4367c82)

Am I using the wrong object?
![image](https://github.com/tarasglek/chatcraft.org/assets/78865303/1f155d89-3c85-4eea-b11e-97451ef27ddf)
![image](https://github.com/tarasglek/chatcraft.org/assets/78865303/5f4b87c4-6b78-48fa-8762-6f81b3ab3309)

Amnish04:
Just found that the version of openai we are using does not support the [speech](https://platform.openai.com/docs/guides/text-to-speech?lang=node) property.

Upgrading to latest fixed it, and I can listen to generated audio.

![image](https://github.com/tarasglek/chatcraft.org/assets/78865303/6b9e988c-1c83-4bd2-a3eb-b24dbdef373d)

humphd:
@Amnish04 nice, want to turn your investigation into a PR?

One thing we'd have to do here is make this aware of different providers/models (cc @kliu57).  For example, if I'm using OpenRouter.ai vs. OpenAI for my provider and API Key, this won't work.

To start, maybe we only do this if you're using OpenAI as your provider?

Another thing that would be cool is to allow users to set the voice to use as a setting: https://platform.openai.com/docs/guides/text-to-speech/voice-options

Also, it looks like we can stream it vs. waiting to download: https://platform.openai.com/docs/guides/text-to-speech

tarasglek:
I think the model should show up in list endpoint. Can use that enable disable this feature.


Love how quickly you got an experimental result!

[...]

Issue #336, opened 2024-01-13
Change mobile interface to be easier to use. 


tarasglek:
We should change the button the left to be "burger" options menu. Should move share into it. Should also move edit system prompt into it...which would scroll up and do equiv of editing system prompt.

We should flip the way mic icon works...instead of holding to record should click it once to start recording and once to stop. This is how bing works on ios.

Should also edit mic icon to when we edit messages and system prompt. Can do that in a followup

Might also make sense to move model selection into this menu.

kosty:
![IMG_2174](https://github.com/tarasglek/chatcraft.org/assets/630193/a885a775-524f-4e1f-a353-2e3881f401d4)

side menu of the message is quite unusable on mobile when most of the time as well

humphd:
Two things I noticed this week on mobile:

<img width="341" alt="Screenshot 2024-02-18 at 12 54 12 PM" src="https://github.com/tarasglek/chatcraft.org/assets/427398/6c87b3a0-096c-4740-b3f0-74bcec88c6e6">

The Audio Recording button isn't the right size

Also, my fix for the wrapping of the placeholder text in the input element is wrong.  On my iOS device, it still wraps.

tarasglek:
@menghif I see that you have a blog. Would you be interested in writing a blog post on why and what you contributed? I'll then do a chatcraft release and link to your blog from release notes

menghif:
@tarasglek great idea, yes I will write a post. Thank you!

PR #357, opened 2024-01-22
Add Text to Speech Support

Amnish04:
Regarding issue #354 

I think I was able to implement an initial version of TTS support.

I've added an additional button in the prompt menu that allows users to select if they want to enable this feature.
image

Whenever we ask a question, the response is read using the TTS api.

I am buffering the words received from LLM stream over time and once the `threshold` is reached, I pass it to the TTS method for audio generation.
image

I've also added a `useAudioPlayer` hook that manages the playback of audio chunks by making sure they play in order.

I feel like its still very rough and lots of optimizations/changes need to be done (variable naming, code organization, performance, UI prefs) before I move on with other features like download.

One would be the ability to choose **different voices**, but I am not sure where to place that UI pref because there's hardly any real estate.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -60,8 +60,8 @@ dependencies:
     specifier: ^5.0.2
     version: 5.0.2
   openai:
-    specifier: ^4.12.4
-    version: 4.12.4
+    specifier: ^4.24.7
+    version: 4.24.7
   react:
     specifier: ^18.2.0
     version: 18.2.0
@@ -3735,7 +3735,7 @@ packages:
   /@types/node-fetch@2.6.7:
     resolution: {integrity: sha512-lX17GZVpJ/fuCjguZ5b3TjEbSENxmEk1B2z02yoXSK9WMEWRivhdSY73wWMn6bpcCDAOh6qAdktpKHIlkDk2lg==}
     dependencies:
-      '@types/node': 18.18.6
+      '@types/node': 20.8.7
       form-data: 4.0.0
     dev: false
 
@@ -3747,7 +3747,6 @@ packages:
     resolution: {integrity: sha512-21TKHHh3eUHIi2MloeptJWALuCu5H7HQTdTrWIFReA8ad+aggoX+lRes3ex7/FtpC+sVUpFMQ+QTfYr74mruiQ==}
     dependencies:
       undici-types: 5.25.3
-    dev: true
 
   /@types/parse-json@4.0.1:
     resolution: {integrity: sha512-3YmXzzPAdOTVljVMkTMBdBEvlOLg2cDQaDhnnhT3nT9uDbnJzjWhKlzb+desT12Y7tGqaN6d+AbozcKzyL36Ng==}
@@ -8337,8 +8336,8 @@ packages:
       is-wsl: 2.2.0
     dev: true
 
-  /openai@4.12.4:
-    resolution: {integrity: sha512-oPNVJkpgxDUKF6WGGdHEZh5m/kjmYxS2Y1q7YVFCkvKUGthb8OGYRGCFBRPq5CQJezifzABTZRlVYnXLd6L4vQ==}
+  /openai@4.24.7:
+    resolution: {integrity: sha512-JUesECWPtsDHO0TlZGb6q73hnAmXUdzj9NrwgZeL4lqlRt/kR1sWrXoy8LocxN/6uOtitywvcJqe0O1PLkG45g==}
     hasBin: true
     dependencies:
       '@types/node': 18.18.6
@@ -9870,7 +9869,6 @@ packages:
 
   /undici-types@5.25.3:
     resolution: {integrity: sha512-Ga1jfYwRn7+cP9v8auvEXN1rX3sWqlayd4HP7OKk4mZWylEmu3KzXDUGrQUN6Ol7qo1gPvB2e5gX6udnyEPgdA==}
-    dev: true
 
   /undici@5.20.0:
     resolution: {integrity: sha512-J3j60dYzuo6Eevbawwp1sdg16k5Tf768bxYK4TUJRH7cBM4kFCbf3mOnM/0E3vQYXvpxITbbWmBafaDbxLDz3g==}
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -38,12 +38,16 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
           icon={<TbSend />}
         />
         {isTtsSupported() && (
-          <Tooltip label={settings.announceMessages ? "TTS Enabled" : "TTS Disabled"}>
+          <Tooltip
+            label={settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"}
+          >
             <IconButton
               type="button"
               size="lg"
               variant="solid"
-              aria-label={settings.announceMessages ? "TTS Enabled" : "TTS Disabled"}
+              aria-label={
+                settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"
+              }
               icon={settings.announceMessages ? <AiFillSound /> : <AiOutlineSound />}
               onClick={() =>
                 setSettings({ ...settings, announceMessages: !settings.announceMessages })
@@ -82,7 +86,9 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
         Ask {settings.model.prettyModel}
       </Button>
       {isTtsSupported() && (
-        <Tooltip label={settings.announceMessages ? "TTS Enabled" : "TTS Disabled"}>
+        <Tooltip
+          label={settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"}
+        >
           <Button
             type="button"
             size="sm"
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -6,6 +6,7 @@ import {
   MenuButton,
   MenuList,
   MenuItem,
+  Tooltip,
 } from "@chakra-ui/react";
 import { TbChevronUp, TbSend } from "react-icons/tb";
 
@@ -13,6 +14,7 @@ import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
 import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
+import { AiFillSound, AiOutlineSound } from "react-icons/ai";
 
 type PromptSendButtonProps = {
   isLoading: boolean;
@@ -64,6 +66,15 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
       <Button type="submit" size="sm" isLoading={isLoading} loadingText="Sending">
         Ask {settings.model.prettyModel}
       </Button>
+      <Tooltip label={settings.announceMessages ? "TTS Enabled" : "TTS Disabled"}>
+        <Button
+          type="button"
+          size="sm"
+          onClick={() => setSettings({ ...settings, announceMessages: !settings.announceMessages })}
+        >
+          {settings.announceMessages ? <AiFillSound /> : <AiOutlineSound />}
+        </Button>
+      </Tooltip>
       <Menu placement="top" strategy="fixed">
         <MenuButton
           as={IconButton}
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -15,6 +15,7 @@ import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
 import { AiFillSound, AiOutlineSound } from "react-icons/ai";
+import { isTtsSupported } from "../../lib/ai";
 
 type PromptSendButtonProps = {
   isLoading: boolean;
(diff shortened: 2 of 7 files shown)


src/hooks/use-chat-openai.ts
```diff
@@ -63,8 +80,27 @@ function useChatOpenAI() {
...
+            const newWords = currentText.split(" ").slice(lastTTSIndex);
+            const newWordsCount = currentText.split(" ").length;
+            lastTTSIndex = newWordsCount;
+
+            ttsWordsBuffer.push(...newWords);
+
+            if (
+              isTtsSupported() &&
+              getSettings().announceMessages &&
+              ttsWordsBuffer.length >= TTS_BUFFER_THRESHOLD
```
tarasglek, line 95: I feel like the buffer should be sentence-based,eg split on '.' and ensure you have > some minimal number of words(3?)..It feels like adding indiv words(eg 2-3 words that fit into 50 chars) to queue would make the experience terrible for slow outputs

Amnish04: I think there's a confusion. The buffer threshold of 50 is not chars, but words.

But I agree with you that it should be sentence-based because most times the buffer ends in the middle of a sentence (at 50 words interval), and it takes time (about a second) for the next audio chunk to start playing. Doing it based on sentences would only apply a pause at the end of sentence.

However, I'll need to think of a way to add back correct punctuations at the end of each sentence for accurate expressions if we are splitting on those.

tarasglek:
I tried this, it works! Thank you!

1) Detection shouldn't hardcode official openai..Just use the models endpoint and look for tts* models
image

2)I think we need to have submenus in message options. Eg I would like TTS to be exposed as "Speak" with a submenu of voices. Would also like to have autospeak and menu items.

Basically there are 2 separate cases: a) hands-free chat using mic and tts to respond and b) using chatcraft to speak messages for producing downloadable audio

tarasglek:
I think we should move 'retry with' to a submenu, and also add "speak with" submenu. Thus this depends on https://github.com/tarasglek/chatcraft.org/issues/359 to proceed

humphd:
In our meeting today, we agreed to try and extract a version of this PR that we can land for 1.1 (next Friday) and iterate in future Issues/PRs as we understand things better.

pnpm-lock.yaml
```diff
@@ -3735,7 +3735,7 @@ packages:
   /@types/node-fetch@2.6.7:
     resolution: {integrity: sha512-lX17GZVpJ/fuCjguZ5b3TjEbSENxmEk1B2z02yoXSK9WMEWRivhdSY73wWMn6bpcCDAOh6qAdktpKHIlkDk2lg==}
     dependencies:
-      '@types/node': 18.18.6
+      '@types/node': 20.8.7
```
rjwignar, line 3738: Is a major version jump like this likely to break anything? So far this branch behaves like the current main branch (i.e. nothing seems to be broken), so I think we're safe.

rjwignar (commented):
I've played around with this and I'm having lots of fun with it.
I like how both TTS responses and TTS input are both supported.

I can't offer suggestions on the code but the naming conventions used for methods and variables looks fine.

I did notice a couple things on desktop:
1. The `speech input` button/microphone icon doesn't show a tooltip when hovered. Is this something that should be added?
2. I also noticed that on desktop, using speech input requires users to click and hold the microphone button, releasing the button when they're finished their request. Would it be better to have the user click the button once to start recording, then to click the button again when they're done recording?

I also wonder how the speech input button would behave on a smartphone. Would it behave the same way (user must press + hold the microphone button while talking)? Is there a way we can test this build on our smartphones?

mingming-ma:
@rjwignar I test the mobile before using the safari on my iphone, I think any browser should be ok, just navigate to the deployed [url](https://amnish04-tts-support-354.console-overthinker-dev.pages.dev), I find sometimes the url will show by bot, but if not you can find that info in the Details in the `All checks`

tarasglek:
> 2. I also noticed that on desktop, using speech input requires users to click and hold the microphone button, releasing the button when they're finished their request. Would it be better to have the user click the button once to start recording, then to click the button again when they're done recording?

> 
> I also wonder how the speech input button would behave on a smartphone. Would it behave the same way (user must press + hold the microphone button while talking)? Is there a way we can test this build on our smartphones?

Same on mobile. It's glitchy on mobile. I would prefer to change behavior to click once to record and once to stop as you suggest as that will likely be less glitchy.

Amnish04:
> I've played around with this and I'm having lots of fun with it. I like how both TTS responses and TTS input are both supported.
> 
> I can't offer suggestions on the code but the naming conventions used for methods and variables looks fine.
> 
> I did notice a couple things on desktop:
> 
>     1. The `speech input` button/microphone icon doesn't show a tooltip when hovered. Is this something that should be added?
> 
>     2. I also noticed that on desktop, using speech input requires users to click and hold the microphone button, releasing the button when they're finished their request. Would it be better to have the user click the button once to start recording, then to click the button again when they're done recording?
> 
> 
> I also wonder how the speech input button would behave on a smartphone. Would it behave the same way (user must press + hold the microphone button while talking)? Is there a way we can test this build on our smartphones?

@rjwignar Thanks for the review! 
I agree with you on that there should be a `tooltip` for **microphone** button as currently it does not have any descriptive text.
And you're right, the microphone part is almost unusable on mobile, but this pull request only focuses on **text to speech** support (#354).

I saw that @tarasglek already mentioned it before in this issue #336 that focuses on better mobile interface.
We can probably have a separate standalone issue for fixing the **mic behaviour**.

Amnish04:
> @rjwignar I have tested the mobile version before using the safari on my iphone, I think any browser should be ok, just navigate to the deployed [url](https://amnish04-tts-support-354.console-overthinker-dev.pages.dev), I find sometimes the url will show by bot, but if not you can find that info in the Details in the `All checks`
> 
> I just tried this on my phone and I didn't hear any sound, I tap the TTS enabled but seems not works.

@mingming-ma Thanks for reviewing!

I also tested using your link on my phone and it didn't work for me either.

But I also tested using my codespace on **phone** and its **working** with that.
https://symmetrical-space-doodle-p6wqvr79qrj26gv9-5173.app.github.dev/c/MWt4JUNQcG-d3jyhHzGtS

Not sure why it didn't work for the other one 🤔

humphd:
> > 2. I also noticed that on desktop, using speech input requires users to click and hold the microphone button, releasing the button when they're finished their request. Would it be better to have the user click the button once to start recording, then to click the button again when they're done recording?
> 
> > I also wonder how the speech input button would behave on a smartphone. Would it behave the same way (user must press + hold the microphone button while talking)? Is there a way we can test this build on our smartphones?
> 
> Same on mobile. It's glitchy on mobile. I would prefer to change behavior to click once to record and once to stop as you suggest as that will likely be less glitchy.

Let's not deal too much with general audio UI in this issue if possible, and move that to an external issue.  I'll file an issue on this change.

rjwignar (approved):

rjwignar:
> Let's not deal too much with general audio UI in this issue if possible, and move that to an external issue. I'll file an issue on this change.
> 
> #368

In that case, the current changes look good to me. I'm very satisfied with the TTS button and functionality.

> In our meeting today [Wednesday], we agreed to try and extract a version of this PR that we can land for 1.1 (next Friday) and iterate in future Issues/PRs as we understand things better.

Does anything in this PR have to be changed before we can merge it?

Amnish04:
Just modified my buffering algorithm in the way I've been thinking about.

The following **pseudocode** is a summary of what's happening.
image

This is following the sentence based approach that @tarasglek [suggested](https://github.com/tarasglek/chatcraft.org/pull/357#discussion_r1462622977).

I tried this, and it's much faster than previous way in many cases.

Please let me know if this works!

Amnish04:
> I tried this, it works! Thank you!
> 
>     1. Detection shouldn't hardcode official openai..Just use the models endpoint and look for tts* models
> 
> 
> image
> 
> 2)I think we need to have submenus in message options. Eg I would like TTS to be exposed as "Speak" with a submenu of voices. Would also like to have autospeak and menu items.
> 
> Basically there are 2 separate cases: a) hands-free chat using mic and tts to respond and b) using chatcraft to speak messages for producing downloadable audio

Also @tarasglek, I couldn't understand your first point.

Is it regarding this function?
image

Or the way I have hardcoded `tts-1` here

image

humphd:
@Amnish04 a few thoughts:

1. We have an nlp tokenizer in our tree already, see https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/summarize.ts.  You can use it to get your sentences, words, etc. See docs at https://github.com/spencermountain/compromise.
2. @tarasglek means, "Don't use Is Official OpenAI" as your means for determining if TTS is supported, since other providers can/will also provide the same or similar models.  It would be better to see if the list of models includes `tts-` or something.

@rjwignar thanks for your review and push to get this landed. I'm loving seeing this evolve so quickly.  Great work.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -34,6 +37,20 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
           isLoading={isLoading}
           icon={<TbSend />}
         />
+        {isTtsSupported() && (
+          <Tooltip label={settings.announceMessages ? "TTS Enabled" : "TTS Disabled"}>
```
humphd, line 41: "TTS" might not be something users understand.  In user-facing UI, maybe "Text-to-Speech" might be more clear.

src/hooks/use-chat-openai.ts
```diff
@@ -63,8 +82,35 @@ function useChatOpenAI() {
         onResume() {
           setPaused(false);
         },
-        onData({ currentText }) {
+        async onData({ currentText }) {
```
humphd, line 85: You're not actually `await`'ing here.  I also wondered about adding `Promise<string>` vs. `string` to the audio queue.  Can we `await` the values before we add the strings?

humphd: Reading the code below for `textToSpeech()`, I would `await` all these calls, and have your queue your queue hold `string` instead.

src/hooks/use-audio-player.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!isPlaying && queue.length > 0) {
+      playAudio(queue[0]);
+    }
+  }, [queue, isPlaying]);
+
+  const playAudio = async (audioClipUri: Promise<string>) => {
+    setIsPlaying(true);
+    const audio = new Audio(await audioClipUri);
+    audio.onended = () => {
+      setQueue((oldQueue) => oldQueue.slice(1));
```
humphd, line 17: You should call `URL.revokeObjectURL()` here so you don't leak these MP3 blobs after you are done playing each one.

humphd (changes requested):
This is looking pretty close, a few things I noticed reading the code.

src/hooks/use-chat-openai.ts
```diff
@@ -63,8 +80,27 @@ function useChatOpenAI() {
...
           if (!pausedRef.current) {
+            // Hook tts code here
+            const newWords = currentText.split(" ").slice(lastTTSIndex);
+            const newWordsCount = currentText.split(" ").length;
+            lastTTSIndex = newWordsCount;
```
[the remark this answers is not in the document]
Amnish04: Actually, I replaced this entire logic. So these lines are already gone.

src/hooks/use-chat-openai.ts
Amnish04: Actually, I was awaiting it outside initially. That's why I have that `async` keyword there (forgot to remove in cleanup).

The reason why I don't `await` it before pushing is that there is no guarantee when `openai` api finishes processing and resolves a specific url.
image

**Example:**
Let's imagine we have 2 strings that need to be processed - `A` and `B`.

 1. We call the `textToSpeech` function for `A`, and the thread starts handling the next available operation until this is processing.
 2. While processing for `A` was happening, we were receiving more data in `onData` and `textToSpeech` got called for `B`.
 3. Unfortunately, the promise returned by `textToSpeech` got resolved first for `B`, and it gets added to queue.
 4. Ultimately, the audio clip for `A` also gets added.

This means our queue is like this now.
[audio for B, audio for A]
which beats the purpose for having a queue in the first place as the order is incorrect.

Just to confirm this, I tried the approach you mentioned again but got really **weird behaviour**.

src/hooks/use-audio-player.ts
Amnish04: I never knew that. Thanks for sharing!

pnpm-lock.yaml
Amnish04: I don't see any typescript warning as such, so should be good?

Also, I don't remember upgrading this manually. But I see that its a peer dependency for `openai`, so maybe it got automatically upgraded when I upgraded `openai`.
image

Amnish04:
@humphd Just pushed some changes as requested and left comments on some.

humphd (approved):

humphd:
Make sure all the follow-ups we need to do are filed before you land.

pnpm-lock.yaml
rjwignar: >We should probably update to node 20 from 18 regardless, and could do it in a separate PR. It might affect the CloudFlare build and Workers functions, so it needs a bit of care.

@humphd if `@types/node` is a peer dependency for the `openai` version required for TTS support, do we still need a separate PR for upgrading `@types/node`?
Also, is there a way we can break test the CloudFlare build and worker functions on this PR branch?

humphd: If it built OK in this PR preview with CloudFlare, then it should be fine.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -34,6 +37,24 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
...
+            label={settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"}
+          >
+            <IconButton
+              type="button"
+              size="lg"
+              variant="solid"
+              aria-label={
+                settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"
+              }
+              icon={settings.announceMessages ? <AiFillSound /> : <AiOutlineSound />}
```
tarasglek, line 51: I find it nearly impossible to tell apart the 2 icons. I suggest one of the chatcraft suggestions

1. **Volume Up / Volume Off**
   - On: `MdVolumeUp`
   - Off: `MdVolumeOff`

2. **Speaker High / Speaker Mute**
   - On: `FaVolumeUp`
   - Off: `FaVolumeMute`

3. **Microphone / Microphone Slash**
   - On: `FaMicrophone`
   - Off: `FaMicrophoneSlash`

4. **Headphones / Headphones Slash**
   - On: `FaHeadphones`
   - Off: `FaHeadphonesAlt`

5. **Play Circle / Pause Circle**
   - On: `FaPlayCircle`
   - Off: `FaPauseCircle`

6. **Soundwave / No Soundwave**
   - On: `BiVolumeFull`
   - Off: `BiVolumeMute`

7. **Music Note / No Music**
   - On: `IoMusicalNote`
   - Off: `IoMusicalNotes

tarasglek:
Also tts seems to on always for me. Icons change, but the damn thing wont shut up

tarasglek:
Need to have audio abort when tts is in progress and one flips icon off

humphd:
@Amnish04 would you be willing to file those for follow-up so we don't lose?

PR #358, opened 2024-01-23
Fix: Better UX for long messages

### LINKED ISSUE #339: Better UX for long messages
@tarasglek shared a few chats last night that include really long blocks of text as part of the prompt's context (in both cases, he's asking ChatCraft to create "chain of summaries," which has the LLM iterate on its own work to make it better over time):

- https://chatcraft.org/c/tarasglek/mcHmCrkvVpLKwK7FfjSC2
- https://chatcraft.org/c/tarasglek/oHZpX9u7nHvyK11CGVGHU

Getting to the replies underneath is painful, since you have to scroll a hundred times.  Apps like Discord already provide a way to deal with this in the UI:

image

image

When a message or code block is over some threshold in size, we should switch to showing it like a file attachment.  Having some way to open it as text if you need to would be useful too.

mingming-ma:
Fix #339 

Similar to the system prompt behaviour, and toggles independently in each card. 
Only display the button when `total message text length > 5000` (and collapse)

[Result](https://mingming-longmessages.console-overthinker-dev.pages.dev/c/chatcraft_dev/FOhhuWdlXQD_mP4AmqZtN)

### DIFF
--- a/src/components/Message/HumanMessage.tsx
+++ b/src/components/Message/HumanMessage.tsx
@@ -33,9 +33,8 @@ function HumanMessage(props: HumanMessageProps) {
   );
 
   const { text } = message;
-  const textLines = text.split("\n");
-  const veryLongMessage = textLines.length > 100 || textLines[0].length > 1000;
-  const summaryText = textLines[0].slice(0, 250).trim() + "...";
+  const veryLongMessage = text.length > 5000;
+  const summaryText = text.split("\n")[0].slice(0, 250).trim() + "...";
 
   return (
     <MessageBase
--- a/src/components/Message/HumanMessage.tsx
+++ b/src/components/Message/HumanMessage.tsx
@@ -1,12 +1,22 @@
 import { memo } from "react";
-import { Avatar } from "@chakra-ui/react";
+import { Avatar, Button, Flex, useDisclosure } from "@chakra-ui/react";
 
 import MessageBase, { type MessageBaseProps } from "./MessageBase";
 
 type HumanMessageProps = Omit<MessageBaseProps, "avatar"> & { avatarUrl?: string; name?: string };
 
 function HumanMessage(props: HumanMessageProps) {
-  const { avatarUrl, name, ...rest } = props;
+  const { avatarUrl, message, name, ...rest } = props;
+  const { isOpen, onToggle } = useDisclosure();
+
+  // If we're showing the whole prompt, don't bother with the "More..." button
+  const footer = (
+    <Flex w="100%" justify="space-between" align="center">
+      <Button size="sm" variant="ghost" onClick={() => onToggle()}>
+        {isOpen ? "Show Less" : "Show More..."}
+      </Button>
+    </Flex>
+  );
 
   // If we have a user's name and GitHub avatar, use that.
   const avatar = avatarUrl ? (
@@ -22,7 +32,21 @@ function HumanMessage(props: HumanMessageProps) {
     <Avatar size="sm" bg="gray.600" _dark={{ bg: "gray.500" }} />
   );
 
-  return <MessageBase {...rest} avatar={avatar} heading={name} />;
+  const { text } = message;
+  const textLines = text.split("\n");
+  const veryLongMessage = textLines.length > 100 || textLines[0].length > 1000;
+  const summaryText = textLines[0].slice(0, 250).trim() + "...";
+
+  return (
+    <MessageBase
+      {...rest}
+      message={message}
+      avatar={avatar}
+      heading={name}
+      summaryText={!isOpen && veryLongMessage ? summaryText : undefined}
+      footer={veryLongMessage ? footer : undefined}
+    />
+  );
 }
 
 export default memo(HumanMessage);
--- a/src/components/Message/HumanMessage.tsx
+++ b/src/components/Message/HumanMessage.tsx
@@ -1,29 +1,12 @@
-import { memo, startTransition } from "react";
-import { Avatar, Button, Flex, useDisclosure } from "@chakra-ui/react";
+import { memo } from "react";
+import { Avatar } from "@chakra-ui/react";
 
 import MessageBase, { type MessageBaseProps } from "./MessageBase";
 
 type HumanMessageProps = Omit<MessageBaseProps, "avatar"> & { avatarUrl?: string; name?: string };
 
 function HumanMessage(props: HumanMessageProps) {
-  const { avatarUrl, message, name, ...rest } = props;
-  const { isOpen, onToggle: originalOnToggle } = useDisclosure();
-
-  // Wrap the onToggle function with startTransition, state update should be deferred due to long message
-  // https://reactjs.org/docs/error-decoder.html?invariant=426
-  const onToggle = () => {
-    startTransition(() => {
-      originalOnToggle();
-    });
-  };
-
-  const footer = (
-    <Flex w="100%" justify="space-between" align="center">
-      <Button size="sm" variant="ghost" onClick={() => onToggle()}>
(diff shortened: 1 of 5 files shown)


src/components/Message/HumanMessage.tsx
```diff
@@ -22,7 +39,20 @@ function HumanMessage(props: HumanMessageProps) {
     <Avatar size="sm" bg="gray.600" _dark={{ bg: "gray.500" }} />
   );
 
-  return <MessageBase {...rest} avatar={avatar} heading={name} />;
+  const { text } = message;
```
WangGithub0, line 42: can you directly use message instead of extracting text from it?

humphd (commented):
After I click "Show More...", I have to scroll all the way to the bottom to find "Show Less"

Screenshot 2024-01-25 at 9 38 39 AM

This makes it hard to toggle full vs. collapsed view.  I wonder if we should show the "Show Less" at the top and bottom?

mingming-ma:
@humphd That's a great feedback, we should have a way to quick collapse here!

Meanwhile, adding the button "Show Less" at the top is a little bit of hard because the `CardHeader` is defined in the `MessageBase` component, while the `Show More/Less` button is defined in the footer which can be passed from `HumanMessage` (and other higher component) to `MessageBase`

Currently we have these buttons in the header, 
image
I suggest we add a new icon button next to the `copy`, to plays as jump to the end of current card (then user can see the `Show less` button and toggle), it also gives user a choice to quick go to next message, and it is implemented quite [easily](https://stackoverflow.com/a/51828976):

```
const ScrollDemo = () => {
   const myRef = useRef(null)

   const executeScroll = () => myRef.current.scrollIntoView()    
   // run this function from an event handler or an effect to execute scroll 

   return (
      <> 
         <div ref={myRef}>Element to scroll to</div> 
         <button onClick={executeScroll}> Click to scroll </button> 
      </>
   )
}
```

What do you think?

humphd:
Could you render the "Show Less" button in the message itself, above the content, when the text is long and expanded?

I understand it might be too complicated the way the components are currently written.

Another thought, should we wrap long message contents in an [Accordion](https://chakra-ui.com/docs/components/accordion/usage)?

mingming-ma:
@humphd The ref can be added at the bottom of the card body

image

I have tried locally and works good, let me commit for test.

> Another thought, should we wrap long message contents in an [Accordion](https://chakra-ui.com/docs/components/accordion/usage)?

I'll try this also and let's compare.

humphd:
At some point we might also redo the structure of how these Messages work.  I never properly designed it, just changed it as we needed to do new things.  I don't think spending time on that for no reason is useful; but if we get to the point that changes are overly complicated, we could.

mingming-ma:
Have a look at my way [Result](https://mingming-longmessages.console-overthinker-dev.pages.dev/c/chatcraft_dev/FOhhuWdlXQD_mP4AmqZtN)

I'll also try implement the suggestions:
> the render the "Show Less" button in the message
> wrap long message contents in an [Accordion](https://chakra-ui.com/docs/components/accordion/usage)

mingming-ma:
@humphd I got it! I realized I can update the `MessageBase` directly instead of passing from `HumanMessage`
image

I'm going to fix the alignment and then commit.

tarasglek:
I find the icon to go to next message to be very confusing..as it shows us going to the right.
 
 for message expansion, https://v0.chakra-ui.com/accordion icon makes more sense.
 
 Maybe we should just use that for system prompt too?

tarasglek:
I would consider changing our messages to use these accordeons or something similar if possible
image

mingming-ma:
Yeah, the right arrow icon is confusing and accordion makes more sense. So now we don't need the 'Show More.../Less' button I think. And wrap long messages and system prompt in accordion. Let me implement in this way.

humphd (changes requested):
I'm not a fan of this approach:

Screenshot 2024-01-26 at 1 31 18 PM

First, the accordion header doesn't render Markdown, so we get raw text.

Second, the extra borders at the top/bottom of the accordion header are no good.

Third, the margin between the text and the down-arrow chevron needs more space.

Fourth, it's unclear what you're seeing. For example, the ```html... header doesn't indicate that there is more text available.

I think the idea of using accordion-like functionality is OK, but the ChakraUI Accordion component isn't a good fit here.

mingming-ma:
@humphd I think it will be tricky to render part of content in the header. And if we only use `Show more...` will need to click each time to see the content.
So the approach now is we need to find some toggle like behaviour component and better shows top of the content with rendered result. Let me try to find some examples.

mingming-ma:
@humphd I just thought maybe we can add the `AccordionIcon`  in the header next to `copy` to let user expand, and it will like before with rendered result, sounds good?

humphd:
It's worth a try.  We could also include the "Show more..." and "Show Less" buttons still, and they would trigger the same thing.  Best of both worlds.

mingming-ma:
@humphd I found in the mobile version the icons will not show, so I think there is no need to add the icon for now as the buttons works fine. Maybe we can leave it for other cases.

WangGithub0:
If we still retain "Show Less" at the bottom, would it be convenient for users who have already dragged it down below?

humphd (changes requested):
This is getting really close.  Here's my feedback on the current design.

When a message is long, and you only show the first N-lines, it's unclear that there is more content that isn't being shown:

Screenshot 2024-01-27 at 8 40 42 AM

Take a look at the original issue I filed and the way Discord handles this case:

image

Here you can see that the bottom of the text element is blurred and intentionally cut-off, so as to indicate that there is more.  We don't need to do exactly what they do, but we need something to indicate that there is more text you aren't seeing.  I'd suggest we somehow change the bottom border to be jagged or blurred or something.

Another change I would make is to move the "Show More..." button to the footer when not expanded.  I think it's more intuitive to the eye to follow the content down vertically until you hit the "Show More..." button, and then it registers that you can do something.

When the user expands the content, I would put the "Show Less" button at the top and bottom (like you do now).  In other words: only put the "Show More..." in the footer, but "Show Less" should go above and below.

mingming-ma:
@WangGithub0 Absolutely! We should also have in the bottom. Thanks for the suggestion.

@humphd Couldn't agree more! The jagged/blurred idea is pretty cool and `Show more` button being bottom makes more sense. Let me implement that!

humphd (approved):
This is really fantastic.  The blur leading your eye down to a "Show More..." button is perfect.

I've left one comment about context length, but I think this is ready to land and we can iterate later as we use it.

Amazing work @mingming-ma!

src/components/Message/MessageBase.tsx
```diff
@@ -381,9 +402,21 @@ function MessageBase({
...
-                  {summaryText || text}
-                </Markdown>
+                <>
+                  <Markdown
+                    previewCode={!hidePreviews && !displaySummaryText}
+                    isLoading={isLoading}
+                    onPrompt={onPrompt}
+                    className={displaySummaryText ? "message-text message-text-blur" : undefined}
+                  >
+                    {displaySummaryText ? summaryText || text.slice(0, 250).trim() : text}
```
humphd, line 412: I wonder if 250 is too short.  It means that we cut it off really fast for content returned by `/import`, and it's hard to know if you care about seeing what's inside, since you don't have enough context:

Screenshot 2024-01-30 at 1 57 04 PM

mingming-ma: True, let's try 500

mingming-ma:
@humphd 
image
Let's try 500 workable and let me increase it later if needed. Thanks!

humphd (approved):
I think 500 works well, nice!

PR #369, opened 2024-01-26
Improve Mic UX

### LINKED ISSUE #368: Rework Audio Recording UI
Spin-off from https://github.com/tarasglek/chatcraft.org/pull/357#issuecomment-1911135568

Our current audio recording UI was designed to mimic "press and hold" audio UX in various apps.  However, getting this to work flawlessly on mobile web across browsers has been very tricky.

We want to move to a "press to start, press to stop" audio UX.  I'm less familiar with audio recording UX than @tarasglek, so I'll let him comment on what he thinks we should do here (e.g., how to handle cancelling vs. stopping).

Amnish04:
I spent a few hours last night fixing the **Mic Drag UX** on mobile based on discussions in #357, and was able make it better to some point but its quite tedious to get around all the edge cases.
This branch has all the work for that.
https://github.com/tarasglek/chatcraft.org/tree/amnish04/mic-mobile

@tarasglek has been suggesting a "press to start, press to stop" audio UX instead, which is much easier to implement for mobile browsers.

Based on the suggestions, I have made the following changes to fix #368 

1. The **Mic Button** on **mobile** now uses "press to start, press to stop".
2. I did not change the drag behaviour for **Desktop** as it seems to work seamlessly on computers. I also noticed there was a lot of work put into it, so wanted to preserve it if possible.
3. Previously, the mic icon could be dragged to both right and left. Dragging and leaving to **left** side **cancelled** the recording, but I couldn't find a reason to allow dragging to **right**, which is why I have disabled that behaviour.

I am not sure if we want it to behave in this way, but we can make more changes as needed.

### DIFF
--- a/src/components/PromptForm/MicIcon.tsx
+++ b/src/components/PromptForm/MicIcon.tsx
@@ -103,7 +103,29 @@ export default function MicIcon({
     onCancel();
   };
 
-  return (
+  const handleMobileMicToggle = () => {
+    if (isRecording) {
+      onRecordingStop();
+    } else {
+      onRecordingStart();
+    }
+  };
+
+  return isMobile ? (
+    <IconButton
+      isRound
+      isDisabled={isDisabled}
+      colorScheme={colorScheme}
+      icon={<TbMicrophone />}
+      variant={isRecording ? "solid" : isMobile ? "outline" : "ghost"}
+      aria-label="Record speech"
+      size={isMobile ? "lg" : "md"}
+      fontSize="18px"
+      ref={micIconRef}
+      onClick={handleMobileMicToggle}
+      onBlur={() => onRecordingCancel()}
+    />
+  ) : (
     <motion.div
       drag="x"
       dragConstraints={{ left: 0, right: 0 }}
@@ -114,6 +136,11 @@ export default function MicIcon({
           setColorScheme("red");
         } else {
           setColorScheme("blue");
+        }
+
+        // If dragging to the right, set x to the maximum allowed value
+        if (info.offset.x > 0) {
+          x.set(0);
         }
       }}
       onDragEnd={(_event, info) => {
--- a/src/components/PromptForm/MicIcon.tsx
+++ b/src/components/PromptForm/MicIcon.tsx
@@ -1,11 +1,10 @@
-import { useState, useRef } from "react";
 import { IconButton } from "@chakra-ui/react";
+import { useRef, useState } from "react";
 import { TbMicrophone } from "react-icons/tb";
-import { motion, useMotionValue } from "framer-motion";
 
-import { SpeechRecognition } from "../../lib/speech-recognition";
 import { useAlert } from "../../hooks/use-alert";
 import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
+import { SpeechRecognition } from "../../lib/speech-recognition";
 
 type MicIconProps = {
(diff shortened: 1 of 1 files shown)


humphd:
I'm OK with trashing the desktop drag behaviour if it makes life easier or you want, since I wrote most of it :). I'll let @tarasglek give initial feedback on this, since he's our heaviest audio user.

kliu57 (commented):
@Amnish04 Tested mobile using Dev Tools and it works well. Was able to record my voice with press to start, press to stop.

I know that dragging the mic icon to the left cancels sending the recording to chatgpt, but is there a way to let the user know that it would do that? It doesn't seem too intuitive. Maybe if there was a cancel (X) icon that appeared to the left of it that people could drag it onto?

tarasglek:
Is there a good reason to support dragging? I think we should remove it

tarasglek:
Eg lets make desktop and mobile behavior same

Amnish04:
@tarasglek @kliu57 I guess you guys seem to prefer the click behaviour over drag. I am okay with both, so sure we can have it that way for both mobile and desktop.

humphd:
Drag complicates things.  It was a good experiment, but getting it perfect is a fool's errand.  Let's go with touch-only as @tarasglek and @kliu57 suggest.

Amnish04:
Just removed the drag behaviour altogether.
It should "touch to start, touch to stop" for all screens now!

Amnish04:
Another thing I noticed, I don't see us using [framer-motion](https://www.npmjs.com/package/framer-motion) lib anywhere else.

Should we remove it from our tree?

humphd:
I'm pretty sure framer-motion is required by ChakraUI, see https://chakra-ui.com/getting-started.

humphd (commented):
Some usability thoughts:

- On Desktop, hovering over the icon should probably help the user discover its purpose (maybe a `title="Click to record audio"` or something)
- When first clicked, you have to grant permission, but then after you do, you have to click it again to start recording, which seems odd.  Might not be fixable, by I tripped over it
- When recording, I would make the button solid and red, indicating that if you click it again, it will stop
- We might also want to add (Click again to stop recording) or something to the left?

tarasglek:
I think we should animate the mic while it's recording..it should be red and changing..Something like https://www.shutterstock.com/video/clip-1071785533-microphone-signal-animated-icon-animation-4k-green ?

tarasglek:
I also think I would merge this as is and do the further UX improvements in a followup. This is a huge improvement.

tarasglek (approved):

humphd:
@Amnish04 if you want to land the current stuff and follow-up, please file new issues first so we don't lose track.

Amnish04:
> Some usability thoughts:
> 
>     * On Desktop, hovering over the icon should probably help the user discover its purpose (maybe a `title="Click to record audio"` or something)
> 
>     * When first clicked, you have to grant permission, but then after you do, you have to click it again to start recording, which seems odd.  Might not be fixable, by I tripped over it
> 
>     * When recording, I would make the button solid and red, indicating that if you click it again, it will stop
> 
>     * We might also want to add (Click again to stop recording) or something to the left?

I agree, will file a follow up now.

Amnish04:
> @Amnish04 if you want to land the current stuff and follow-up, please file new issues first so we don't lose track.

@humphd Just filed a follow up issue #413 for this

Issue #360, opened 2024-01-24
Fetch subtitles for YouTube videos with /import


humphd:
Spin-off from https://github.com/tarasglek/chatcraft.org/issues/325#issuecomment-1907573483.

Let's update [our CloudFlare `proxy` function](https://github.com/tarasglek/chatcraft.org/blob/34fbb43a595a5adcce9e85e84a3b5dea4d898222/functions/api/proxy.ts#L7) to support fetching subtitles from YouTube for a YouTube link using https://www.npmjs.com/package/youtube-captions-scraper.

If I give a link to a YouTube video, it should return subtitle text to us.

NOTE: YouTube URLs are complicated https://gist.github.com/rodrigoborgesdeoliveira/987683cfbfcc8d800192da1e73adc486

PR #370, opened 2024-01-26
Add transformers, include GitHub and YouTube transformers

humphd:
I wanted to make it cleaner to add transformations for the `/proxy` endpoint used when we use the `/import` slash command.

This adds two types of transformers:

- `UrlTransformer`: used to take a URL and rewrite it before we do the `fetch()`
- `ResponseTransformer`: used to update a fetched `Response` before we return it

I've converted the various GitHub URL rewrites to use `UrlTransformer`, and also added one for working with PRs, like I showed in class the other day.

I included the stuff to transform `Response`s before we return them, but I'm not using it yet.  It would be useful it you wanted to turn something with a particular `Content-Type` into some other format, or parse the body and generate some different content, etc.

I've included tests to show that these all work.


humphd:
Not sure why that test fails in CI, it passes locally.  Maybe the node version?  Seems odd.

humphd:
Turns out it shouldn't have been passing locally!  Fixed.

humphd:
[Here is a sample chat that uses it](https://29d177d4.console-overthinker-dev.pages.dev/c/chatcraft_dev/aD20M8FXI1LxoDM3yx4ut), showing this PR being `/import`'ed, and the patch being returned.

humphd:
Fixes #360.

I've updated the transformer API so that it's possible to:

- rewrite the URL
- rewrite the Response
- ignore the Response and use a different logic for fetching the data

In order to demonstrate a number of examples, I've done GitHub URL rewrites and YouTube captioning.  You can try using `/import` with GitHub files, Gists, or PRs.

You can try YouTube videos, too. Here's an [example of using `/import` with a YouTube video](https://f8e728ab.console-overthinker-dev.pages.dev/c/chatcraft_dev/4g7sHO0SnNce8J4-rrffu).

I ran into a few bugs doing this.  One in https://github.com/algolia/youtube-captions-scraper/pull/25 (can't use `axios` in CloudFlare workers), which I've hacked around by re-implementing the code in our tree temporarily.  The other relates to `&` in URLs being rewritten in our Markdown render to `&amp;`, which I'll file separately.

Another thing I'm not sure how to handle is the case where a video doesn't have any subtitles. Currently we'll return an error, but there is likely a better UX.

humphd:
There's also an extra 3 backticks at the end of the data, which I don't understand:

Screenshot 2024-01-28 at 11 33 38 AM

It might be getting added in the browser, not sure.

I also realize that in a follow-up, we could transform the GitHub stuff so it properly syntax highlights the files (e.g., return Markdown with fenced code blocks vs. text).

humphd:
Filed #373 on the & vs. &amp; issue.

tarasglek:
Note, rendering these in any fancy way will eat tokens for breakfast thus process slower and more expensively. I would even strip timestamps and llms do without those.

humphd:
> Note, rendering these in any fancy way will eat tokens for breakfast thus process slower and more expensively. I would even strip timestamps and llms do without those.

Good point.  I've altered it to return only text, and sentences are separated by newlines.

Example: https://b957613a.console-overthinker-dev.pages.dev/c/chatcraft_dev/RlW21J19uxWlP5CdXYEPt

mingming-ma:
@humphd I think the 3 backticks at the end comes from file `src/lib/commands/ImportCommand.ts`

```ts
    const text =
      `${command}\n\n` +
      // If it's already markdown, dump it into the message as is;
      // otherwise, wrap it in a code block with appropriate type
      (type === "markdown" ? `${content}` : `\`\`\`${type}\n${content}`) +
      `\n\`\`\``;
```

Possible fix:

```ts
      (type === "markdown" ? `${content}` : `\`\`\`${type}\n${content}\n\`\`\``);
```

humphd:
- Fixed the trailing backticks, thanks @mingming-ma 
- Added support for transforming GitHub responses to syntax highlight diffs and blobs (e.g., files)

To test this you can `/import` a PR URL or a GitHub file.

I'll probably stop extending this now, since I have examples of how to use everything, which others could copy.

tarasglek:
Is there a way to test this before landing on main? Can we call these rewriters(or something)? It's gonna make someone have a bad day if they confuse these with huggingface transformers.

tarasglek:
lgtm on the code other than name

tarasglek:
Got it to work.

```
import https://github.com/tarasglek/chatcraft.org/pull/370#issuecomment-1915383498
```
Does not work...

tarasglek:
i think for pull reqs it should not do the patch translation unless one points it at git diff view.. eg we might wanna mine pull comments

humphd:
Fixed so it won't do PR comments (I'll file a follow-up to add that), removed my `youtube-captions-scraper` in favour of upstream fix https://github.com/algolia/youtube-captions-scraper/pull/25, did some cleanup, rebased.

@tarasglek RE: name, what should I call these instead of "Transofmer"? "Rewriter", "Handler", "Processor", "Mutator" ??

tarasglek:
Markdown renderer is the true name for what it does. Simplifier or Rewriter is fine too

humphd:
Renamed transformers to rewriters.  MarkdownRenderer is true only if we operate on the response, which we don't always.

I think I'll land this and we can continue in subsequent PRs.

PR #371, opened 2024-01-27
Allow storage of multiple providers info

### LINKED ISSUE #345: Allow multiple providers to be used simultaniously
At the moment to switch providers we have to go to preferences and swap out provider + key. Would like to just have a set of providers, with keys and switch between them and their models as easily as we do within one provider.

Currently openrouter has some really unique models eg perplexity model that can query the internet:
image

Useless openai response:
image

However there are some downsides of accessing gpt through openrouter(robustness, speed, privacy, etc). Also in the future we might want to use dedicated openai within Azure. For all these reasons we would like to be able to list models across multiple providers and switch between then when doing retry-with...

### LINKED ISSUE #383: Store multiple providers
Part 1 of [Issue 345](https://github.com/tarasglek/chatcraft.org/issues/345)

- Save provider url and key in settings.providers whenever a new key is entered
- Do not yet read from settings.providers (use the old settings.apiUrl and settings.apiKey)
- Remove button to remove provider from settings.providers as well as removing settings.apiKey

kliu57:
Closes #383 

Save all newly inputted provider info to associative array settings.providers (stored in localStorage in providers)

Existing functionality is unchanged. We are only storing to settings.providers, not reading from settings.providers yet.

**Code changes:**

- `src/lib/ChatCraftProvider.ts` — new class which stores information for a provider such as provider name, url, key.
- `src/lib/settings.ts` — added the variable `providers` to Settings, which is an associative array of `ChatCraftProvider` objects
- `src/components/Message/AppMessage/Instructions.tsx` — added code so when new ChatCraft users enters a key and clicks the Save button, the provider info is stored to `settings.providers`
- `src/components/PreferencesModal.tsx` — added code to the event handlers of the API URL field, API Key field, Remove button, so that in addition to performing their usual behaviour, they also store/remove data from settings.providers

**Behaviour of the new code:**

1) New users to ChatCraft — the moment they input their api key, the provider information will be stored in settings.providers. You can test this by checking localStorage.

2) Existing users to ChatCraft — their current api key will not yet be stored in settings.providers. Will be addressed in release 1.2.

3) When a user changes their provider url in Settings — the moment they toggle to another API url, settings.providers is cleared, to match the current behaviour. This behaviour will change in release 1.2 (In 1.2 we will not clear the providers and instead keep them stored in settings.providers to be read from at a later date)

4) When a user enters a new/different provider key in Settings — the moment they enter a key, the key stored in settings.providers is stored or overwritten. There will never be more than one key stored for openai, for example.

5) When a user clicks the Remove button in Settings — if the current provider is stored in settings.providers it is removed from settings.providers.

image1

**How to check localStorage in Chrome DevTools:**

Right click + Inspect (or F12)
image2

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -36,6 +36,7 @@ import { download, isMac } from "../lib/utils";
 import db from "../lib/db";
 import { useModels } from "../hooks/use-models";
 import { ChatCraftModel } from "../lib/ChatCraftModel";
+import { ChatCraftProvider } from "../lib/ChatCraftProvider";
 import { OPENAI_API_URL, OPENROUTER_API_URL } from "../lib/settings";
 import { openRouterPkceRedirect, validateApiKey } from "../lib/ai";
 import { useAlert } from "../hooks/use-alert";
@@ -164,9 +165,14 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <FormLabel>API URL</FormLabel>
               <Select
                 value={settings.apiUrl}
-                onChange={(e) =>
-                  setSettings({ ...settings, apiUrl: e.target.value, apiKey: undefined })
-                }
+                onChange={(e) => {
+                  setSettings({
+                    ...settings,
+                    apiUrl: e.target.value,
+                    apiKey: undefined,
+                    providers: {},
+                  });
+                }}
               >
                 <option value={OPENAI_API_URL}>OpenAI ({OPENAI_API_URL})</option>
                 <option value={OPENROUTER_API_URL}>OpenRouter.ai ({OPENROUTER_API_URL})</option>
@@ -190,7 +196,23 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
                   <Button
                     size="xs"
                     colorScheme="red"
-                    onClick={() => setSettings({ ...settings, apiKey: undefined })}
+                    onClick={() => {
+                      const currentProvider = ChatCraftProvider.fromUrl(
+                        settings.apiUrl,
+                        settings.apiKey
+                      );
+                      const updatedProviders = { ...settings.providers };
+
+                      if (currentProvider.name in updatedProviders) {
+                        delete updatedProviders[currentProvider.name];
+                      }
+
+                      setSettings({
+                        ...settings,
+                        apiKey: undefined,
+                        providers: updatedProviders,
+                      });
+                    }}
                     isDisabled={!settings.apiKey}
                   >
                     Remove
@@ -200,7 +222,18 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <RevealablePasswordInput
                 type="password"
                 value={settings.apiKey || ""}
-                onChange={(e) => setSettings({ ...settings, apiKey: e.target.value })}
+                onChange={(e) => {
+                  const newProvider = ChatCraftProvider.fromUrl(settings.apiUrl, e.target.value);
+
+                  setSettings({
+                    ...settings,
+                    apiKey: e.target.value,
+                    providers: {
+                      ...settings.providers,
+                      [newProvider.name]: newProvider,
+                    },
+                  });
+                }}
               />
               {provider === "OpenRouter.ai" && !settings.apiKey && (
                 <Button mt="3" size="sm" onClick={openRouterPkceRedirect}>
--- a/src/lib/ChatCraftProvider.ts
+++ b/src/lib/ChatCraftProvider.ts
@@ -1,0 +1,45 @@
+import { nanoid } from "nanoid";
+
+type ProviderName = "OpenAI" | "OpenRouter.ai";
+
+export const OPENAI_API_URL = "https://api.openai.com/v1";
+export const OPENROUTER_API_URL = "https://openrouter.ai/api/v1";
+
+const urlToNameMap: { [key: string]: ProviderName } = {
+  [OPENAI_API_URL]: "OpenAI",
+  [OPENROUTER_API_URL]: "OpenRouter.ai",
+};
+
+export type SerializedChatCraftProvider = {
+  id: string;
+  name: ProviderName;
+  apiUrl: string;
+  apiKey?: string;
+};
+
+export class ChatCraftProvider {
+  id: string;
+  name: ProviderName;
+  apiUrl: string;
+  apiKey?: string;
+
+  constructor(url: string, key?: string) {
+    this.id = nanoid();
+    this.name = urlToNameMap[url];
+    this.apiUrl = url;
+    this.apiKey = key;
+  }
+
+  // Parses url into instance of ChatCraftProvider
+  static fromUrl(url: string, key?: string) {
+    return new ChatCraftProvider(url, key);
+  }
+
+  // Parse from serialized JSON
+  static fromJSON({ id, name, apiUrl, apiKey }: SerializedChatCraftProvider): ChatCraftProvider {
+    const provider = new ChatCraftProvider(apiUrl, apiKey);
+    provider.id = id;
+    provider.name = name;
+    return provider;
+  }
+}
--- a/src/components/Message/AppMessage/Instructions.tsx
+++ b/src/components/Message/AppMessage/Instructions.tsx
@@ -12,6 +12,7 @@ import {
 
 import MessageBase, { type MessageBaseProps } from "../MessageBase";
 import { ChatCraftAppMessage } from "../../../lib/ChatCraftMessage";
+import { ChatCraftProvider } from "../../../lib/ChatCraftProvider";
 import RevealablePasswordInput from "../../RevealablePasswordInput";
 import { useSettings } from "../../../hooks/use-settings";
 import { openRouterPkceRedirect, validateApiKey } from "../../../lib/ai";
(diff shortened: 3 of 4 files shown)


kliu57:
@humphd You asked for the UI in the settings to iterate over the list of providers; however right now my settings.providers is an associative array that only stores the providers the user has given (so it could only contain 1). We would want to show them 2 providers in the dropdown and not 1. Thoughts?

image

src/components/PreferencesModal.tsx
```diff
@@ -200,7 +200,20 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <RevealablePasswordInput
                 type="password"
                 value={settings.apiKey || ""}
-                onChange={(e) => setSettings({ ...settings, apiKey: e.target.value })}
+                onChange={(e) => {
+                  const newProvider =
```
humphd, line 204: I wouldn't do this logic in the UI code, move it to the `ChatCraftProvider` class via a `static` method named `fromUrl()`, which parses a URL into a provider instance:

```ts
const newProvider = ChatCraftProvider.fromUrl(settings.apiUrl);
...
[newProvider.name]: newProvider,
```

humphd:
> @humphd You asked for the UI in the settings to iterate over the list of providers; however right now my settings.providers is an associative array that only stores the providers the user has given (so it could only contain 1). We would want to show them 2 providers in the dropdown and not 1. Thoughts?
> 
> image

I think we have to separate the list of providers a user has configured in their settings from the list of providers we support.  In other words, we might support 3 different providers in the future, but a user might only have 1 or 2 configured.  So the UI can use the list of supported providers to show possible providers for a user to configure.

humphd:
I saw [Simon tweeting yesterday](https://twitter.com/simonw/status/1751276501809647919?t=Q-jSo3oqjJCmqVa-pXrlSQ) about this same thing (using OpenAI compatible providers), and people mention:

- [TogetherAI](https://www.together.ai/products#inference)
- [Anyscale](https://www.anyscale.com/endpoints)
- [Fireworks.ai](https://readme.fireworks.ai/docs/quickstart)

Might be interesting to support/try these with ChatCraft.

src/lib/ChatCraftProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+  setUrl(url: string) {
+    this.apiUrl = url;
+  }
+
+  setKey(key: string) {
+    this.apiKey = key;
+  }
+
+  // Parses url into instance of ChatCraftProvider
+  static fromUrl(url: string) {
```
humphd, line 28: What if you do `static fromUrl(url: string, apiKey: string) {` so that you require the `apiKey` at the same time, thus getting rid of the case where it's set to `""`.

src/lib/ChatCraftProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+  apiKey: string;
+
+  constructor(name: string, url: string, key: string) {
+    this.id = nanoid();
+    this.name = name;
+    this.apiUrl = url;
+    this.apiKey = key;
+  }
+
+  setUrl(url: string) {
```
humphd, line 19: I don't think `setUrl` or `setKey` are necessary, since you are exposing `apiUrl` and `apiKey` as public members.

src/lib/ChatCraftProvider.ts
```diff
@@ -0,0 +1,44 @@
+import { nanoid } from "nanoid";
+
+export const OPENAI_API_URL = "https://api.openai.com/v1";
+export const OPENROUTER_API_URL = "https://openrouter.ai/api/v1";
+
+export class ChatCraftProvider {
+  id: string;
+  name: string;
```
humphd, line 8: Let's narrow this type from `string` to the values we want:

```ts
type ProviderName = 'openai' | 'openrouter.ai';
...
name: ProviderName;
```

src/lib/ChatCraftProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+      case OPENAI_API_URL:
+        name = "OpenAI";
+        break;
+      case OPENROUTER_API_URL:
+        name = "OpenRouter.ai";
+        break;
+      default:
+        throw new Error("Invalid URL");
+    }
+
```
humphd, line 41: You should also do `fromJSON()` so we can parse this out of JSON and back into an instance.

humphd (changes requested):
This is getting better and better, nice work.  I've left a few thoughts for things I'd do differently.

src/lib/ChatCraftProvider.ts
Amnish04: I was about to comment something similar.

Shouldn't we rather use an **enum** here so we don't have to hard code string values like below
image

With an enum:
```
enum ProviderName {
  OPEN_AI = "OpenAI",
  OPEN_ROUTER = "OpenRouter.ai",
}

...
...

// Parses url into instance of ChatCraftProvider
static fromUrl(url: string) {
  let name = null;

  switch (url) {
    case OPENAI_API_URL:
      name = ProviderName.OPEN_AI;
      break;
    case OPENROUTER_API_URL:
      name = ProviderName.OPEN_ROUTER;
      break;
    default:
      throw new Error("Invalid URL");
  }

  return new ChatCraftProvider(name, url, "");
}
```

At this point, I feel like we could even have an `apiUrl` to `ProviderName` mapping. That could simplify this to\
```ts
// Parses url into instance of ChatCraftProvider
static fromUrl(url: string) {
  let name = null;

  if (mapping[url]) {
    const providerName = mapping[url];
    return new ChatCraftProvider(providerName, url, "");
  }
  else {
    throw new Error("Invalid URL");
  }
}
```

The decision depends on how we are our use case and vision for the future.

Amnish04:
@kliu57 I've been testing this for a while, and the first thing I was about to comment was that we shouldn't delete the existing `apiKey` when switching providers. But then I saw this
image
So I am assuming that's not an issue.

Rest works well and as you've described!

Just left a [comment](https://github.com/tarasglek/chatcraft.org/pull/371#discussion_r1473678243) for handling `ProviderNames`. But I am not sure what would be the best approach in our case. Maybe @humphd can confirm.

kliu57:
@humphd Made the following changes to `ChatCraftProvider`, addressing all your review comments
- changed `fromUrl` function to take `apiKey` param as well
- removed `setUrl` and `setKey` functions
- changed `name` member var of ChatCraftProvider from `string` data type to `ProviderName` where `type ProviderName = "OpenAI" | "OpenRouter.ai"`
- added `fromJSON` function

@Amnish04 I created an urlToNameMap object which can map the url to the name
```
const urlToNameMap: { [key: string]: ProviderName } = {
  [OPENAI_API_URL]: "OpenAI",
  [OPENROUTER_API_URL]: "OpenRouter.ai",
};
```

src/components/PreferencesModal.tsx
```diff
@@ -164,9 +165,14 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
                 value={settings.apiUrl}
-                onChange={(e) =>
-                  setSettings({ ...settings, apiUrl: e.target.value, apiKey: undefined })
-                }
+                onChange={(e) => {
+                  setSettings({
+                    ...settings,
+                    apiUrl: e.target.value,
+                    apiKey: undefined,
+                    providers: {},
```
humphd, line 173: Question: are you losing data here?  Why clear all values vs. the one you're working on?

humphd (approved):
Looks good, just one question.

kliu57:
> Looks good, just one question.

@humphd 

**Why I wanted to clear the key from settings.providers when the user toggled the API URL field:**

If I kept the key stored in settings.providers, the user may not know that we have it stored. For example they toggle from OpenAI to OpenRouter, and we kepts their OpenAI key stored in settings.providers. Say they toggle back to OpenAI, our UI shows a blank where the key is, even though we have their key stored in settings.providers. They would have no idea we stored their key and wouldn't have the ability to clear it from storage.

**Why I didn't I just clear the current key from settings.providers when the user toggled the API URL field (and instead set it to {} thus clearing all values ):**

I did this because in the next release, I won't even be clearing the key at all. The code to clear the key will be removed, so I figured it didn't matter if I cleared the current key by clearing the entire array, or I cleared the current key by removing it specifically. This code of clearing the key will be removed in the next release so that we can keep the stored provider when we toggle.

humphd:
Sounds good, I wanted to confirm we hadn't missed a case where we'd lose data without meaning to.  Thanks for clarifying.

Amnish04:
> @humphd Made the following changes to `ChatCraftProvider`, addressing all your review comments
> 
>     * changed `fromUrl` function to take `apiKey` param as well
> 
>     * removed `setUrl` and `setKey` functions
> 
>     * changed `name` member var of ChatCraftProvider from `string` data type to `ProviderName` where `type ProviderName = "OpenAI" | "OpenRouter.ai"`
> 
>     * added `fromJSON` function
> 
> 
> @Amnish04 I created an urlToNameMap object which can map the url to the name
> 
> ```
> const urlToNameMap: { [key: string]: ProviderName } = {
>   [OPENAI_API_URL]: "OpenAI",
>   [OPENROUTER_API_URL]: "OpenRouter.ai",
> };
> ```
> 
> Tested `fromJSON()` function to make sure it can parse JSON to `ChatCraftProvider` object:
> 
> image

@kliu57 Thanks for the changes. I would name the mapping something like `providerNamesByUrl` to be more specific. But that's just my personal preference, and it's up to you if you want to change that.

Overall, looks good to me!

Amnish04 (approved):

PR #374, opened 2024-01-29
Created a react-menu component and migrated MessageBase menu to use the component

### LINKED ISSUE #359: Use a different Menu component
ChakraUI has been amazing, but it lacks a menu component with everything we need.  Specifically, we are in need of sub-menus.

Let's add a new React menu component to our code.  I think https://szhsin.github.io/react-menu looks good, but if people have another idea, I'm open to suggestions. 

I don't want to pull in a whole UI framework, just a menu.

I'd also like to get the default styling of this menu component to match what we have as close as possible, make sure light/dark mode works with our current code, etc.

Rachit1313:
Fixes #359


humphd (changes requested):
Thanks for getting started on this.  Some initial feedback:

We discussed this a bit on Discord, but I wanted to circle back on it now that I've had a chance to test.  The react-menu demo page does do highlight-on-hover, why does this not support that?

Screenshot 2024-01-29 at 8 52 08 AM

The way you're doing submenus via props on the `Menu`, can we expose more of the react-menu API so you can do things declaratively?

```tsx
<Menu menuButton={<MenuButton>Menu</MenuButton>}>
  <MenuItem>New File</MenuItem>
  <SubMenu label="Edit">
    <MenuItem>Cut</MenuItem>
    <MenuItem>Copy</MenuItem>
    <MenuItem>Paste</MenuItem>
    <SubMenu label="Find">
      <MenuItem>Find...</MenuItem>
      <MenuItem>Find Next</MenuItem>
      <MenuItem>Find Previous</MenuItem>
    </SubMenu>
  </SubMenu>
  <MenuItem>Print...</MenuItem>
</Menu>
```

We've lost the border, which I think is important:

Screenshot 2024-01-29 at 8 55 23 AM

I think you should use their built-in `theming` vs hard-coding colours:


theming | string |   | It sets a CSS class on the container element of menu for theming purpose. E.g., 'dark' will add szh-menu-container--theme-dark.
-- | -- | -- | --

We need things like dividers, to be able to style menu items, etc.

Rachit1313:
> The way you're doing submenus via props on the `Menu`, can we expose more of the react-menu API so you can do things declaratively?

By this you mean creating separate components for submenu right i.e creating a saperate SubMenu.tsx?

humphd:
Right.  I would assume you'd expose the various components from react-menu in a wrapped form so they play nicely with the rest of our stuff.  So you can create `src/components/Menu/*` and put all the different things we want in there.  I'm not sure we need to expose everything at first, but we should have all the common stuff: menu, submenu, menuitems, divider, etc.

Rachit1313:
> Right. I would assume you'd expose the various components from react-menu in a wrapped form so they play nicely with the rest of our stuff. So you can create `src/components/Menu/*` and put all the different things we want in there. I'm not sure we need to expose everything at first, but we should have all the common stuff: menu, submenu, menuitems, divider, etc.

Okay., For the duplicate chat until message 
image
This is the current functionality, but the react-menu's MenuItem does not support as or to prop. Is that okay if I handle it with this?
image

humphd:
I think you can do it, see https://github.com/szhsin/react-menu/issues/248#issuecomment-910261822

Rachit1313:
> I think you should use their built-in `theming` vs hard-coding colours:
> 
> theming	string	 	It sets a CSS class on the container element of menu for theming purpose. E.g., 'dark' will add szh-menu-container--theme-dark.

I got all the components ready and updated in the MessageBase class declaratively. I need someone's help with css part. I am stuck on that and not able to match the theming perfectly.

humphd:
@Amnish04 is good at CSS.  Can you help?

Amnish04:
I've been looking at this PR for a while, and one thing I noticed **except** theming is that we have a weird **hover resize** on messages that did not exist before.
MessagesHoverResize

I think that's because the `MenuButton` from [React-Menu](https://szhsin.github.io/react-menu#menu-button) is a smaller size than buttons from **Chakra**, and when the appear they resize the container div pushing stuff down.

We can probably use a Chakra [IconButton](https://chakra-ui.com/docs/components/icon-button/usage) from before to fix this issue.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
...
+              size="sm"
+              aria-label="Options"
+              backgroundColor={menuButtonBg}
+              color={menuButtonColor}
+              _hover={{ bg: menuButtonBg }}
+              _active={{ bg: menuButtonBg }}
+              _focus={{ boxShadow: theme.shadows.outline }}
+            />
+          </MenuButton>
+        }
```
Amnish04, line 35:
```suggestion
        menuButton={
          <IconButton
            aria-label="Message Menu"
            icon={<TbDots />}
            variant="ghost"
            isDisabled={isLoading}
          ></IconButton>
        }
```

This fixes the issue I mentioned [above](https://github.com/tarasglek/chatcraft.org/pull/374#issuecomment-1918133731)

Amnish04 (changes requested):

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { Menu as ReactMenu, MenuButton } from "@szhsin/react-menu"; // Alias for the imported Menu
```
Amnish04, line 2:
```suggestion
import { Menu as ReactMenu } from "@szhsin/react-menu"; // Alias for the imported Menu
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Box, IconButton, useColorModeValue, useTheme } from "@chakra-ui/react";
+import { TbDots } from "react-icons/tb";
+
+interface MenuProps {
+  children: React.ReactNode;
+}
+
+const Menu: React.FC<MenuProps> = ({ children }) => {
+  const theme = useTheme();
+  const menuButtonBg = useColorModeValue(theme.colors.gray[100], theme.colors.gray[700]);
```
Amnish04, line 13:
```suggestion
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { Menu as ReactMenu, MenuButton } from "@szhsin/react-menu"; // Alias for the imported Menu
+import "@szhsin/react-menu/dist/index.css";
+import { Box, IconButton, useColorModeValue, useTheme } from "@chakra-ui/react";
+import { TbDots } from "react-icons/tb";
+
+interface MenuProps {
+  children: React.ReactNode;
```
Amnish04, line 8:
```suggestion
  children: React.ReactNode;
  isLoading?: boolean;
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const Menu: React.FC<MenuProps> = ({ children }) => {
```
Amnish04, line 11:
```suggestion
const Menu: React.FC<MenuProps> = ({ children, isLoading }) => {
```

src/components/Message/MessageBase.tsx
```diff
@@ -278,55 +278,48 @@ function MessageBase({
                 </ButtonGroup>
               )}
               <Menu>
```
Amnish04, line 280:
```suggestion
              <Menu isLoading={isLoading}>
```

Amnish04 (changes requested):
Also added an optional `isLoading` property as I noticed previously `MenuButton` was disabled when loading.

https://github.com/tarasglek/chatcraft.org/blob/9307f3a6a9f3f2ce6045ea7016cb2dce0ef4c35c/src/components/Message/MessageBase.tsx#L299

src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { MenuItem as ReactMenuMenuItem } from "@szhsin/react-menu";
```
Amnish04, line 2:
```suggestion
import { MenuItem as ReactMenuItem } from "@szhsin/react-menu";
```

Amnish04 (changes requested):
I think your custom styling for `MenuItem` was **overriding** the **hover effect** for menu items.
These changes fix that.
HoverHighlight


Also I renamed `MenuItem` **alias** from `ReactMenuMenuItem` to `ReactMenuItem` if that's okay :)

```
import { MenuItem as ReactMenuItem } from "@szhsin/react-menu";
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const theme = useTheme();
+  const menuButtonBg = useColorModeValue(theme.colors.gray[100], theme.colors.gray[700]);
+  const menuButtonColor = useColorModeValue(theme.colors.black, theme.colors.white);
+  const menuListBg = useColorModeValue(theme.colors.white, theme.colors.gray[700]);
+  const menuListBorderColor = useColorModeValue(theme.colors.gray[200], theme.colors.gray[600]);
```
Amnish04, line 16:
```suggestion
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,49 @@
...
+              _focus={{ boxShadow: theme.shadows.outline }}
+            />
+          </MenuButton>
+        }
+        menuStyle={{
+          borderColor: menuListBorderColor,
+          backgroundColor: menuListBg,
+          color: menuButtonColor,
+          boxShadow: theme.shadows.md,
+        }}
```
Amnish04, line 34:
```suggestion
```

Amnish04 (changes requested):
I am not sure if these styles are doing anything we need. Looked better to me after removing these.

Please let me know if I am missing something

Amnish04:
@Rachit1313 I would suggest using this option to add changes to batch instead of **commit suggestion**, so there's only one commit and no CI errors :)
image

Rachit1313:
> @Rachit1313 I would suggest using this option to add changes to batch instead of **commit suggestion**, so there's only one commit and no CI fails :) image

Thanks and
image

After checking your changes I see that its good in the light mode but we need to do something for the dark mode.

Rachit1313:
> @Rachit1313 I would suggest using this option to add changes to batch instead of **commit suggestion**, so there's only one commit and no CI fails :) image

btw the add suggestion to batch is disabled for me.

Amnish04:
> > @Rachit1313 I would suggest using this option to add changes to batch instead of **commit suggestion**, so there's only one commit and no CI fails :) image
> 
> btw the add suggestion to batch is disabled for me.

I think you need to go to the Files tab for that

image

Amnish04:
> > @Rachit1313 I would suggest using this option to add changes to batch instead of **commit suggestion**, so there's only one commit and no CI fails :) image
> 
> Thanks and image
> 
> After checking your changes I see that its good in the light mode but we need to do something for the dark mode.

I think that's something related to [theming](https://szhsin.github.io/react-menu/docs) property. Still looking into it.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,42 @@
...
+const Menu: React.FC<MenuProps> = ({ children, isLoading }) => {
+  const theme = useTheme();
+  const menuButtonColor = useColorModeValue(theme.colors.black, theme.colors.white);
+  const menuListBg = useColorModeValue(theme.colors.white, theme.colors.gray[700]);
+  const menuListBorderColor = useColorModeValue(theme.colors.gray[200], theme.colors.gray[600]);
+
+  return (
+    <Box>
+      <ReactMenu // Use the alias here
+        menuButton={
```
Amnish04, line 21:
```suggestion
        transition={true}
        menuButton={
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import { Menu as ReactMenu } from "@szhsin/react-menu"; // Alias for the imported Menu
+import "@szhsin/react-menu/dist/index.css";
```
Amnish04, line 3:
```suggestion
import "@szhsin/react-menu/dist/index.css";
import "@szhsin/react-menu/dist/transitions/slide.css";
```

Amnish04 (changes requested):
Adding this property gives similar open close animation like we had with **Chakra**

OpenCloseTransition


This comes with `@szhsin/react-menu/dist/transitions/slide.css`

humphd:
@Amnish04 @Rachit1313 another approach you can take to doing something like this is for each of you to push/pull commits to the branch.

This approach requires a bit of care, but is a great way to work together.  Before you do it, ask the other person if they are OK with you pushing commits.

Then, check out their code:

```
# Assuming tarasgleck/chatcraft.org is your upstream remote
git fetch remote
# Get the branch for this PR
git checkout react-menu
# do your changes
git add ...
git commit -m "Fixing ..."
git push upstream react-menu
```

Now, the key here is to always `git pull upstream react-menu` before you start doing more work, or you risk getting out of sync.

If that happens, you can do:

```
git pull --rebase upstream react-menu
```

You don't have to work this way, it's a bit more advanced and both parties need to feel comfortable with git to do it.  But FYI, you can, since we're working on a branch in the same repo.

Also, I'm really glad to see you to helping each other get this landed.  Amazing work!

Amnish04:
> @Amnish04 @Rachit1313 another approach you can take to doing something like this is for each of you to push/pull commits to the branch.
> 
> This approach requires a bit of care, but is a great way to work together. Before you do it, ask the other person if they are OK with you pushing commits.
> 
> Then, check out their code:
> 
> ```
> # Assuming tarasgleck/chatcraft.org is your upstream remote
> git fetch remote
> # Get the branch for this PR
> git checkout react-menu
> # do your changes
> git add ...
> git commit -m "Fixing ..."
> git push upstream react-menu
> ```
> 
> Now, the key here is to always `git pull upstream react-menu` before you start doing more work, or you risk getting out of sync.
> 
> If that happens, you can do:
> 
> ```
> git pull --rebase upstream react-menu
> ```
> 
> You don't have to work this way, it's a bit more advanced and both parties need to feel comfortable with git to do it. But FYI, you can, since we're working on a branch in the same repo.
> 
> Also, I'm really glad to see you to helping each other get this landed. Amazing work!

Thanks for sharing! I was having an itch to push directly from the beginning, but wasn't sure if that's acceptable 😝

@Rachit1313 If it's okay with you, can I push changes directly here? We can always reset if things go in the wrong direction.

Amnish04:
Just pushed the following changes:

1. Switch to dark mode theme based on `colorMode` state. 
image
2. A Menu open/close transition.
image
3. Added a `Menu.css` file to adapt the **React-Menu** theming to align with Chakra colors.
```css
/*
  Colors from official Chakra docs
  https://chakra-ui-git-fix-typescript-autocomplete.chakra-ui.vercel.app/docs/theming/theme#gray

  Classes from official @szhsin/react-menu docs
  https://szhsin.github.io/react-menu/style-guide
*/
.szh-menu-container {
  /* Light Mode */
  --chakra-gray-100: #edf2f7;
  --chakra-gray-200: #e2e8f0;

  /* Dark Mode */
  --chakra-gray-500: #718096;
  --chakra-gray-600: #4a5568;
  --chakra-gray-700: #2d3748;
}
...
...
```
4. Some [block padding](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block) adds attention to detail. 
image

@Rachit1313 @humphd 
Please let me know if its good so far.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface MenuProps {
+  children: React.ReactNode;
+  isLoading?: boolean;
+}
+
+const Menu: React.FC<MenuProps> = ({ children, isLoading }) => {
+  const { colorMode } = useColorMode();
+  return (
+    <Box>
+      <ReactMenu // Use the alias here
```
humphd, line 23: Get rid of this comment.

src/components/Menu/Menu.tsx
humphd: Is this still necessary?  I see `transition={true}` below.

src/components/Menu/MenuDivider.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const MenuDivider: React.FC = () => {
+  const dividerColor = useColorModeValue("gray.200", "gray.600");
+
+  return (
+    <Box
+      as={ReactMenuDivider}
+      borderColor={dividerColor}
+      my={1} // Adjust margin as needed
+      // Add any additional styles you want to apply to the divider
```
humphd, line 13: Get rid of this

src/components/Menu/MenuDivider.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Box, useColorModeValue } from "@chakra-ui/react";
+
+const MenuDivider: React.FC = () => {
+  const dividerColor = useColorModeValue("gray.200", "gray.600");
+
+  return (
+    <Box
+      as={ReactMenuDivider}
+      borderColor={dividerColor}
+      my={1} // Adjust margin as needed
```
humphd, line 12: Same, drop this comment.

src/components/Message/MessageBase.tsx
```diff
@@ -34,8 +29,13 @@ import {
   Kbd,
   Spacer,
 } from "@chakra-ui/react";
+
+import Menu from "../Menu/Menu"; // Assuming Menu is the new component
```
humphd, line 33: Can we create `../Menu/index.tsx` and re-export all of the different files?  A nicer API would be:

```ts
import { Menu, MenuItem, SubMenu, MenuDivider } from '../Menu';
```

Also, drop these comments.

src/components/Message/MessageBase.tsx
```diff
@@ -277,56 +277,49 @@ function MessageBase({
...
+                <MenuItem label="Copy" onClick={handleCopy} />
+                <MenuItem label="Download" onClick={handleDownload} />
+                {!disableFork && (
+                  <MenuItem
+                    label={
+                      <Link as={ReactRouterLink} to={`./fork/${id}`} target="_blank">
+                        Duplicate Chat until Message...
+                      </Link>
+                    }
+                    onClick={() => {}} // You can leave this empty or handle additional logic if needed
```
humphd, line 290: What is the purpose of `onClick` here?

src/components/SubMenu.css
```diff
@@ -0,0 +1,6 @@
+.szh-menu__submenu {
```
humphd, line 1: Let's move these `.css` files related to the menu into `src/componenents/Menu/*`

humphd (changes requested):
@Rachit1313 @Amnish04 I think this is really looking great, and we're on track to have it in for 1.1.  What a huge improvement over what we have now!

@Rachit1313 please see my comments below and let's get this fixed up to land before Friday's deadline.  If you need further help from @Amnish04, please let him know.

src/components/Message/MessageBase.tsx
Rachit1313: nothing, its just I set onClick as a required prop so had to leave it empty.  If you want I can make it optional and remove this.

src/components/Message/MessageBase.tsx
```diff
@@ -1,4 +1,4 @@
-import {
+import React, {
```
humphd, line 1: Why are we adding `React` here?

src/components/Message/MessageBase.tsx
```diff
@@ -277,56 +274,48 @@ function MessageBase({
...
+                )}
 
-                  {(!disableEdit || onDeleteClick) && <MenuDivider />}
-                  {!disableEdit && (
-                    <MenuItem onClick={() => onEditingChange(!editing)}>
-                      {editing ? "Cancel Editing" : "Edit"}
-                    </MenuItem>
-                  )}
-                  {onDeleteBeforeClick && (
-                    <MenuItem onClick={() => onDeleteBeforeClick()} color="red.400">
```
humphd, line 315: We've lost `color="red.400"` for these, can you add that back, either in this PR or a follow-up?

humphd (approved):
I think this is excellent. I have two questions, but neither is a blocker for me.  Just fix or file issues to deal with it later.

Amazing work.

src/components/Message/MessageBase.tsx
Rachit1313: Yeah I see. its not getting used anywhere so looks like might be auto imported while making some changes. I'll just remove it.

Rachit1313:
> I think this is excellent. I have two questions, but neither is a blocker for me. Just fix or file issues to deal with it later.
> 
> Amazing work.

Made the changes.

src/components/Menu/Menu.css
```diff
@@ -44,3 +44,13 @@
 .szh-menu-container--theme-dark .szh-menu__item:active {
   background-color: var(--chakra-gray-500);
 }
+
+/* Light Mode Delete Button Color */
+.delete-button {
+  color: #fc8181; /* Chakra UI's red.400 */
+}
```
Amnish04, line 51: Can we define a css variable for this?

```css
--chakra-red-400: #fc8181;
```

src/components/Menu/Menu.css
```diff
@@ -44,3 +44,13 @@
...
+
+/* Dark Mode Delete Button Color */
+.szh-menu-container--theme-dark .delete-button {
+  color: #fc8181; /* Chakra UI's red.400 for dark mode */
+}
```
Amnish04, line 56:
```suggestion
```

I don't think we need to have this style if we're using same color in both cases

Amnish04 (changes requested):
@Rachit1313 I've left some comments for minor changes

Amnish04:
Other than that, I also observed a couple of unexpected behaviours.

**1.** When opening submenu, the view automatically exhibits some weird scroll.
NewScrollBug

I don't think that was happening before.

**2.** When the sidebar is open or on a lower width, opening the menu makes a horizontal scrollbar appear.
ScrollbarAppears

I think the **second** issue can be fixed in a follow up.

Should we try to address the first one now?

humphd:
Yes, that scroll bug is not good.  Let's get that sorted before we merge.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <Box>
+      <ReactMenu
+        theming={colorMode === "dark" ? "dark" : undefined}
+        menuButton={
+          <IconButton
+            aria-label="Message Menu"
+            icon={<TbDots />}
+            variant="ghost"
+            isDisabled={isLoading}
+          ></IconButton>
```
WangGithub0, line 31: If a component doesn't have any children, you can use a self-closing tag 
  <IconButton
            aria-label="Message Menu"
            icon={<TbDots />}
            variant="ghost"
            isDisabled={isLoading}
          />

Rachit1313:
> Yes, that scroll bug is not good. Let's get that sorted before we merge.

@Amnish04 @humphd , For me I can't see the scroll to top and the scroll bar, both issues. Can you please confirm?

Rachit1313:
https://github.com/tarasglek/chatcraft.org/assets/58596563/574e2397-a8a3-4295-8739-1cbc1d5ba39f

Amnish04:
@Rachit1313 The scroll bug I mentioned isn't happening for me either, it was only limited to that particular browser tab I was testing in.
image

But the second one is still there for me. Not sure what's up.
Feb-01-2024 13-32-47

Rachit1313:
> @Rachit1313 The scroll bug I mentioned isn't happening for me either, it was only limited to that particular browser tab I was testing in.
> 
> image
> 
> 
> 
> But the second one is still there for me. Not sure what's up.
> 
> Feb-01-2024 13-32-47
> 
> 

Can you check if it shows you this even in cloudflare preview Url?

Amnish04:
@Rachit1313 Just retested, I see same behavior there as well.
image


@humphd Is this happening for you as well, or is it just me?

humphd:
> @humphd Is this happening for you as well, or is it just me?

I've tested on macOS with Chrome and Safari, and I can't reproduce this bug.  It might be Windows only?  Or specific to the browser you are using?

Amnish04:
> > @humphd Is this happening for you as well, or is it just me?
> 
> I've tested on macOS with Chrome and Safari, and I can't reproduce this bug. It might be Windows only? Or specific to the browser you are using?

Not sure why its only happening for me then. I tested in both **Windows** and **Mac** - **Firefox** and **Chrome**.

Maybe we have different scaling settings? I have mine set to 100%.

In any case, if it's not happening for anyone else we should be good to land this for now.
And open a follow up if people face issues.

humphd:
@Amnish04 agreed.  I think if you hit it, it's possible the bug is there and subtle.  Let's land this and improve later on.

Rachit1313:
> @Amnish04 agreed. I think if you hit it, it's possible the bug is there and subtle. Let's land this and improve later on.

Are we good to merge this for release 0.1, will work as another issue if anything comes up.

Amnish04 (approved):

tarasglek:
great fix, thank you @Amnish04

humphd:
> great fix, thank you @Amnish04

Great team effort by @Rachit1313 and @Amnish04.  Love to see the collaboration here.

PR #375, opened 2024-01-29
Chore: Don't show hints when editing on mobile

### LINKED ISSUE #289: Don't show keyboard hints when editing on mobile
I notice that the newly updated Edit form is not optimized for mobile:

Screenshot 2023-11-14 at 6 45 45 AM

Two things here:

1. For some reason, the MacOS check breaks when I try this on iOS Safari and in Chrome when in Dev Tools and using a mobile viewport (i.e., `Command` becomes `Ctrl`).
2. There isn't room for the keyboard hint, so it wraps

Let's fix both 1. and 2. by not showing the keyboard hints on mobile.

rjwignar:
This fixes #289

Summary
---

Uses [use-mobile-breakpoint](https://github.com/tarasglek/chatcraft.org/blob/main/src/hooks/use-mobile-breakpoint.ts) (as `isNarrowScreen`) to render the keyboard hint when on mobile (when ChatCraft detects a screen below the 480px cutoff).

### Without conditional rendering (iPhone SE)
image

### With conditional rendering (iPhone SE)
image

### How to test
- Run `npm run dev`
- submit any query to ChatCraft
- Left-click "edit message" on your message
image

- Enter your browser's developer tools by pressing **F12** (or equivalent) on your keyboard
- Enable device emulation
- Select viewports with a pixel width **<= 480px** (e.g., **iPhone SE**, **iPhone 14 Pro Max**)

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -352,12 +352,14 @@ function MessageBase({
                     />
                     <Flex width="100%" alignItems="center" alignContent="end" gap={2}>
                       <Spacer />
-                      <Text fontSize="sm" color="gray">
-                        <span>
-                          <Kbd>{meta}</Kbd> + <Kbd>Enter</Kbd>
-                          <span> to save</span>
-                        </span>
-                      </Text>
+                      {!isNarrowScreen && (
+                        <Text fontSize="sm" color="gray">
+                          <span>
+                            <Kbd>{meta}</Kbd> + <Kbd>Enter</Kbd>
+                            <span> to save</span>
+                          </span>
+                        </Text>
+                      )}
                       <Button size="sm" variant="ghost" onClick={() => onEditingChange(false)}>
                         Cancel
                       </Button>


Amnish04:
@rjwignar This works well on a mobile device.

But I noticed that if I change the screen size to an **IPad**, it we still get that hint even though there's no `Ctrl+Enter` on mobile devices.
image

Maybe our mobile breakpoint hook won't be enough for this.

I asked ChatCraft and it says to check the [navigator.userAgent](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent) property with the following example

```js
if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)) {
    console.log('You are on a mobile device');
} else {
    console.log('You are on a PC');
}
```

But I am not sure if that's the best way to do it.

@humphd What do you think about this?

humphd:
I think it's fine for iPad widths (everything fits).  Many people use keyboards with iPads, so this isn't a huge deal.

Let's land this if the rest is good.

Amnish04 (approved):
@humphd I see. That was the only concern. Looks good to me!

Issue #330, opened 2024-01-13
Allow users to authenticate with Google OAuth


humphd:
We currently support a limited GitHub OAuth for logging in and unlocking some features (e.g., sharing, being able to invoke certain API functions on CloudFlare, etc).  It would be good to also allow users to use their Google account.

WangGithub0:
I wanna try this one.

humphd:
@WangGithub0 excellent.  I'm not 100% sure what's involved with doing this.  I suspect that @tarasglek is going to need to set up some kind of "app" with Google, which configured the accepted URL endpoints.

Probably the first step for you is to add a list of all the steps we need to do in this issue.

tarasglek:
Let me know what I need to help with here. I think anyone can setup auth with google using own token and it's easy to migrate to other people's tokens.

WangGithub0:
I'll start by integrating Google Sign-In into chatcraft according to [google's doc](https://developers.google.com/identity/sign-in/web/sign-in#sign_out_a_user), I plan to proceed as follows:

1. [Create authorization credentials](https://developers.google.com/identity/sign-in/web/sign-in#create_authorization_credentials), this step need some developer contact info, @tarasglek could you help me set this step and provide app's client ID for me?
2. write a functions/google.ts and refer to github.ts
3. write functions/google.test.ts
4. add google login in [functions/api/login.ts](https://github.com/tarasglek/chatcraft.org/blob/64b0b0ab37d87e2695a12738feea8da1795d17e8/functions/api/login.ts#L14)
5. add login test

humphd:
This sounds good @WangGithub0.  A few more thoughts:

1. I would break this into pieces.  For example, we can begin with the CloudFlare bits, but not connect them in the UI.  Follow-up PRs can integrate things.  You could also write the implementation of the function and test separate.  More, smaller, PRs are better.
2. Our UI will eventually need to evolve to include different sign-in options:

<img width="236" alt="Screenshot 2024-01-29 at 4 15 07 PM" src="https://github.com/tarasglek/chatcraft.org/assets/427398/392f3a50-18ca-4136-aeba-32e150c8c405">

We might change that menu item to "Sign In..." and have it open a modal dialog with different options for GitHub, Google, etc.  Doesn't have to be done that way, just thinking out loud.

PR #380, opened 2024-01-30
google oath keys + privacy policy + tos

tarasglek:
Added oauth secrets for #330 . For now I got it as a test app, with my and @humph emails. Please share any other emails you need added


humphd:
Awesome, thank you @tarasglek.  @WangGithub0 is this good to merge?

PR #393, opened 2024-02-01
Use material design icons for tts button

### LINKED ISSUE #392: Use a better icon for TTS button
Follow up for #357 

Currently, the icons used for **TTS** button are not very descriptive of the setting's state.

**Enabled:**
image

**Disabled:**
image

We need to use better icons as suggested by @tarasglek at
https://github.com/tarasglek/chatcraft.org/pull/357#discussion_r1473577322

Amnish04:
I've replaced the current icons with the material design icons as suggested in
https://github.com/tarasglek/chatcraft.org/pull/357#discussion_r1473577322

**Enabled:**
image

**Disabled:**
image

This fixes #392

### DIFF
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -14,7 +14,7 @@ import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
 import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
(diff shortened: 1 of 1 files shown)


humphd (approved):
Let's go for this.  I think, longer term, once #359 lands, we should think about moving some of these secondary options to menus (e.g., you can toggle Text to Speech on in a menu when picking your model).

But having it visible right now while we iterate quickly on it is good.  Already you can see that merging your work has elicited feedback you couldn't get in review!

PR #396, opened 2024-02-01
Add a command to show list of supported commands

### LINKED ISSUE #356: Show help for unknown command
I noticed that when a user enters an invalid command, we get the following toast message.

image

I found a **TODO** while working on another issue that we need to **handle** this as appropriate for the application.

image

The regular `/help` command shows a generic list of in guidelines for the app. 

Should we rather only show the commands supported by **ChatCraft** when this happens?

Which would be **this section**:
image

Amnish04:
Currently, when a user enters an `unsupported command`, a **toast message** is displayed suggesting the commands was invalid.

I have added a new command called `/commands` that allows users to **quickly** refer to the list of commands instead of the general `/help` command.
image

The list of commands is also displayed when an invalid command is entered instead of showing the toast message.

This fixes #356

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -168,14 +168,19 @@ function ChatBase({ chat }: ChatBaseProps) {
             });
           }
         } else {
-          error({
-            title: `Unknown Command`,
-            message: `Command not recognized. Use /help to get help on valid commands.`,
-          });
-
-          console.log("TODO: show help");
           // The input was a command, but not a recognized one.
           // Handle this case as appropriate for your application.
+          const commandFunction = ChatCraftCommandRegistry.getCommand("/commands")!;
+          setShouldAutoScroll(true);
+          try {
+            await commandFunction(chat, user);
+            forceScroll();
+          } catch (err: any) {
+            error({
+              title: `Unknown Command`,
+              message: `Command not recognized. Use /help to get help on valid commands.`,
+            });
+          }
         }
 
         setLoading(false);
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -23,6 +23,7 @@ import { ChatCraftFunction } from "../lib/ChatCraftFunction";
 import { useAutoScroll } from "../hooks/use-autoscroll";
 import { useAlert } from "../hooks/use-alert";
 import { ChatCraftCommandRegistry } from "../lib/commands";
+import { ChatCraftCommand } from "../lib/ChatCraftCommand";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -170,13 +171,13 @@ function ChatBase({ chat }: ChatBaseProps) {
         } else {
           // The input was a command, but not a recognized one.
           // Handle this case as appropriate for your application.
-          const commandFunction = ChatCraftCommandRegistry.getCommand("/commands")!;
+
+          // We are sure that this won't return null
+          // since prompt is definitely a command
+          const { command } = ChatCraftCommand.parseCommand(prompt)!;
+          const commandFunction = ChatCraftCommandRegistry.getCommand(`/commands ${command}`)!;
           setShouldAutoScroll(true);
           try {
-            error({
-              title: `Unknown Command`,
-              message: `Command '${prompt}' not recognized. Please refer to the list of supported commands.`,
-            });
             await commandFunction(chat, user);
             forceScroll();
           } catch (err: any) {
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -173,12 +173,16 @@ function ChatBase({ chat }: ChatBaseProps) {
           const commandFunction = ChatCraftCommandRegistry.getCommand("/commands")!;
           setShouldAutoScroll(true);
           try {
+            error({
+              title: `Unknown Command`,
+              message: `Command '${prompt}' not recognized. Please refer to the list of supported commands.`,
+            });
             await commandFunction(chat, user);
             forceScroll();
           } catch (err: any) {
             error({
-              title: `Unknown Command`,
-              message: `Command not recognized. Use /help to get help on valid commands.`,
+              title: `Error Running Command`,
+              message: `There was an error running the command: ${err.message}.`,
             });
           }
         }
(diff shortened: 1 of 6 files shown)


rjwignar (changes requested):
The feature works, and the code looks great!
I've added some general comments but the only change I require before approving this is **updating the Commands** section of the `helpText` constant in `src/components/Message/AppMessage/Help.tsx` to include a description for your new command.

src/Chat/ChatBase.tsx
```diff
@@ -168,14 +168,19 @@ function ChatBase({ chat }: ChatBaseProps) {
...
           // The input was a command, but not a recognized one.
           // Handle this case as appropriate for your application.
+          const commandFunction = ChatCraftCommandRegistry.getCommand("/commands")!;
+          setShouldAutoScroll(true);
+          try {
+            await commandFunction(chat, user);
+            forceScroll();
+          } catch (err: any) {
+            error({
+              title: `Unknown Command`,
```
tarasglek, line 180: Why is /commands better than showing /help with all commands?

We should show the command that we failed to identify. I get a lot of errors from chatcraft when i try to paste /something and it's not immediately obvious without an indicator of what i typed

mingming-ma: it reminds me of 
image

rjwignar: > We should show the command that we failed to identify. I get a lot of errors from chatcraft when i try to paste /something and it's not immediately obvious without an indicator of what i typed

That's a good point.
Would it be better to have that indicated in `/commands` or inside the original Toast error message we currently have?
Something like below?

**/commands**
image


**Toast Message:**
image

src/Chat/ChatBase.tsx
```diff
@@ -168,14 +168,19 @@ function ChatBase({ chat }: ChatBaseProps) {
...
+              message: `Command not recognized. Use /help to get help on valid commands.`,
+            });
```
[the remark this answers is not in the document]
Amnish04: That is just a fallback if for some reason, the execution of `/commands` command runs into an exception (which I don't think should ever happen).

Should we remove it?

src/components/Message/AppMessage/Help.tsx
```diff
@@ -133,9 +133,31 @@ have an idea for a feature, or think you've found a bug, get in touch with us:
...
+Some commands accept arguments as well.
+
+| Command | Description |
+|-----|------|
+| /help         | Shows this help message. |
+| /commands | Shows a list of **supported commands** in ChatCraft |
+| /new          | Creates a new chat. |
+| /clear        | Erases all messages in the current chat. |
+| /summary&nbsp;[max-length] | Uses ChatGPT to create a summary of the current chat. Optionally takes a maximum word length (defaults to 500). |
+| /import&nbsp;<url> | Loads the provided URL and imports the text. Where possible, ChatCraft will try to get raw text vs. HTML from sites like GitHub. NOTE: to prevent abuse, you must be logged into use the import command. |`;
```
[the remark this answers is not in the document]
Amnish04: Sure, I missed that. Thanks for noticing

src/Chat/ChatBase.tsx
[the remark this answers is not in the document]
humphd: "Should never happen" is not the same as "Is impossible."  I'd leave it so we don't crash.  If it never shows the error toast, that's fine.

src/Chat/ChatBase.tsx
Amnish04: @tarasglek The `/help` command displays all the general help information related to ChatCraft. So it gets a little harder for user to **skim** through and **find** the **supported** commands.
image

Using `/commands`, it just becomes a little more convenient to look them up.
image

But you're right, we should also show the command that we typed as its not even added to the chat.

@rjwignar I think we should add it to the toast message?

@mingming-ma I like the idea of suggesting a similar command. Should we have a separate issue for that?

Amnish04:
Just pushed some changes. I've restored the toast message for `Unknown Command` and now it also displays what command was typed. The list of commands is also displayed.
image

src/Chat/ChatBase.tsx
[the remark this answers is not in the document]
Amnish04: You're right. I did not remove it, but changed the wording to be same as if it happened with any other command since the `Unknown Command` toast is always shown now
image

rjwignar (approved):
I like how the toast message displays the unrecognized command now.

With these changes, ChatCraft will now **display both** a Toast error message print the list of commands (by invoking `/commands`) when an unrecognized command is sent:
image.

These changes have my approval, but I'd like to see how others feel about both being displayed. @humphd, @mingming-ma, what are your thoughts on this?

mingming-ma (commented):
For the most similar commands, I think it might be not so helpful if we have all commands displayed. I think we can import the `commandsHelpText` to `Help` so when new commands coming in we don't need to change twice

src/components/Message/AppMessage/Help.tsx
```diff
@@ -91,6 +91,7 @@ Some commands accept arguments as well.
 | Command | Description |
 |-----|------|
 | /help         | Shows this help message. |
+| /commands | Shows a list of **supported commands** in ChatCraft |
 | /new          | Creates a new chat. |
 | /clear        | Erases all messages in the current chat. |
 | /summary&nbsp;[max-length] | Uses ChatGPT to create a summary of the current chat. Optionally takes a maximum word length (defaults to 500). |
```
mingming-ma, line 97:
```suggestion
${commandsHelpText}
```

Amnish04: @mingming-ma Agreed, I'll commit that.

tarasglek:
I don't think we need to toast here, how about we show the error inline in the commands help message reply?

Amnish04:
> I don't think we need to toast here, how about we show the error inline in the commands help message reply?

Yes, I think that makes more sense.

Amnish04:
@tarasglek Just pushed the changes.

Any invalid command now shows up inline with the commands help text.

Here are the scenarios:

**1. /typo**
image

**2. /commands**
image

**3. /commands query-some-command**
image

I am now allowing the `CommandsHelpCommand` to accept `arguments` which we can later use to implement a **suggest closest command** feature based on @mingming-ma's suggestion.

My only concern is if I could make my approach simpler.

tarasglek:
I think this is good for me

Amnish04:
@mingming-ma @rjwignar If you guys also think this is good, should I merge?

mingming-ma:
lgtm

rjwignar (approved):
Changes look great! LGTM

Amnish04:
Thanks for reviewing everyone!

PR #403, opened 2024-02-04
Allow running JS/TS remotely via val.town's eval endpoint

### LINKED ISSUE #384: Support running severside deno via val.town eval api
They offer a public eval endpoint. This is a good first step towards further integration.

Some examples:
```bash
curl -X POST https://api.val.town/v1/eval -H "Content-Type: application/json" -d '{"code":"async () => (await fetch(\"https://google.com\")).text()", "args":[]}'
```
```bash
curl -X POST https://api.val.town/v1/eval -H 'Content-Type: application/json' -d '{"code":"async() =>(Math.sqrt(1 + 3))", "args":[]}'
```

https://chatcraft.org/c/tarasglek/EO7koh0tZBBEdnJPR5Jf6

Idea is that in addition to run button for js, we also have a way to choose whether it runs serverside or clientside...When running serverside, we package up any code into an async anon function and POST it to api

I would like this to be a start of functionality of supporting valtown as a way to do custom /import-like stuff

humphd:
Adds the ability to run code in the browser, or remotely via val.town's [/eval endpoint](https://docs.val.town/api/eval/):

Screenshot 2024-02-04 at 2 21 50 PM

In testing this, I find their endpoint pretty fragile when it comes to the format of the code.  Lots of valid things don't seem to work, and you don't get good error messages back.

I've tried to make this safe to run and give a bit of feedback.  But it's not possible to make it perfect.

Fixes #384

### DIFF
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -180,7 +180,7 @@ function CodeHeader({
         </Box>
         <ButtonGroup isAttached pr={2}>
           {shouldShowRunButton && (
-            <Menu>
+            <Menu strategy="fixed">
               <MenuButton
                 as={IconButton}
                 size="sm"
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -7,6 +7,10 @@ import {
   useColorModeValue,
   Text,
   Box,
+  Menu,
+  MenuButton,
+  MenuItem,
+  MenuList,
 } from "@chakra-ui/react";
 import { TbCopy, TbDownload, TbRun, TbExternalLink } from "react-icons/tb";
 
@@ -32,7 +36,7 @@ function CodeHeader({
   codeDownloadFilename,
 }: PreHeaderProps) {
   const { onCopy } = useClipboard(code);
-  const { info } = useAlert();
+  const { info, error } = useAlert();
   // Only show the "Run" button for JS code blocks, and only when we aren't already loading
   const shouldShowRunButton = isRunnable(language) && onPrompt;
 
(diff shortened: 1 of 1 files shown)


tarasglek:
it only runs deno, unfortunately no way for us to detect valid deno without applying some treesitter grammar

humphd:
Moving to 1.3, as I haven't had reviews yet.

src/components/CodeHeader.tsx
```diff
@@ -52,7 +56,61 @@ function CodeHeader({
...
+    try {
+      if (typeof result === "string") {
+        // catch corner cases with strings
+        if (!result.length || result[0] === "/") {
+          result = formatAsCodeBlock(JSON.stringify(result), "js");
+        } else if (result.startsWith("<")) {
+          result = formatAsCodeBlock(result, "html");
+        } else {
+          // result is good to include inline, might have formatting, etc
+        }
```
rjwignar, line 101: Just curious, are there other corner cases we would have to look out for?

src/components/CodeHeader.tsx
```diff
@@ -122,17 +180,23 @@ function CodeHeader({
...
-              aria-label="Run code"
-              title="Run code"
-              icon={<TbRun />}
-              color="gray.600"
-              _dark={{ color: "gray.300" }}
-              variant="ghost"
-              onClick={handleRun}
-              isDisabled={isLoading}
-            />
+            <Menu>
```
rjwignar, line 183: I'm surprised this doesn't cause the hidden scrollbar issue we observed in https://github.com/tarasglek/chatcraft.org/issues/442.

Using `<Menu strategy="fixed">` here doesn't seem to make a difference based on what I've observed, so I think we can stick to `<Menu>`

rjwignar (approved):
@humphd 
I think this is good to ship.
I added a couple questions I'd appreciate your insight on, but it's nothing preventing me from approving this PR.

src/components/CodeHeader.tsx
humphd: @tarasglek figured these out back in the summer.  I can't remember all the details, but there are many cases where we get back things which can be rendered in better ways, but you have to do these checks.

src/components/CodeHeader.tsx
humphd: Likely because it's not at the edge of the window.  Probably I should add it.  What do you think @Amnish04?

Amnish04: I don't see a problem in adding it. It actually looks better in certain scenarios

**Absolute:**
image

**Fixed:**
image

humphd:
Added `strategy="fixed"`

src/components/CodeHeader.tsx
```diff
@@ -52,7 +56,61 @@ function CodeHeader({
...
+        title: "Server unable to parse code",
+        message: "Try rewriting the code as an async function returning a value.",
+      });
+      return;
+    }
+
+    try {
+      if (typeof result === "string") {
+        // catch corner cases with strings
+        if (!result.length || result[0] === "/") {
```
Amnish04, line 95: Why don't we do `result.startsWith` here instead of `result[0]`?

Amnish04, line 95: Why don't we do `result.startsWith` here instead of `result[0]`?

PR #404, opened 2024-02-04
Don't output undefined when logging in run code

### LINKED ISSUE #385: Do not output undefined if console.log was intercepted
See https://chatcraft.org/c/tarasglek/ZJ8pC2uHG6jglid4tE_o4

There there are 2 ways of how js run button behaves... We should avoid printing the undefined value if code produced output via console logging.

humphd:
When we run code, if the return value is `undefined` but we have logs, prefer the logs:

Screenshot 2024-02-04 at 4 31 37 PM

fixes #385

### DIFF
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -57,38 +57,51 @@ function CodeHeader({
       return;
     }
 
-    let ret = undefined;
     try {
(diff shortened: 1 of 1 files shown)


tarasglek (approved):
thank you

PR #408, opened 2024-02-05
HTML Preview: Fix Iframe auto-resize

### LINKED ISSUE #379: Iframe auto-resize broken
The auto-resize logic in HTML preview Iframes seems to break, cutting off the final line of content in HTML previews:

Screenshot 2024-01-29 at 8 42 51 AM

_Originally posted by @humphd in https://github.com/tarasglek/chatcraft.org/pull/376#pullrequestreview-1848709983_

rjwignar:
This fixes #379
This PR prevents content overflowing out of an iFrame by setting 

How to Test
---
Open two separate ChatCraft conversations and supply one prompt to each:
- First conversation: Show me a few example HTML pages with Plotly.js
- Second conversation: Show me three example HTML pages of varying sizes

Explanation
---

By default, the IframeResizer [heightCalculationMethod](https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/parent_page/options.md#heightcalculationmethod) option calculates the iFrame height by converting the `body` margin to px then adding the top and bottom figures to the `body` tag's `offSetHeight` (bodyOffSet).

With this default `heightCalculationMethod`, HTML content tends to overflow outside the body:
image

There are four main values for **heightCalculationMethod**:

- **bodyOffset** uses document.body.offsetHeight
- **bodyScroll** uses document.body.scrollHeight 
- **documentElementOffset** uses document.documentElement.offsetHeight
- **documentElementScroll** uses document.documentElement.scrollHeight 

A `heightCalculationMethod` value of **max** uses the largest value of the above 4, preventing overflow:
image

To keep or remove scrolling
---
I've experimented with other values for `heightCalculationMethod`. For example, `bodyScroll` also prevents text overflow, but when ChatCraft generates HTML pages with Plotly.js, scroll-bars are added to the preview. Using `heightCalculationMethod={max}` prevents both of these from happening.

Using `max` will render entire HTML previews, making scrollbars unnecessary. It is important to note that the default setting, `bodyOffset`, **already renders** ~95% of an HTML preview body.
**We must determine** whether we want to render entire IFrames, or add a `maxHeight` property to the Iframe to make scrollbars necessary.

However, in my tests, adding a maxHeight property to the IFrame allows text overflow/cutting off

### DIFF
--- a/src/components/HtmlPreview.tsx
+++ b/src/components/HtmlPreview.tsx
@@ -49,6 +49,7 @@ const HtmlPreview = ({ children, isLoading = false }: HtmlPreviewProps) => {
           src={url}
           style={{ width: "1px", minWidth: "100%" }}
           scrolling={true}
+          heightCalculationMethod={"max"}
         />
       </CardBody>
     </Card>
--- a/src/components/HtmlPreview.tsx
+++ b/src/components/HtmlPreview.tsx
@@ -48,7 +48,6 @@ const HtmlPreview = ({ children, isLoading = false }: HtmlPreviewProps) => {
           checkOrigin={false}
           src={url}
           style={{ width: "1px", minWidth: "100%" }}
-          scrolling={true}
           heightCalculationMethod={"max"}
         />
       </CardBody>


mingming-ma (approved):
I've tested and all passed. The explanation is really detailed 👍 Thanks!

src/components/HtmlPreview.tsx
```diff
@@ -49,6 +49,7 @@ const HtmlPreview = ({ children, isLoading = false }: HtmlPreviewProps) => {
           src={url}
           style={{ width: "1px", minWidth: "100%" }}
           scrolling={true}
+          heightCalculationMethod={"max"}
```
rjwignar, line 52: @mingming-ma what do you think about removing the scrollbars (set scrolling to false)?
If we use `heightCalculationMethod={"max"}` the scrollbars won't appear.

mingming-ma: Maybe we can remove `scrolling={true}` since `heightCalculationMethod={"max"}` the scrollbars won't appear. So there would be less conflicts of the two settings I guess.

rjwignar: Noted. I recently removed `scrolling={true}`.

WangGithub0:
Very good explanation.  
If user inputs a very long HTML page, should we preview all the HTML page or provide a scrollbar if the height is bigger than a value?

rjwignar:
> Very good explanation. If user inputs a very long HTML page, should we preview all the HTML page or provide a scrollbar if the height is bigger than a value?

If we provide a scrollbar after a maxHeight threshold, we risk the last line of text being cutoff at that maxHeight.

Currently, a **user's** long messages are treated with the "Show More/Show Less" buttons, see [here](https://chatcraft.org/api/share/rjwignar/v_TLXUGE-f5PtJaezpoLt) for an example.

It looks like very long HTML code from **ChatCraft's responses** is **already** given this treatment, as seen [here](https://chatcraft.org/api/share/rjwignar/bRw2wYMh5UvOc5J7UDen_):
image
, based on the 5000 character threshold in MessageBase.tsx
https://github.com/tarasglek/chatcraft.org/blob/a1ab66515a39e06684cc455a36263c67c95f5f33/src/components/Message/MessageBase.tsx#L117

humphd:
I like the "Show More..." UI taking over for really long content.  Let's not do a scrollbar.

WangGithub0:
@rjwignar  Your screenshot "Show more" is perfect, but my test show all the html code and preview page. Maybe it's because the HTML page is long, but it hasn't exceeded the word limit we set.
https://cc8cd9a7.console-overthinker-dev.pages.dev/c/chatcraft_dev/Lg8_sJvyWuPWwrAcPi2Kl

humphd:
@WangGithub0 we added a fix to not hide long messages when they are the last message.  The idea being that whatever you just asked the LLM about, you are going to want to see it.  If you add another message to that chat, it should shrink the long HTML+Preview.

WangGithub0:
> @WangGithub0 we added a fix to not hide long messages when they are the last message. The idea being that whatever you just asked the LLM about, you are going to want to see it. If you add another message to that chat, it should shrink the long HTML+Preview.

Yes, it make sense. After adding a new message, it still shows all the html code and preview page:
https://cc8cd9a7.console-overthinker-dev.pages.dev/c/chatcraft_dev/awjajAaEyfB4KP6jFz4Pb

humphd:
cc @mingming-ma, what do you think about @WangGithub0's findings in https://cc8cd9a7.console-overthinker-dev.pages.dev/c/chatcraft_dev/awjajAaEyfB4KP6jFz4Pb (e.g., it's not the last message, and not using Show More...)?

mingming-ma:
> Yes, it make sense. But after adding a new message, it still shows all the html code and preview page:
https://cc8cd9a7.console-overthinker-dev.pages.dev/c/chatcraft_dev/awjajAaEyfB4KP6jFz4Pb
should we also consider about the height, not only the word number for html?

> cc @mingming-ma, what do you think about @WangGithub0's findings in https://cc8cd9a7.console-overthinker-dev.pages.dev/c/chatcraft_dev/awjajAaEyfB4KP6jFz4Pb (e.g., it's not the last message, and not using Show More...)?

After check, here is my findings: `it's not the last message, and not using Show More`, This is only happening at the first time of this chat. i.e. an interactive way messaging on and on, user will see the previous last massage still expanded. I think it might be good to let user see the last message unchanged for not being distracted.

And when user open the chat again from `Saved Chat` or get it from a shared link, all long message is folded except the last one.

How do you think about this behaviour? Should I do always collapse the reviewed message? 

@WangGithub0 's  example does not show the button. This is because the message's `height` is huge, however the characters are still within the threshold. It depends on how LLM response, for the height checking , maybe continue checking the '\n' including in long message would impact the performance. Let me know if there are better way to fix this situation.

WangGithub0 (approved):
looks good

humphd (approved):

PR #412, opened 2024-02-06
Fix: Don't use Show More... for final message 

### LINKED ISSUE #394: Don't use Show More... for final message
After the LLM responds with a long message, we are now shrinking it and only showing the first few lines.  I think this is wrong when it's the last response.  I want to read the whole thing, and not have to expand it to see what was said.

mingming-ma:
Fixes #394 

## Explanation 

Currently all long messages are collapsed, (`isOpen` is `false` by default)

This update adds a `useEffect` to detect message status, if it is the final message and not system message, it toggles the `Show More...` button for users to have a better view.

~~The `onDeleteAfterClick` property is `undefined` when it is the final message and so can be used for the check.~~  updated to `hasMessagesAfter` 

[Result](https://mingming-longmessages-394.console-overthinker-dev.pages.dev/c/chatcraft_dev/xRoHj8zi1inmiPZOykJQi)

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -47,6 +47,7 @@ import {
   ChatCraftAiMessage,
   ChatCraftAiMessageVersion,
   ChatCraftMessage,
+  ChatCraftSystemMessage,
 } from "../../lib/ChatCraftMessage";
 import { ChatCraftModel } from "../../lib/ChatCraftModel";
 import { useModels } from "../../hooks/use-models";
@@ -127,6 +128,14 @@ function MessageBase({
       message.tokens().then(setTokens).catch(console.warn);
     }
   }, [settings.countTokens, message]);
+
+  // If last message is collapsed, auto expand for better view
+  useEffect(() => {
+    if (!onDeleteAfterClick && !isOpen && !(message instanceof ChatCraftSystemMessage)) {
+      onToggle();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   const handleCopy = useCallback(() => {
     onCopy();
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -134,8 +134,11 @@ function MessageBase({
     if (!onDeleteAfterClick && !isOpen && !(message instanceof ChatCraftSystemMessage)) {
       onToggle();
     }
+
+    // ignore isOpen as onToggle() will change isOpen status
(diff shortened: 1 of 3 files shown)


src/components/Message/MessageBase.tsx
```diff
@@ -128,6 +129,14 @@ function MessageBase({
     }
   }, [settings.countTokens, message]);
 
+  // If last message is collapsed, auto expand for better view
+  useEffect(() => {
+    if (!onDeleteAfterClick && !isOpen && !(message instanceof ChatCraftSystemMessage)) {
+      onToggle();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
humphd, line 138: Why not include your dependencies here?  `onDeleteAfterClick`, `isOpen`, etc?

mingming-ma: When I write that code I think we can run the useEffect only once as there is only one final message. 

After double checking the dependencies here, I find we should not include `isOpen` in the dependencies because the `onToggle` will change the `isOpen` status and so will call `onToggle` again, so user could not collapse the final message. 

`onDeleteAfterClick ` can be added as user may delete the final message, and the message on top of that if it's a collapsed one, it will then auto expanded. How do you think about this behaviour?

humphd: Sounds good, let's try it.

src/components/Message/MessageBase.tsx
```diff
@@ -128,6 +129,17 @@ function MessageBase({
...
+  // If last message is collapsed, auto expand for better view
+  useEffect(() => {
+    if (!onDeleteAfterClick && !isOpen && !(message instanceof ChatCraftSystemMessage)) {
+      onToggle();
+    }
+
+    // ignore isOpen as onToggle() will change isOpen status
+    // ignore message as each message has its corresponding onDeleteAfterClick
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [onDeleteAfterClick]);
```
humphd, line 141: I don't love using `onDeleteAfterClick` to mean "is last message," since we actually have this info:

https://github.com/tarasglek/chatcraft.org/blob/ee9e42b936c5e503f9650f59fac82d125890b874/src/components/MessagesView.tsx#L92

Let's add a new prop to `MessageBase` that lets you know if this is the lastMessage or not, based on that.  Then use it here.

humphd (changes requested):

humphd (approved):
I love how you always include example, shared chats.  This is great.  Thanks for doing a proper fix.

PR #414, opened 2024-02-07
Allow users to authenticate with Google OAuth - step1 get google info

WangGithub0:
Add google login according to [OAuth 2.0 for Client-side Web Applications](https://developers.google.com/identity/protocols/oauth2/javascript-implicit-flow)

Step 1: Redirect to Google's OAuth 2.0 server
Step 2: Google prompts user for consent
Step 3: Handle the OAuth 2.0 server response - get access_token (google will redirect_uri to "/" with accessToken)
Step 4: Calling Google APIs - get user info using access_token

I provide my google client_id, which can be used in testing on http://localhost:5173/ ,  and uncomment line142-line152 in `src/components/Header.tsx` to test.
image

@humphd I have some questions need your help:
1. is it right to use this "OAuth 2.0 for Client-side Web Applications" for our pure browser-based web app?
2. is there a better way to get access_token? I'm not sure is it good to add this part in [router.tsx](https://github.com/tarasglek/chatcraft.org/compare/issue-330?expand=1#diff-2194a569c1e2cb2d5e6be82f85a9b8aa671cb1178cec3dabf1910af929285cb2)?
3. Should I save access_token in hook or cookie?  Then I can separate the step3 and step4.

### DIFF
--- a/functions/google.ts
+++ b/functions/google.ts
@@ -1,0 +1,72 @@
+// https://developers.google.com/identity/protocols/oauth2/javascript-implicit-flow#redirecting
+export async function oauthSignIn() {
+  // export async function oauthSignIn(CLIENT_ID: string,REDIRECT_URI: string) {
+
+  // Google's OAuth 2.0 endpoint for requesting an access token
+  const oauth2Endpoint = "https://accounts.google.com/o/oauth2/v2/auth";
+
+  // Create <form> element to submit parameters to OAuth 2.0 endpoint.
+  const form = document.createElement("form");
+  form.setAttribute("method", "GET"); // Send as a GET request.
+  form.setAttribute("action", oauth2Endpoint);
+
+  // Parameters to pass to OAuth 2.0 endpoint.
+  const params = {
+    client_id: "70478082635-iaa28pt6bg1h06ooeic3vo8fgtu90trh.apps.googleusercontent.com", // This is my test accout, can be used in testing on http://localhost:5173/
+    redirect_uri: "http://localhost:5173",
+    response_type: "token",
+    scope: "profile email",
+    include_granted_scopes: "true",
+    state: "pass-through value",
+  };
+
+  // Add form parameters as hidden input values.
+  for (const p in params) {
+    const input = document.createElement("input");
+    input.setAttribute("type", "hidden");
+    input.setAttribute("name", p);
+    input.setAttribute("value", params[p]);
+    form.appendChild(input);
+  }
+
+  // Add form to page and submit it to open the OAuth 2.0 endpoint.
+  document.body.appendChild(form);
+  form.submit();
+}
+
+export async function requestUserInfo(accessToken: string) {
+  const xhr = new XMLHttpRequest();
+  const url = `https://www.googleapis.com/drive/v3/about?fields=user&access_token=${accessToken}`;
+  xhr.open("Get", url);
+
+  xhr.onreadystatechange = function () {
+    if (xhr.readyState === 4 && xhr.status === 200) {
+      const response = JSON.parse(xhr.responseText);
+      console.log(response);
+      const { login, name, avatar_url } = response;
+      const userInfo = {
+        username: login,
+        name: name ?? login,
+        avatarUrl: avatar_url,
+      };
+      console.log(userInfo);
+    } else if (xhr.readyState === 4 && xhr.status === 401) {
+      // Token invalid, so prompt for user permission.
+      oauthSignIn();
+    }
+  };
+
+  // xhr.onreadystatechange = function () {
+  //   console.log(xhr.response);
+  // };
+  xhr.send(null);
+
+  // this will be used later to create user
+  // const { login, name, avatar_url } = (await res.json()) as {
+  //   login: string;
+  //   name: string | null;
+  //   avatar_url: string;
+  // };
+
+  // return { username: login, name: name ?? login, avatarUrl: avatar_url };
+}
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -20,12 +20,14 @@ import {
 } from "@chakra-ui/react";
 import { BiSun, BiMoon, BiMenu } from "react-icons/bi";
 import { BsGithub } from "react-icons/bs";
+import { FcGoogle } from "react-icons/fc";
 import { TbSearch } from "react-icons/tb";
 import { Form } from "react-router-dom";
 
 import PreferencesModal from "./PreferencesModal";
 import DefaultSystemPromptModal from "./DefaultSystemPromptModal";
 import { useUser } from "../hooks/use-user";
+import { oauthSignIn } from "../../functions/google";
 
 type HeaderProps = {
   chatId?: string;
@@ -136,6 +138,17 @@ function Header({ chatId, inputPromptRef, searchText, onToggleSidebar }: HeaderP
                   </>
                 )}
               </MenuItem>
+              {/* Google login */}
+              {/* <MenuItem
+                onClick={async () => {
+                  oauthSignIn();
+                  console.log("google");
+                }}
+              >
+                <>
+                  <FcGoogle /> <Text ml={2}>Sign in with Google</Text>
+                </>
+              </MenuItem> */}
               <MenuDivider />
               <MenuItem
                 as="a"
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -17,6 +17,25 @@ export default createBrowserRouter([
   {
     path: "/",
     async loader() {
+      // Because google will redirect_uri to "/" with accessToken
+      const requestedURI = location.pathname;
+      // Accessing anchor values
+      const anchor = location.hash.substring(1); // Exclude the '#' character
+      console.log("Requested URI:", requestedURI);
+      console.log("Anchor:", anchor);
+      const match = anchor.match(/access_token=([^&]*)/);
+      const accessToken = match ? match[1] : null;
+      console.log("accessToken:", accessToken);
+      const xhr = new XMLHttpRequest();
+      xhr.open(
+        "GET",
+        "https://www.googleapis.com/drive/v3/about?fields=user&" + "access_token=" + accessToken
+      );
+      xhr.onreadystatechange = function (e) {
+        console.log(xhr.response);
+      };
+      xhr.send(null);
+
       try {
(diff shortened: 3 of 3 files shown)


functions/google.ts
```diff
@@ -0,0 +1,72 @@
...
+  const oauth2Endpoint = "https://accounts.google.com/o/oauth2/v2/auth";
+
+  // Create <form> element to submit parameters to OAuth 2.0 endpoint.
+  const form = document.createElement("form");
+  form.setAttribute("method", "GET"); // Send as a GET request.
+  form.setAttribute("action", oauth2Endpoint);
+
+  // Parameters to pass to OAuth 2.0 endpoint.
+  const params: { [key: string]: string } = {
+    client_id: "70478082635-iaa28pt6bg1h06ooeic3vo8fgtu90trh.apps.googleusercontent.com", // This is my test accout, can be used in testing on http://localhost:5173/
```
humphd, line 15: This should probably come from an ENV variable that we set in CloudFlare vs. hard coding it here.

functions/google.ts
```diff
@@ -0,0 +1,72 @@
...
+    redirect_uri: "http://localhost:5173",
```
humphd, line 16: For testing I understand this, but it won't work for prod.

functions/google.ts
```diff
@@ -0,0 +1,72 @@
...
+  for (const p in params) {
+    const input = document.createElement("input");
+    input.setAttribute("type", "hidden");
+    input.setAttribute("name", p);
+    input.setAttribute("value", params[p]);
+    form.appendChild(input);
+  }
+
+  // Add form to page and submit it to open the OAuth 2.0 endpoint.
+  document.body.appendChild(form);
```
humphd, line 33: We won't have a `document` in CloudFlare.  The JS you run in a CloudFlare Worker Function is a lot like running node.js.

functions/google.ts
```diff
@@ -0,0 +1,72 @@
+// https://developers.google.com/identity/protocols/oauth2/javascript-implicit-flow#redirecting
+export async function oauthSignIn() {
+  // export async function oauthSignIn(CLIENT_ID: string,REDIRECT_URI: string) {
+
+  // Google's OAuth 2.0 endpoint for requesting an access token
+  const oauth2Endpoint = "https://accounts.google.com/o/oauth2/v2/auth";
+
+  // Create <form> element to submit parameters to OAuth 2.0 endpoint.
+  const form = document.createElement("form");
```
humphd, line 9: There's no document in CloudFlare, but you can use https://developer.mozilla.org/en-US/docs/Web/API/FormData if you need to work with `FormData`.

humphd (changes requested):
We should probably set a time to meet and go over this approach.  Doing this in CloudFlare will require some alternate approaches.

I also notice that other people have solved this, for example: https://github.com/Schachte/cloudflare-google-auth and the blog post at https://ryan-schachte.com/blog/oauth_cloudflare_workers/

src/components/Header.tsx
```diff
@@ -127,15 +131,35 @@ function Header({ chatId, inputPromptRef, searchText, onToggleSidebar }: HeaderP
             <MenuList>
               <MenuItem onClick={onPrefModalOpen}>Settings...</MenuItem>
               <MenuItem onClick={onSysPromptModalOpen}>Default System Prompt...</MenuItem>
-              <MenuItem onClick={handleLoginLogout}>
-                {user ? (
-                  "Logout"
-                ) : (
-                  <>
+              {user && (
```
humphd, line 134: Combine both cases into a single expression

```tsx
{ user ? (
  <MenuItem ...>
) : (
  <MenuItmem ...>
) }
```

src/router.tsx
```diff
@@ -25,7 +25,6 @@ export default createBrowserRouter([
       } catch (err) {
         console.warn("Error getting most recent chat", err);
       }
-
```
humphd, line 28: Avoid changing files that aren't part of your feature/bug so you don't alter the git history unnecessarily.

humphd (changes requested):
This is an amazing piece of work, well done @WangGithub0!  I'm really excited to see this land for 1.3.

I've left some feedback.  You can fix this now, or in follow-ups, but if you do it in follow-ups, please file issues.

functions/api/login.ts
```diff
@@ -101,24 +112,173 @@ export async function handleDevLogin({
...
+    GOOGLE_CLIENT_ID,
+    GOOGLE_CLIENT_SECRET,
+    GOOGLE_REDIRECT_URI,
+    GOOGLE_RESPONSE_TYPE,
+    GOOGLE_SCOPE,
+  } = env;
+
   const reqUrl = new URL(request.url);
+  // Determine the login provider
+  provider = reqUrl.searchParams.get("provider") ? reqUrl.searchParams.get("provider") : provider;
```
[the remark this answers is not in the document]
WangGithub0: I have to save `provider` out of this function, because only the first time it sends the `provider` from `use-user.tsx`, after coming back from google, it doesn't have `provider` any more.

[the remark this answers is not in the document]
humphd: What if one person logs in with Google, another with GitHub, at the same moment?  The order of these will be different, no?

I would use `reqUrl.origin` to figure out which provider you are dealing with.

humphd:
@WangGithub0 please click "Resolve" for each of the review comments above that you've fixed, to indicate that they are no longer an issue we need to track/discuss.

functions/api/login.ts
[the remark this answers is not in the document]
WangGithub0: > What if one person logs in with Google, another with GitHub, at the same moment? The order of these will be different, no?
> 
> I would use `reqUrl.origin` to figure out which provider you are dealing with.

I found "The origin property of a URL refers to the scheme (protocol), hostname, and port number. It does not include the path, query parameters, or fragment. Therefore, it cannot be used to store or retrieve the "provider" information." from GPT4, could you tell me how to use reqUrl.origin to get provider?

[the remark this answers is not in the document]
humphd: Correct me if I'm wrong, but don't you just need to know if this is "google" or "github"?  If so, when you get redirected back from Google's or GitHub's auth page with your code, you will be able to look at the origin or hostname, and figure out where you came from.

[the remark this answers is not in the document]
WangGithub0: Great idea! I was struggling before on how to get the provider. Thanks for pointing that out.

functions/api/login.ts
```diff
@@ -28,7 +27,11 @@ export const onRequestGet: PagesFunction<Env> = async ({ request, env }) => {
   const reqUrl = new URL(request.url);
 
   // Determine the login provider
-  provider = reqUrl.searchParams.get("provider") ? reqUrl.searchParams.get("provider") : provider;
+  const provider = reqUrl.searchParams.get("provider")
+    ? reqUrl.searchParams.get("provider")
+    : reqUrl.origin.includes("github")
```
humphd, line 32: We might add more providers later.  Make this a function:

```ts
function getProvider(url: URL) {
  const { hostname } = url;
  if(hostname.includes('github')) {
    return 'github';
  }
  if(hostname.includes('google')) {
    return 'google';
  }
  return null;
}
```

humphd (changes requested):

functions/api/login.ts
WangGithub0: add a provider in google/github state

Amnish04 (commented):
I tried logging in with google on Cloudfare deployed url and was a success, but I guess it was a dev login.
image

I've also gone through the code, and even though I don't fully each and everything, I've gained a basic understanding of the flow and how we're managing different auth providers.

As @humphd said, this is a great piece of work! I can understand the pain of implementing Auth since I recently had to do it for the capstone project.

I can't suggest any big changes at this point as lots of improvements have already been made going through the conversations, but I've suggested some minor changes and commented some questions for my clarity.

functions/api/login.ts
```diff
@@ -1,124 +1,82 @@
...
-    return Response.redirect(`https://chatcraft.org/?github_login_error`, 302);
+  // Determine the login provider
+  let provider = reqUrl.searchParams.get("provider");
+  if (!provider) {
+    let state = reqUrl.searchParams.get("state");
+    if (state) {
+      state = decodeURIComponent(state);
+      const stateParams = new URLSearchParams(state);
+      provider = stateParams.get("provider");
+    }
```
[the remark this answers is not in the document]
WangGithub0: yes, when redirect from google/github, we have to get provider information to decide next step run  handleGoogleLogin() or handleGithubLogin().

WangGithub0:
> I tried logging in with google on Cloudfare deployed url and was a success, but I guess it was a dev login. image
> 
> I've also gone through the code, and even though I don't fully understand each and everything, I've gained a basic understanding of the flow and how we're managing different auth providers.
> 
> As @humphd said, this is a great piece of work! I can understand the pain of implementing Auth since I recently had to do it for the capstone project.
> 
> I can't suggest any big changes at this point as lots of improvements have already been made going through the conversations, but I've suggested some minor changes and commented some questions for my clarity.

Yes, it's dev. You can set isDev:false at line58 in `login.ts` to test product environment
if (provider === "google") {
    return handleGoogleLogin({
      isDev: true,

humphd (approved):

Amnish04 (approved):

humphd:
@WangGithub0 can you "Squash and merge" this when you have a minute?  I'd love to see this land in 1.3!  Such a cool new feature.

WangGithub0:
> @WangGithub0 can you "Squash and merge" this when you have a minute? I'd love to see this land in 1.3! Such a cool new feature.

Do we have a `.dev.ars` for product environment? could you add ChatCraft [`GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`](https://github.com/tarasglek/chatcraft.org/pull/380) in it?

humphd:
> > @WangGithub0 can you "Squash and merge" this when you have a minute? I'd love to see this land in 1.3! Such a cool new feature.
> 
> Do we have a `.dev.vars` for product environment? could you add ChatCraft [`GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`](https://github.com/tarasglek/chatcraft.org/pull/380) in it?

I've updated the CloudFlare environment variables for production.

PR #419, opened 2024-02-07
Fix image styling when using gpt4-vision

### LINKED ISSUE #418: Follow Up issues from PR#286
Image width (either make it 100% or responsive or centered)
Preview size of the image while clicking whether limit it to the size of original image

https://github.com/tarasglek/chatcraft.org/pull/286#issuecomment-1932439118

Amnish04:
For PR #286

Improved the styling of images in **messages** and **modal preview**.

**Before:**

image

image

***

**After:**

image

image

***

Smaller images will only take as much space as they need.

If the image can't fit, there will be scrollbars to preserve aspect ratio

This fixes #418

### DIFF
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -23,9 +23,15 @@ const ImageModal: React.FC<ImageModalProps> = ({ isOpen, onClose, imageSrc }) =>
       <ModalCloseButton />
       <ModalBody>
         <Flex height={"100%"} justifyContent={"center"} alignItems={"center"}>
-          <Box maxWidth="100%" maxHeight="70vh" overflow={"auto"}>
-            <Image src={imageSrc} alt="Selected Image" m="auto" objectFit="contain" />
-          </Box>
+          <Image
+            maxWidth="100%"
+            maxHeight="70vh"
+            overflow={"auto"}
+            src={imageSrc}
+            alt="Selected Image"
+            m="auto"
+            objectFit="contain"
+          />
         </Flex>
       </ModalBody>
     </ModalContent>
--- a/src/components/ImageModal.tsx
+++ b/src/components/ImageModal.tsx
@@ -6,6 +6,8 @@ import {
   ModalCloseButton,
   ModalBody,
   Image,
+  Box,
+  Flex,
 } from "@chakra-ui/react";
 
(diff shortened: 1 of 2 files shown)


humphd (approved):
This looks good to me, nice work!

@kliu57, can you confirm that this solves your issues?

mingming-ma (approved):
@Amnish04 That was so fast fix. Thanks a lot!

Amnish04:
> @Amnish04 That was so fast fix. Thanks a lot!

You're welcome :D

kliu57 (commented):
- I like the fact that the image now shows the original sized image in the centered in the chat
- In the image preview popup, when the image width and height both exceed the popup window, we get a vertical scrollbar but no horizontal scrollbar (we resized the image but not enough to fit it in the window) In this case I think mingming's code worked better for these super large images, since he fit the entire image in the window.

Amnish's code (when it comes to super large img):

image

Mingming's code (when it comes to super larg img):

image

- When it comes to small images, Amnish's image preview is better since it allows those small images to stay their original size (Mingming's code made small images extremely large to fit the entire window). I'm not really liking the empty white space in the popup window though...

image

Amnish04:
@kliu57 Thanks for showing me the issues on call. 
I just pushed a commit to fix that scrollbar issue and better scaling in modal. Can you please recheck?

For the whitespace, that's because I am fixing the size of modal for consistency but we can do whatever everyone prefers.

kliu57:
> @kliu57 Thanks for showing me the issues on call. I just pushed a commit to fix that scrollbar issue and better scaling in modal. Can you please recheck?
> 
> For the whitespace, that's because I am fixing the size of modal for consistency but we can do whatever everyone prefers.



> @kliu57 Thanks for showing me the issues on call. I just pushed a commit to fix that scrollbar issue and better scaling in modal. Can you please recheck?
> 
> For the whitespace, that's because I am fixing the size of modal for consistency but we can do whatever everyone prefers.

I re-tested. It looks good now. There is no more scrollbar and it behaves similar to mingming's code.

My only concern now would be the white space for small images. I feel like that could be an enhancement for later releases though.

mingming-ma:
Thanks @kliu57 for the review, let me merge this.

PR #420, opened 2024-02-07
Query available models to determine TTS support

### LINKED ISSUE #387: Use a better approach for determining TTS Support
Follow up for #357 

Currently, the method to determine if **TTS** is supported is **hardcoded** to check if we are using OpenAI.
image

We should rather query the **list of models** to determine this as mentioned in the following comments:
298836122-5439c726-c888-4c63-bbe2-c3ae256a93f0
https://github.com/tarasglek/chatcraft.org/pull/357#issuecomment-1905443626
https://github.com/tarasglek/chatcraft.org/pull/357#issuecomment-1913156997

Amnish04:
I am now querying the available models to determine if tts is supported rather than just checking if we are using openai.

For this, I have modified the `queryModels` function to accept a custom predicate for filtering purposes. I have added the previously used predicate as default value so other placed don't break.
```ts
filterPredicate: (model: ChatCraftModel) => boolean = (model) =>
    !usingOfficialOpenAI() || model.id.includes("gpt")
```

and `isTtsSupported` now looks like this

```ts
export async function isTtsSupported() {
  const { apiKey } = getSettings();
  if (!apiKey) {
    throw new Error("Missing API Key");
  }

  return (
    (
      await queryModels(apiKey, (model: ChatCraftModel) => {
        return model.id.includes("tts");
      })
    )?.length > 0
  );
}
```

This closes #387

### DIFF
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -15,9 +15,7 @@ import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
 import { MdVolumeUp, MdVolumeOff } from "react-icons/md";
-import { isTtsSupported, usingOfficialOpenAI } from "../../lib/ai";
-import { useEffect, useState } from "react";
-import { useAlert } from "../../hooks/use-alert";
+import { usingOfficialOpenAI } from "../../lib/ai";
 
 type PromptSendButtonProps = {
   isLoading: boolean;
@@ -27,24 +25,9 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
 
-  const { error } = useAlert();
-
-  const [isTTSSupported, setIsTTSSupported] = useState(false);
-
-  useEffect(() => {
-    async function checkTtsSupport() {
-      try {
-        const supported = await isTtsSupported();
-        setIsTTSSupported(supported);
-      } catch (err: any) {
-        error({
-          title: "Error checking TTS support",
-          message: err.message,
-        });
-      }
-    }
-    checkTtsSupport();
-  }, [error]);
+  function isTtsSupported() {
+    return !!models.filter((model) => model.id.includes("tts"))?.length;
+  }
 
   return (
     <ButtonGroup variant="outline" isAttached>
@@ -58,7 +41,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
           isLoading={isLoading}
           icon={<TbSend />}
         />
-        {isTTSSupported && (
+        {isTtsSupported() && (
           <Tooltip
             label={settings.announceMessages ? "Text-to-Speech Enabled" : "Text-to-Speech Disabled"}
           >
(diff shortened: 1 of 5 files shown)


Amnish04:
> lgtm...though I think .filter applied to results is always better than having every func support filter functionality, but pre-existing code already set the usage pattern

Yes, do you mean we shouldn't filter inside the function?

src/lib/ai.ts
```diff
@@ -353,7 +362,7 @@ export async function queryModels(apiKey: string) {
     }
 
     return models
-      .filter((model: any) => !usingOpenAI || model.id.includes("gpt"))
+      .filter((model: any) => filterPredicate(model))
```
humphd, line 365: This is overly clever.  I'd prefer to see you filter the results on line 462 below.  You can do stuff like this, but you risk creating code that is harder to maintain.

humphd (changes requested):

tarasglek:
> > lgtm...though I think .filter applied to results is always better than having every func support filter functionality, but pre-existing code already set the usage pattern
> 
> Yes, do you mean we shouldn't filter inside the function?

yes. queryModels().filter() is much better than queryModels(filterCallback) as .filter is standard JS

Amnish04:
@humphd @tarasglek Agreed, I also don't like that style.

Just removed default filtering from `queryModels` so all available models are returned (from `useModels` as well). Any model filtering is done by consumers now.

Is that what we were expecting?

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -25,6 +27,25 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
 
+  const { error } = useAlert();
+
+  const [isTTSSupported, setIsTTSSupported] = useState(false);
```
humphd, line 32: `isTtsSupported` and `isTTSSupported` is not a good idea.

Also, can you not use the `models` from the `useModels()` hook in here?  Ideally you should use our hook for getting that info.

humphd (changes requested):

src/lib/ai.ts
```diff
@@ -341,9 +341,14 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
   };
 };
 
+/**
+ *
+ * @param apiKey Api key for the provider
+ * @param filterPredicate A function used to filter the array of returned models
```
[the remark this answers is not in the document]
Amnish04: Right I forgot about that. Just removed

src/components/PromptForm/PromptSendButton.tsx
Amnish04: I wasn't using that initially because it was always filtering to gpt models. But now we can definitely use it. Thanks for pointing out!

Amnish04:
@humphd Just pushed the requested changes. I am using the `useModels` hook in components now.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -25,6 +25,10 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
 
+  function isTtsSupported() {
```
humphd, line 28:
```tsx
const isTtsSupported = useMemo(() =>
  models.filter((model) => model.id.includes("tts"))?.length > 0,
  [models]
);
```

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -82,6 +88,10 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
...
```
humphd, line 91: Same here, you should be able to cache this and only re-calculate on models changing.

humphd (approved):
Looks good.  A few optimizations you can make.

src/components/PromptForm/PromptSendButton.tsx
Amnish04: That's right! Just pushed the changes.

humphd (approved):
Nice

Amnish04:
Thanks for helping. Are we good to land this?

humphd:
Check in with the sheriff.

Amnish04:
@Rachit1313 Do you think we need more reviews here?

humphd:
@Rachit1313 ping, what's up with this one?  Can we merge?

PR #421, opened 2024-02-08
Allow store and toggle of multiple providers info

### LINKED ISSUE #345: Allow multiple providers to be used simultaniously

kliu57:
Closes #345 

### Summary:
Api keys for multiple provider are stored in settings.providers so user can toggle between providers without re-entering api key.

image

See previously merged PR https://github.com/tarasglek/chatcraft.org/pull/371 (set up done to make this possible)

### Code changes in this PR:

`src/components/PreferencesModal.tsx`: Modified code of the event handler of the `API URL` dropdown field, so that when users toggle the dropdown:

- we save the old provider key to settings.providers (if it is not there already and if the key is not blank) (this logic is necessary because for old users, their current api info is saved to settings.apiUrl and settings.apiKey but not yet saved to settings.providers)
- then, we check if we have any stored provider key for the newly selected provider in settings.providers and load it if we do

### Ways to test

**New user:** Clear browser cache, enter key, go to Settings, toggle to a different provider, enter another key, toggle back and forth

**Old user:** Copy the following string 

```
{"model":"gpt-3.5-turbo","apiUrl":"https://api.openai.com/v1","temperature":0,"enterBehaviour":"send","countTokens":false,"sidebarVisible":false,"alwaysSendFunctionResult":false,"announceMessages":false,"apiKey":"YOUR_API_KEY"}
```

, replace YOUR_API_KEY with a valid openai key, and paste it in the localStorage, settings field. This will mimic you being a user of ChatCraft that has not logged on since prior to release 1.1. We can use this to test the following:

- we save the old provider key to settings.providers (if it is not there already and if the key is not blank)

Note: the reason we cannot use a new user to test this is because for new users, your provider will be in settings.providers already.

---
Made a separate issue #423 for allowing the API URL dropdown to read from a list of ChatCraftProvider objects

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -153,6 +153,39 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
     [inputRef]
   );
 
+  const handleProviderChange = useCallback(
+    (e: ChangeEvent<HTMLSelectElement>) => {
+      const currentProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
+      const selectedProvider = ChatCraftProvider.fromUrl(e.target.value);
+
+      // If new provider exists in settings.providers, parse it to ChatCraftProvider
+      const newProvider = settings.providers[selectedProvider.name]
+        ? ChatCraftProvider.fromJSON(settings.providers[selectedProvider.name])
+        : selectedProvider;
+
+      // Store old provider in settings.providers array if not already there
+      // Set current apiKey to value loaded from settings.providers or undefined
+      if (currentProvider.apiKey && !(currentProvider.name in settings.providers)) {
+        setSettings({
+          ...settings,
+          apiUrl: newProvider.apiUrl,
+          apiKey: newProvider.apiKey,
+          providers: {
+            ...settings.providers,
+            [currentProvider.name]: currentProvider,
+          },
+        });
+      } else {
+        setSettings({
+          ...settings,
+          apiUrl: newProvider.apiUrl,
+          apiKey: newProvider.apiKey,
+        });
+      }
+    },
+    [settings, setSettings]
+  );
+
   return (
     <Modal isOpen={isOpen} onClose={onClose} size="lg" finalFocusRef={finalFocusRef}>
       <ModalOverlay />
@@ -163,17 +196,7 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
           <VStack gap={4}>
             <FormControl>
               <FormLabel>API URL</FormLabel>
-              <Select
-                value={settings.apiUrl}
-                onChange={(e) => {
-                  setSettings({
-                    ...settings,
-                    apiUrl: e.target.value,
-                    apiKey: undefined,
-                    providers: {},
-                  });
-                }}
-              >
+              <Select value={settings.apiUrl} onChange={handleProviderChange}>
                 <option value={OPENAI_API_URL}>OpenAI ({OPENAI_API_URL})</option>
                 <option value={OPENROUTER_API_URL}>OpenRouter.ai ({OPENROUTER_API_URL})</option>
               </Select>
@@ -227,7 +250,7 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
 
                   setSettings({
                     ...settings,
-                    apiKey: e.target.value,
+                    apiKey: newProvider.apiKey,
                     providers: {
                       ...settings.providers,
                       [newProvider.name]: newProvider,


src/components/PreferencesModal.tsx
```diff
@@ -166,12 +166,32 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <Select
                 value={settings.apiUrl}
                 onChange={(e) => {
-                  setSettings({
-                    ...settings,
-                    apiUrl: e.target.value,
-                    apiKey: undefined,
-                    providers: {},
-                  });
+                  const oldProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
```
humphd, line 169: Not sure if it's just noise from you doing development locally, but FYI, the new Clarity metrics we added are showing `ChatCraftProvider.fromUrl is not a function`:

Screenshot 2024-02-08 at 9 28 36 AM

src/components/PreferencesModal.tsx
```diff
@@ -166,12 +166,32 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
+                  let newProvider = ChatCraftProvider.fromUrl(e.target.value, undefined);
```
humphd, line 170: Instead of passing `undefined` here, can you pass a single argument, and have the function deal with that in its type?

src/components/PreferencesModal.tsx
```diff
@@ -166,12 +166,32 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
+
+                  // Check if we have saved key for the new provider
+                  if (newProvider.name in settings.providers) {
+                    newProvider = settings.providers[newProvider.name];
```
humphd, line 174: This is not type-safe.  You are treating a `ChatCraftProvider` the same as the JSON data we get from `settings.providers[]`, which won't be the case (e.g., you won't have methods).

A better approach here is to take the data in `settings.providers[newProvider.name]` and use it to call the `ChatCraftProvider` constructor or a `.fromJson()` method, which would return a proper type.

In general, you want to think about the database like the network, where you are working with serialized, raw JSON versions of your objects.  In the case of an `apiKey` that's fine, since it's just a `string`.  However, with complex types like a provider, you need to serialize/deserialize when moving between the database and memory.

humphd: Another change I would make it so consolidate the creation of the `newProvider` into a single line.  In general, it's a bad idea in React to have mutable data, and using `let` is often a sign that you're at risk.  Maybe the `ChatCraftProvider` class can do more of the work for us here so you don't have to create the provider, then do an if-check and modify it a few lines later.

humphd (changes requested):

src/components/PreferencesModal.tsx
kliu57: @humphd Yes I was doing some testing locally where I deleted that function. I restored the changes already.

src/components/PreferencesModal.tsx
```diff
@@ -166,12 +166,37 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
-                  });
+                  const oldProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
+                  const newProvider = ChatCraftProvider.fromUrl(e.target.value);
+
+                  // Check if the old or new provider exists in settings.providers
+                  const oldProviderExistsInProviders = oldProvider.name in settings.providers;
+                  const newProviderExistsInProviders = newProvider.name in settings.providers;
+
+                  // If new provider exists in settings.providers, parse it to ChatCraftProvider
+                  const updatedNewProvider = newProviderExistsInProviders
```
humphd, line 177: Let's reduce the number of variables we have: `newProvider`, `updatedNewProvider`.  What about something like:

```
const currentProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
const selectedProvider = ChatCraftProvider.fromUrl(e.target.value);
const newProvider = settings.providers[selectedProvider.name] ? new ChatCraftProvider(settings.providers[selectedProvider.name]) : selectedProvider;

if(currentProvider.apiKey && !(currentProvider.name in settings.providers)) {
  ...
  apiUrl: newProvider.apiUrl,
  apiKey: newProvider.apiKey,
  ...
```

I think we should use the `newProvider.apiUrl` vs. the `e.target.value` once we've created the object, so it always matches what's in that object.

humphd (changes requested):

src/components/PreferencesModal.tsx
kliu57: > Let's reduce the number of variables we have: `newProvider`, `updatedNewProvider`. What about something like:
> 
> ```
> const currentProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
> const selectedProvider = ChatCraftProvider.fromUrl(e.target.value);
> const newProvider = settings.providers[selectedProvider.name] ? new ChatCraftProvider(settings.providers[selectedProvider.name]) : selectedProvider;
> 
> if(currentProvider.apiKey && !(currentProvider.name in settings.providers)) {
>   ...
>   apiUrl: newProvider.apiUrl,
>   apiKey: newProvider.apiKey,
>   ...
> ```
> 
> I think we should use the `newProvider.apiUrl` vs. the `e.target.value` once we've created the object, so it always matches what's in that object.

Made all changes as requested. Also found another location where I could use `newProvider.apiKey` rather than `e.target.value` and made the change there as well.

src/components/PreferencesModal.tsx
```diff
@@ -166,12 +166,36 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
               <Select
                 value={settings.apiUrl}
                 onChange={(e) => {
```
Amnish04, line 168: I think it this logic is too complicated to have inside the JSX.

Maybe we should have a handler function like this

```ts
const handleProviderChanged = useCallback(
  (e: ChangeEvent<HTMLSelectElement>) => {
    const currentProvider = ChatCraftProvider.fromUrl(settings.apiUrl, settings.apiKey);
    const selectedProvider = ChatCraftProvider.fromUrl(e.target.value);

    // If new provider exists in settings.providers, parse it to ChatCraftProvider
    const newProvider = settings.providers[selectedProvider.name]
      ? ChatCraftProvider.fromJSON(settings.providers[selectedProvider.name])
      : selectedProvider;

    // Store old provider in settings.providers array if not already there
    // Set current apiKey to value loaded from settings.providers or undefined
    if (currentProvider.apiKey && !(currentProvider.name in settings.providers)) {
      setSettings({
        ...settings,
        apiUrl: newProvider.apiUrl,
        apiKey: newProvider.apiKey,
        providers: {
          ...settings.providers,
          [currentProvider.name]: currentProvider,
        },
      });
    } else {
      setSettings({
        ...settings,
        apiUrl: newProvider.apiUrl,
        apiKey: newProvider.apiKey,
      });
    }
  },
  [settings, setSettings]
);
```

and in jsx

```
<Select value={settings.apiUrl} onChange={handleProviderChanged}>
```

Amnish04 (changes requested):
I tested this locally and works as expected. All my providers and keys are preserved as I switch.

image


Just posted one suggestion for code.

humphd (approved):

kliu57:
@Amnish04 Good idea Amnish. I had not noticed that the current code already uses handler functions for some of the onChange event handling. I refectored the code to use my newly created `handleProviderChange` function.

Amnish04 (approved):

humphd (approved):

Issue #422, opened 2024-02-08
Opengraph summaries for shared chats have body-main uses aria-hidden=true


bronaugh:
As it says in the title, opengraph summaries for shared chats have body-main uses aria-hidden=true

This causes an accessibility problem for users of screen readers.

tarasglek:
this would be a good first fix to drop all attributes from main tag. For bonus points would be good to fix yaml to retain newlines when rendered in browser

PR #424, opened 2024-02-08
Add pre tag and remove all attributes from the main element to remove aria-hidden tagging

bronaugh:
This should fix issue #422.

### DIFF
--- a/src/lib/share.ts
+++ b/src/lib/share.ts
@@ -40,11 +40,7 @@ function generateSharingHTML(chat: ChatCraftChat, user: User) {
 
   mainElement.innerHTML = "";
   // Use html-escaped yaml for payload
-<<<<<<< HEAD
-  preElement = document.createElement("pre");
-=======
   const preElement = document.createElement("pre");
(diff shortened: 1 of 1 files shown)


src/lib/share.ts
```diff
@@ -40,7 +40,10 @@ function generateSharingHTML(chat: ChatCraftChat, user: User) {
 
   mainElement.innerHTML = "";
   // Use html-escaped yaml for payload
-  mainElement.textContent = chat.toYAML();
+  preElement = document.createElement("pre");
```
humphd, line 43: `const preElement = ...`

humphd (changes requested):

src/lib/share.ts
bronaugh: Thanks, fixed.

tarasglek:
Verified in prod that this works :)
image
Awesomely our prerendered sharing page even renders in [links](https://en.wikipedia.org/wiki/Links_(web_browser)) text-mode browser now...so search engines should give us more love too

PR #428, opened 2024-02-08
Update all deps for 1.2.0

humphd:
Updating our deps.  Things mostly seem OK, but I hit an [issue with types for dexie-export-import (filed upstream)](https://github.com/dexie/Dexie.js/issues/1900).


humphd:
Upstream fix for dexie-export-import now included (hurray for fast open source maintainers!)

This should get some testing before we merge, but seems OK in my (limited) testing.

rjwignar (approved):
Tested this branch and couldn't find any breaking changes.

PR #430, opened 2024-02-08
Fix html not syntax highlight

### LINKED ISSUE #429: [Regression] HTML content not being syntax-highlighted anymore
It appears that HTML isn't being highlighted anymore:

Screenshot 2024-02-08 at 1 13 15 PM

mingming-ma:
Fixes #429 

This is weird, because I couldn't find the html language from the demo list [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter). Currently I can see it is default use `text` type for the html from developer tools.

I trying to use `htmlbars` to solve this issue.

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -42,6 +42,8 @@ const fixLanguageName = (language: string | null) => {
     case "sh":
     case "shell":
       return "bash";
+    case "html":
+      return "htmlbars";
     default:
       return language;
   }
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -43,7 +43,7 @@ const fixLanguageName = (language: string | null) => {
     case "shell":
       return "bash";
     case "html":
-      return "htmlbars";
+      return "xml";
     default:
       return language;
   }


humphd (changes requested):
This doesn't make sense to me, it's worked for months.

Can we `git bisect` to see what broke it?

mingming-ma:
Yeah, sure, somehow I find this deployed still not working.

mingming-ma:
It may at least exists long time, see here the screenshots from my PR from October https://github.com/tarasglek/chatcraft.org/assets/133393905/49bb9503-69dd-4b59-a24a-249ecb4e19b7

humphd:
Amazing.  Maybe we're wrong and this never worked (cc @tarasglek)?  Looks like people use `xml` as the language for HTML content with highlight.js

mingming-ma:
It works! `xml` more suitable 👍  There is some difference between local and cloud, maybe from OS, my `htmlbars` version is also working locally, and I don't know why first try fails. Maybe we can solve by this way for now.

humphd (approved):

PR #431, opened 2024-02-09
Command button enhancements

### LINKED ISSUE #362: Improve visibility of "Share" button for chats
Spin-off from https://github.com/tarasglek/chatcraft.org/issues/355#issuecomment-1908580741

We should make it easier to share a chat.  Currently, sharing is available via the chat header menu:

Screenshot 2024-01-24 at 1 34 28 PM

This opens the Sharing modal, where users can add/edit/generate a summary, then get a sharable link:

Screenshot 2024-01-24 at 1 35 05 PM

We should [add a Share icon](https://react-icons.github.io/react-icons/search/#q=share) to the prompt area, to make this easier to discover.  It can also open the same modal (e.g., we'd have two ways to do it).

A third way we could explore is using the [Web Share API](https://web.dev/articles/web-share), where you can hook into the OS share UX (e.g., on mobile).  This way would require us to generate the sharable link first.

humphd:
Fixes #362 

@tarasglek and I were discussing today how it's too hard to find the "Share" feature.  I've done a first pass at improving our left-hand menu, which I'm now calling the "Command Button."

Screenshot 2024-02-08 at 8 41 13 PM

I've copied everything from the Chat Header menu down here, simplified the naming of what we already have, and added "Share."  The Share feature will use the [WebShare API](https://web.dev/articles/web-share) if it's supported (e.g., mobile), and fallback to our Share Modal when it's not.

I'm not sure if the "+" icon is the best one to use here, but I looked at others and nothing stood out to me as being better.

We don't need to try and land this for 1.2, since it's so late in the release cycle.


Amnish04:
@humphd Wow that's pretty cool. I didn't know that was possible :D

I tried this on my mobile, but for some reason nothing happens when I click on share.

Also for the commands button I tried [MdKeyboardCommandKey](https://react-icons.github.io/react-icons/icons/md/) icon, what do you think?
image

Also, shouldn't the button title be plural as its showing a list of commands.

humphd:
> I tried this on my mobile, but for some reason nothing happens when I click on share.

Are you logged in?  I currently disable the menu item if you aren't logged in.  Maybe there's a better way to handle it.

> Also for the commands button I tried [MdKeyboardCommandKey](https://react-icons.github.io/react-icons/icons/md/) icon, what do you think?

I thought about this too, but it's too closely associated with macOS.  This has nothing to do with what it does on Mac, so I don't think it's a good idea.

> Also, shouldn't the button title be plural as its showing a list of commands.

I don't know, maybe?  Prior art implies the singular: you use the `File` menu, not `Files`; the `Window` menu, not `Windows`.

Amnish04:
> Are you logged in? I currently disable the menu item if you aren't logged in. Maybe there's a better way to handle it.
That was it, should we have a toast suggesting to login if someone like me tries to use that button. Or some other to let the users know? But if it is temporarily disabled, I don't think its an issue.

> I thought about this too, but it's too closely associated with macOS. This has nothing to do with what it does on Mac, so I don't think it's a good idea.
> Agreed, I was also confused so wanted to know your opinion.
 
> I don't know, maybe? Prior art implies the singular: you use the `File` menu, not `Files`; the `Window` menu, not `Windows`.
I was thinking from the context "Click to see a list of commands", but I guess you're right.

tarasglek:
Command seems like an odd name. Why not options or condense down to burger icon?

tarasglek:
Likewise, share needs the standard share icon next to it

tarasglek:
What do you think of using all that empty space on the left to show a toolbar with items from Command menu?
image
Would be cool to have the share button clearly visible...Could also move the toolbar into menu area next to user avatar

humphd:
Made some changes:

Screenshot 2024-02-10 at 1 43 33 PM

- Renamed "Command" to "Options."  I don't think it works to only have an icon, but I like "Options," as this is exactly what it is.
- Added icons to menu items that use them elsewhere. I not sure if it works or not, see what you think.
- Renamed "Delete" to "Delete Chat," which is what it does
- Renamed "Open New" to "New Window"
- Improved menu grouping a bit
- Moved "Attach Files..." into this menu vs. having in prompt.  I don't think it's worth having it take space in the prompt area for normal usage, since it won't be the default thing you do.

I'm not sure about toolbars.  Let's do that as a separate thing.

Amnish04:
Looks better now!

Shouldn't we also add an icon to Download option as well?
Looks a little inconsistent with alignment. Maybe within a 'group' of options, either all items should have icons or none?

humphd:
Added a Download icon to fix alignment in menu group, per review.

humphd:
Screenshot 2024-02-10 at 6 00 49 PM

Amnish04:
Looks good now. Thanks!

Amnish04 (approved):

Amnish04:
I think it would be cool to also allow users to copy in 3 different format like with **Download** - `text`, `markdown` and `image`.

Should we have a follow up for that?

humphd:
> I think it would be cool to also allow users to copy in 3 different format like with **Download** - `text`, `markdown` and `image`.
> 
> Should we have a follow up for that?
> 
> **Edit:** Sorry I thought this was **Download as Image** PR which is already merged, should have been relevant to post there.

I'm not sure if multiple formats work for the entire chat.  With a single message, we can provide different formats more easily, since we can reference the body of the referenced message.  With an entire chat, we have an array of `ChatCraftMessage` instances, which are basically Markdown content.

If we want to provide more options for download, we better do that in a follow-up, since it would be more than a menu change.

Amnish04:
> > I think it would be cool to also allow users to copy in 3 different format like with **Download** - `text`, `markdown` and `image`.
> > Should we have a follow up for that?
> > **Edit:** Sorry I thought this was **Download as Image** PR which is already merged, should have been relevant to post there.
> 
> I'm not sure if multiple formats work for the entire chat. With a single message, we can provide different formats more easily, since we can reference the body of the referenced message. With an entire chat, we have an array of `ChatCraftMessage` instances, which are basically Markdown content.
> 
> If we want to provide more options for download, we better do that in a follow-up, since it would be more than a menu change.

That's right, it only makes sense for individual messages. For some reason, my subconscious was still reviewing PR #402, but later realized this was a different one.

tarasglek:
lgtm to land. 
image

I guess the only thing left to get rid of the "new chat" header is adding "Rename Chat". We can do that in followup if you like.

I think on mobile we can get rid of the other titlebar(with search) too by adding option to "hide toolbar" or something

humphd:
Switched to gear icon:

Screenshot 2024-02-12 at 8 09 11 AM
Screenshot 2024-02-12 at 8 09 03 AM

PR #436, opened 2024-02-09
Don't include functions with vision model messages

### LINKED ISSUE #435: function_call extra fileds not permitted
Trying to use @mingming-ma's new vision preview feature, I'm seeing this error:

image

humphd:
Fixes #435 

To trigger the bug without this fix:

- Modify your system prompt to include a function, for example:

```
- @fn-url:https://gist.github.com/humphd/74cf88283239c62f53caff5cddf4cfe5
```

- Add an image
- Send the chat
- Notice the error message about validation errors for `function_call`

Now with this branch, you should be able to do it with or without functions defined.


mingming-ma (approved):
Looks good to me!

PR #437, opened 2024-02-10
Sidebar Enhancement Reboot

### LINKED ISSUE #299: Sidebar Enhancement

Amnish04:
Pull Request #300 has been open for a while, and since then a lot of our preferences have changed regarding sidebar.

I really wanted to get it done, but would have been really difficult to continue on that branch. 

Which is why, I have **redone** the sidebar enhancement from scratch.

1. As per our discussion, the overlay effect is only applied to mobile devices now and I am using **Chakra** [Drawer](https://chakra-ui.com/docs/components/drawer/usage) for that.
2. I have also moved the **search field** to the sidebar on mobile devices due to lack of real estate in the header.
3. The behaviour on desktop remains the same, except I have added a few animations to make the **push content** effect smoother.
4. I also made sure there is no redundant animation on page reloads as it was one of the problems in last PR.
5. The Pin Sidebar feature has also been removed

Here's how it looks:

**On Desktop:**
SidebarRebootDesktop

**On Mobile:**
SidebarRebootMobile

**Breakpoint:**
SidebarRebootBreakpoint

This fixes #299

### DIFF
--- a/src/Search.tsx
+++ b/src/Search.tsx
@@ -9,9 +9,18 @@ import {
   Center,
   Card,
   CardBody,
+  Drawer,
+  DrawerOverlay,
+  DrawerContent,
+  DrawerHeader,
+  InputGroup,
+  Input,
+  IconButton,
+  DrawerCloseButton,
+  DrawerBody,
 } from "@chakra-ui/react";
-import { type LoaderFunctionArgs, redirect, useLoaderData } from "react-router-dom";
-import { TbListSearch } from "react-icons/tb";
+import { type LoaderFunctionArgs, redirect, useLoaderData, Form } from "react-router-dom";
+import { TbListSearch, TbSearch } from "react-icons/tb";
 
 import Header from "./components/Header";
 import Sidebar from "./components/Sidebar";
@@ -20,6 +29,7 @@ import Message from "./components/Message";
 import { ChatCraftMessage } from "./lib/ChatCraftMessage";
 import OptionsButton from "./components/OptionsButton";
 import { useSettings } from "./hooks/use-settings";
+import useMobileBreakpoint from "./hooks/use-mobile-breakpoint";
 
 export async function loader({ request }: LoaderFunctionArgs) {
   const url = new URL(request.url);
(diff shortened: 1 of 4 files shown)


humphd:
@Amnish04 can you rebase this, and add the placeholder change there too?

Amnish04:
@humphd Sure, just rebased

src/Search.tsx
```diff
@@ -73,16 +86,42 @@ export default function Search() {
     setSettings({ ...settings, sidebarVisible: newValue });
   }, [isSidebarVisible, settings, setSettings, toggleSidebarVisible]);
 
+  const isMobile = useMobileBreakpoint();
+  const sidebarColor = useColorModeValue("blue.600", "blue.200");
+
+  const sidebarOpenAnimationKeyframes = keyframes`
```
humphd, line 92: We're repeating a bunch of code.  It would be good to think about how we might reuse any of this down the road (doesn't need to block this work).

humphd (approved):
This is a fantastic improvement.  The only thing I wonder about, as I read this, is the amount of code duplication.  However, we have that now, too.

src/Search.tsx
Amnish04: I agree, this duplication has been itching me as well. To address this, I have packed the entire sidebar into its own module like we are doing the `PromptFrom`.

But I don't want to push it on this branch as it might delay this PR. Can I merge this now, so I can open another PR right away as a follow up for this?

Amnish04:
@rjwignar @kliu57 Are you guys also okay with this?

kliu57 (approved):
@Amnish04 I tested it and the mobile changes are really good - big improvement from what it is like in production right now (in prod right now if you have the mobile and the sidebar open and refresh the app becomes unuseable)

Amnish04:
> @Amnish04 I tested it and the mobile changes are really good - big improvement from what it is like in production right now (in prod right now if you have the mobile and the sidebar open and refresh the app becomes unuseable)

Great, thanks for reviewing!

humphd:
Go ahead and land it, then improve in follow-ups.

PR #445, opened 2024-02-13
Fix viewport pushed by switching chats

### LINKED ISSUE #442: Page Moves Up When Clicking Saved Chat
## Issue

Today, I've noticed the ChatCraft window moves up when I access saved chats:
scrollbar issue

The effect increases if I click on an older Chat (lower in the list).

Is anyone else experiencing this issue?

## System Information
Issue was encountered opening ChatCraft on Microsoft Edge on Windows 10

## How to Reproduce

1. Navigate to https://chatcraft.org.
2. Open Saved Chats drawer
3. Open a Chat from the bottom of the list

Amnish04:
This fixes #442 

Just re-read my blog post for the solution
https://dev.to/amnish04/no-backing-away-when-hacking-away-2h27

When the page is loaded and the options menu is never opened, the menu list is overflowing below the visible page, effectively creating hidden scrollbars.
image

The view is fixed when you open the menu at least once
image

But setting the position to fixed eliminates the issue altogether

### DIFF
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -155,7 +155,7 @@ function OptionsButton({
   }, [chat, fetcher]);
 
   return (
-    <Menu>
+    <Menu strategy="fixed">
       {iconOnly ? (
         <MenuButton
           isDisabled={isDisabled}


rjwignar (approved):
Tested and this fixes the problem.
Read your blog too.
It's amazing how helpful it is to blog our work!

Now, does this mean every ChakraUI Menu component added in the future must have `strategy` value of `fixed`?

Amnish04:
@rjwignar That is so true, my blogs have saved me many times.

About setting it to all menus, I am not too sure. Maybe only to ones that cause problems?
@humphd What do you think?

humphd:
I'm also worried about the fact that we have 2 menus now.  I'm sure that within the next week or two, this options menu will need sub-menus, and then we'll have to switch to the new menu component.  Does it need this fix?

I think we need to fix this once for all time.  What if @Amnish04 isn't around to save us?

humphd:
If you want to land this in the mean time and do a better fix later, that's fine too.

Issue #439, opened 2024-02-12
Searchbar is hard to find


rjwignar:
Until recently I didn't know there was a search bar.
![image](https://github.com/tarasglek/chatcraft.org/assets/78163326/1a9901ab-eaf9-4afa-9c8d-f175f7ab2a8f)

Did anyone else have a similar experience?

rjwignar:
Here's what ChatCraft 1.2 thinks:
https://chatcraft.org/api/share/rjwignar/oOes55CNAXukfvwrofjxL

While the search bar is well-positioned, there are always opportunities for improvement in UI/UX design. Here are some recommendations that could enhance the search functionality and overall user experience:

1. **Increase Contrast**: Ensure that the search bar has sufficient contrast against the background to make it easily noticeable in various lighting conditions and to users with visual impairments.

1. **Placeholder Text**: Including placeholder text within the search bar can provide users with a hint about what they can search for or the format they should use, improving usability.

1. **Search Bar Width**: Depending on the expected length of the search queries, consider adjusting the width of the search bar to accommodate the text without scrolling.

1. **Clear Button**: Include a clear (x) button inside the search bar when a user has entered text. This allows users to quickly clear their query and start a new search.

1. **Keyboard Accessibility**: Ensure that users can access the search bar using keyboard shortcuts, which is important for accessibility and power users.

1. **Instant Feedback**: If applicable, implement instant search or suggestions that appear as the user types, which can help users find what they're looking for more quickly.

1. **Consistent Styling**: Make sure the search bar's design is consistent with the overall design language of the application, including fonts, colors, and button styles.

1. **Mobile Responsiveness**: If this interface is to be used on mobile devices, ensure that the search bar is easily accessible and usable on smaller screens.

1. **Search Button**: While the magnifying glass icon is a common symbol for search, adding an explicit 'Search' button could be beneficial for clarity, especially for users who are less familiar with modern UI patterns.

1. **Focus State**: Highlight the search bar with a clear focus state when it's active to provide visual feedback to the user.

---
I can't speak on all of these, but I think placeholder text would make the search bar more noticeable.
What does everyone else think?

Amnish04:
I have a pull request open that effects the positioning of the sidebar on mobile devices.

I guess anyone dealing with this issue should be aware #437

tarasglek:
I agree re placeholder text. Clever use of gpt image feature!

rjwignar:
[...]

PR #448, opened 2024-02-13
Search Bar Placeholder Text

### LINKED ISSUE #447: Add Placeholder Text to Search Bar
I agree re placeholder text. Clever use of gpt image feature!

_Originally posted by @tarasglek in https://github.com/tarasglek/chatcraft.org/issues/439#issuecomment-1938525042_

rjwignar:
Summary
---
Task of #439 

This fixes #447 

This adds placeholder text to the searchbar:
image

Concerns
---
PR https://github.com/tarasglek/chatcraft.org/pull/437 adds additional search bar renders, which may also require placeholder text.

@Amnish04, I think I'll update this PR once #437 lands. Or, would it be better to add placeholder text in your PR?
@humphd which approach would be better?

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -93,7 +93,7 @@ function Header({ chatId, inputPromptRef, searchText, onToggleSidebar }: HeaderP
               name="q"
               defaultValue={searchText}
               isRequired
-              placeholder="Search"
+              placeholder="Search chat history"
             />
             <IconButton aria-label="Search" variant="ghost" icon={<TbSearch />} type="submit" />
           </InputGroup>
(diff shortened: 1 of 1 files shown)


Amnish04:
> @Amnish04, I think I'll update this PR once #437 lands. Or, would it be better to add placeholder text in your PR? @humphd which approach would be better?

I would prefer merging into my branch directly, so you don't have to wait for it to land. But if we do that, we cannot close the issue until #437 lands.

tarasglek:
`Search chat history` might be more clear

rjwignar:
Updated the placeholder text:
image

Amnish04 (approved):
Looks good. Just updated the text in #437 as well.

humphd (approved):

PR #456, opened 2024-02-15
Shorten mobile placeholder

humphd:
Playing with the new GPT-4 Vision on Mobile tonight, and I can't stand the word-wrapping on the placeholder for the input.

I'm sure we'll fix this in a better way later, but this improves things a bit:

Screenshot 2024-02-14 at 8 56 54 PM


Amnish04 (approved):
I also agree. We can do it like this until we find a better place for TTS button.

PR #458, opened 2024-02-15
Add: Resize the image size to meet OpenAI requirements

### LINKED ISSUE #395: Resize the image in the browser within the max size of the image that OpenAI will process
Unresolved task from PR #286

mingming-ma:
## Explanation

This updates make sure users' input image's size is within OpenAI [GPT-4V](https://openai.com/research/gpt-4v-system-card) limit.

The OpenAI requirements for images can be found at [Is there a limit to the size of the image I can upload?](https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload)
 > Yes, we restrict image uploads to 20MB per image.

And [Calculating costs](https://platform.openai.com/docs/guides/vision/calculating-costs)
> `detail: high` images are first scaled to fit within a 2048 x 2048 square, maintaining their aspect ratio. 

I found this tool [browser-image-compression](https://github.com/Donaldcwl/browser-image-compression) focus on the image compression in web browser is great to handle it, supports both file size and px constraints.

I set the values in imageCompressionOptions as:
```
      const imageCompressionOptions = {
        maxSizeMB: 20,
        maxWidthOrHeight: 2048,
        useWebWorker: true,
      };
```

The `useWebWorker` is optional, means the tool will use multi-thread web worker for the compression, fallback to run in main-thread (default: `true`). Although it is default `true`, I added it to align with the document's [Usage](https://github.com/Donaldcwl/browser-image-compression?tab=readme-ov-file#usage)

## How to test

Use some large size image in the prompt, I got a 30MB image from [here](https://sample-videos.com/download-sample-jpg-image.php)

Before

image

After

image

Fixes #395 https://github.com/tarasglek/chatcraft.org/issues/465

### DIFF
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -22,7 +22,7 @@ import { ChatCraftChat } from "../lib/ChatCraftChat";
 import { useUser } from "../hooks/use-user";
 import { useAlert } from "../hooks/use-alert";
 import ShareModal from "./ShareModal";
-import { download } from "../lib/utils";
+import { download, imageCompressionOptions } from "../lib/utils";
 
 function ShareMenuItem({ chat }: { chat?: ChatCraftChat }) {
   const supportsWebShare = !!navigator.share;
@@ -105,13 +105,6 @@ function OptionsButton({
 
       const files = event.target.files;
 
-      const imageCompressionOptions = {
-        maxSizeMB: 20,
-        maxWidthOrHeight: 2048,
-        useWebWorker: true,
-      };
-
-      // Helper function
       const readFile = (file: File) => {
         const reader = new FileReader();
         reader.onload = (e) => {
@@ -123,12 +116,7 @@ function OptionsButton({
       if (files) {
         for (let i = 0; i < files.length; i++) {
           const file = files[i];
-          if (
-            file.type.startsWith("image/") &&
-            // Make sure image's size is within 20MB
-            // https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
-            file.size > imageCompressionOptions.maxSizeMB * 1024 * 1024
-          ) {
+          if (file.type.startsWith("image/")) {
             imageCompression(file, imageCompressionOptions)
               .then((compressedFile) => readFile(compressedFile))
               .catch((err) => {
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -106,7 +106,7 @@ function OptionsButton({
       const files = event.target.files;
 
       const imageCompressionOptions = {
-        maxSizeMB: 20,
+        maxSizeMB: 2,
         maxWidthOrHeight: 2048,
         useWebWorker: true,
       };
@@ -123,21 +123,18 @@ function OptionsButton({
       if (files) {
         for (let i = 0; i < files.length; i++) {
           const file = files[i];
-          if (file.type.startsWith("image/")) {
+          if (
+            file.type.startsWith("image/") &&
             // Make sure image's size is within 20MB
             // https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
-            if (file.size > imageCompressionOptions.maxSizeMB * 1024 * 1024) {
-              imageCompression(file, imageCompressionOptions)
-                .then((compressedFile) => {
-                  return readFile(compressedFile);
-                })
-                .catch((err) => {
-                  console.error(err);
-                  error({ title: "Unable to share chat", message: err.message });
-                });
-            } else {
-              readFile(file);
-            }
+            file.size > imageCompressionOptions.maxSizeMB * 1024 * 1024
+          ) {
+            imageCompression(file, imageCompressionOptions)
+              .then((compressedFile) => readFile(compressedFile))
+              .catch((err) => {
+                console.error(err);
+                error({ title: "Unable to share chat", message: err.message });
+              });
           } else {
             readFile(file);
           }
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -117,6 +117,8 @@ function OptionsButton({
         for (let i = 0; i < files.length; i++) {
           const file = files[i];
           if (file.type.startsWith("image/")) {
+            // Make sure image's size is within 20MB and 2048x2048 resolution
+            // https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
             imageCompression(file, imageCompressionOptions)
               .then((compressedFile) => readFile(compressedFile))
               .catch((err) => {
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -16,13 +16,12 @@ import { BsPaperclip } from "react-icons/bs";
 import { useCallback, useRef } from "react";
 import { useLiveQuery } from "dexie-react-hooks";
 import { useCopyToClipboard } from "react-use";
-import imageCompression from "browser-image-compression";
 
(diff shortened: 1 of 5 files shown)


src/components/OptionsButton.tsx
```diff
@@ -103,19 +104,49 @@ function OptionsButton({
...
+                .catch((err) => {
+                  console.error(err);
+                  error({ title: "Unable to share chat", message: err.message });
+                });
+            } else {
+              readFile(file);
+            }
+          } else {
+            readFile(file);
+          }
```
Amnish04, line 143:
```suggestion
          if (
            file.type.startsWith("image/") &&
            // Make sure image's size is within 20MB
            // https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
            file.size > imageCompressionOptions.maxSizeMB * 1024 * 1024
          ) {
            try {
              const compressedFile = await imageCompression(file, imageCompressionOptions);
              readFile(compressedFile);
            } catch (err: any) {
              console.error(err);
              error({ title: "Unable to share chat", message: err.message });
            }
          } else {
            readFile(file);
          }
```

I see that there's an unnecessary if statement here, we can probably combine both conditions to reduce repetition.
image


Also can we use try catch syntax as for better readability? (Not a big deal anyways as this is a short piece of code)

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -222,13 +223,40 @@ function DesktopPromptForm({
...
+              resolve(base64data);
+            };
+          })
+          .catch((err) => {
+            console.error("Error processing images", err);
+            error({
+              title: "Error Processing Images",
+              message: err.message,
+            });
+          });
```
Amnish04, line 251: Same here

Amnish04 (changes requested):
I've tested this and it works well! Just a few suggestions I commented.

Amnish04:
I was also wondering if we should set a lower maximum size of image upload for better performance, or would that be a negligible difference?

mingming-ma:
@Amnish04 Yeah, that's a good way to try. if we set lower value of `maxSizeMB`, it should improve the response since the payload now have smaller image, and also help reduce the costs. 

Maybe we add options in follow up PRs for the user to select uploaded image with different compression rate or wanted size. User can adjust it accordingly by the result.

src/components/PromptForm/DesktopPromptForm.tsx
mingming-ma: Do you mean the `then()` syntax here? I'd like this way so the function name will not be changed ( add the `async`) and don't need to modify other function invoking places.

Amnish04: I guess we have to manually wrap a Promise here anyways, so you're right we don't need to change this.

src/components/OptionsButton.tsx
```diff
@@ -16,6 +16,7 @@ import { BsPaperclip } from "react-icons/bs";
 import { useCallback, useRef } from "react";
 import { useLiveQuery } from "dexie-react-hooks";
 import { useCopyToClipboard } from "react-use";
+import imageCompression from "browser-image-compression";
```
humphd, line 19: In a follow-up, let's do a dynamic import for this only when we need it, so we don't download it and add weight to the bundle otherwise.

We can land this as is for now.

src/components/OptionsButton.tsx
```diff
@@ -103,19 +104,46 @@ function OptionsButton({
...
 
       const files = event.target.files;
+
+      const imageCompressionOptions = {
+        maxSizeMB: 20,
+        maxWidthOrHeight: 2048,
+        useWebWorker: true,
+      };
+
+      // Helper function
```
humphd, line 114: Get rid of this, it adds nothing.

src/components/OptionsButton.tsx
```diff
@@ -103,19 +104,46 @@ function OptionsButton({
...
-          reader.onload = (e) => {
-            onFileSelected(e.target?.result as string);
-          };
-          reader.readAsDataURL(files[i]);
+          const file = files[i];
+          if (
+            file.type.startsWith("image/") &&
+            // Make sure image's size is within 20MB
+            // https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
+            file.size > imageCompressionOptions.maxSizeMB * 1024 * 1024
```
humphd, line 130: So is this ever going to run? Unless the file is larger than 20M, you won't ever run this.  I feel like we should be running this for images with dimensions larger than 2048 regardless of size.

If I take an image with my phone, it could be 10M, but I don't need to send all that pixel data for an AI to use a 2048x2048 reduced image.  Since OpenAI will throw away all this data after we upload it, in order to shrink it down, let's do it first so we don't a) waste bandwidth; b) store massive images we don't need in the db.

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -222,13 +223,40 @@ function DesktopPromptForm({
   };
 
   const getBase64FromFile = (file: File): Promise<string> => {
+    const imageCompressionOptions = {
```
humphd, line 226: All this duplication of code across the two forms should get put in a separate file, where the image manipulation happens.  It doesn't need to live in React components.

humphd (changes requested):

src/components/OptionsButton.tsx
mingming-ma: Got it, filed https://github.com/tarasglek/chatcraft.org/issues/465 to track

src/lib/utils.ts
```diff
@@ -83,3 +83,10 @@ export const screenshotElement = (element: HTMLElement): Promise<Blob> => {
         })
     );
 };
+
+// Make sure image's size is within 20MB and 2048x2048 resolution
+// https://platform.openai.com/docs/guides/vision/is-there-a-limit-to-the-size-of-the-image-i-can-upload
+export const imageCompressionOptions = {
```
humphd, line 89: Can you move this in here too?

```ts
      imageCompression(file, imageCompressionOptions)
        .then((compressedFile) => {
          const reader = new FileReader();
          reader.readAsDataURL(compressedFile);
          reader.onloadend = () => {
            const base64data = reader.result as string;
            resolve(base64data);
          };
        })
```

humphd (changes requested):

src/lib/utils.ts
humphd: Can we do this too?

mingming-ma: Do you mean the comments? I've moved in [e5a164b](https://github.com/tarasglek/chatcraft.org/pull/458/commits/e5a164be3409f23959a1af536a956582f5191d8a).

mingming-ma: Strange, it's also not showing `Outdated` on my side, don't know why

mingming-ma:
@humphd During refactoring the code, I realized can also solve the dynamic import issue 😄  Let me know if other updates needed. Thanks a lot for the feedback!

src/lib/utils.ts
mingming-ma: Fixed at [6c0c2ee](https://github.com/tarasglek/chatcraft.org/pull/458/commits/6c0c2eed9559f3e1ccb66d49d63b1ff50644b87d)

src/lib/utils.ts
```diff
@@ -83,3 +83,34 @@ export const screenshotElement = (element: HTMLElement): Promise<Blob> => {
...
```
humphd, line 89: Move this inline below into your `compressImageToBase64` function, since we don't use this externally now.

humphd (changes requested):
Much better!  One small fix.

humphd (approved):

PR #459, opened 2024-02-16
Add twitter card metadata

### LINKED ISSUE #432: Add twitter card metadata
Silly twitter doesn't render OG in our shared chats, eg https://chatcraft.org/api/share/tarasglek/5szbfO7gPyjwQLzJE96wb

Need to add twitter-specific cards mentioned in https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards

rjwignar:
This fixes #432.

Summary
---
- Added an `og:type` property.
From https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup,
image
A Twitter Summary card can be rendered if an `og:type`, `og:title` and `og:description` exist but not `twitter:card`, but I ended up also adding required Twitter meta tags for robustness.
- added Twitter meta tags required to render a Summary Card w/ Image

Testing Requirements?
---
I am going to run the generated preview against the [Twitter Card Validator](https://cards-dev.twitter.com/validator).
I hope the validator can test the preview website

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,6 +10,10 @@
     <meta property="og:type" content="website" />
     <meta property="og:title" content="chatcraft.org" />
     <meta property="og:description" content="Web-based AI Assistant for Software Developers" />
+    <meta name="twitter:card" content="summary" />
+    <meta name="twitter:title" content="chatcraft.org" />
+    <meta name="twitter:description" content="Web-based AI Assistant for Software Developers" />
+    <meta name="twitter:image" content="https://chatcraft.org/favicon-32x32.png" />
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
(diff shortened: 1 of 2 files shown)


humphd:
In a follow-up, I'd like to see us take a screenshot of the final message, and upload that along with the chat when we share, and we could use it as our image.  For now, no image is fine.

humphd (changes requested):
This is a good start.

But can you add it to the share logic too?

https://github.com/tarasglek/chatcraft.org/blob/4016e91a955822d1119fd3a5d90b24e3bc4af907/src/lib/share.ts#L59

kliu57 (approved):
Just tested on twitter with the preview url, works well.

image

humphd (approved):

rjwignar:
Tested Twitter meta tags in shared chat, https://rjwignar-issue-432.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/BD3ThijHaa5gfs51H0JEN, against Twitter Card validator and it passes:
image

Here's what it looks like in the Tweet Composer (using summary_large_image):
image

>In a follow-up, I'd like to see us take a screenshot of the final message, and upload that along with the chat when we share, and we could use it as our image. For now, no image is fine.

TODO in followup: Modify twitter:image meta tags to use screenshot of final message.

Issue #453, opened 2024-02-14
Chat Area and Prompt Menu are not aligned perfectly


Amnish04:
The `GridItem` container for Chat Messages has a minor misalignment with the prompt form.

![image](https://github.com/tarasglek/chatcraft.org/assets/78865303/ca9cdc66-d392-43bf-82da-6b88b642991d)

We need to make sure its perfectly aligned.

PR #462, opened 2024-02-16
Replaced padding to use relative value

Rachit1313:
- @Amnish04 and I noticed that the alignment was perfect on a 13 inch macbook whereas was wrong in a monitor of a bigger size.
- Changed padding value to use relative % so that it should adjust according to screen width.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -320,7 +320,7 @@ function ChatBase({ chat }: ChatBaseProps) {
       </GridItem>
 
       <GridItem overflowY="auto" ref={messageListRef} pos="relative">
-        <Flex direction="column" h="100%" maxH="100%" maxW="900px" mx="auto" px={1}>
+        <Flex direction="column" h="100%" maxH="100%" maxW="900px" mx="auto" px="0.9%">
           {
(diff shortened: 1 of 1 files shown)


Rachit1313:
Fixes #453

humphd (changes requested):
This is still wrong:

Screenshot 2024-02-16 at 8 43 05 AM

I think it should be identical to the content above it.

Rachit1313:
> This is still wrong:
> 
> Screenshot 2024-02-16 at 8 43 05 AM
> I think it should be identical to the content above it.

Its a bit confusing in this one. I fixed this based on a monitor with windows, it broke in mac. In our production it is perfect in mac even if I change the zoom scaling to any %.

image

Amnish04:
If you look closely, both GridItems are same width and centred correctly. But its the scrollbar that's pushing our chat area away
image

One thing we can do is set a left margin equal to width of scrollbar, but I don't know how we can get that width as its different across browsers and platforms.
@humphd Can we really fix this?

humphd:
OK, if it's per-platform stuff with scrollbars, we probably won't ever win this battle.  Maybe we should close?

Amnish04:
> OK, if it's per-platform stuff with scrollbars, we probably won't ever win this battle. Maybe we should close?

I agree, lets close it.

PR #463, opened 2024-02-16
Pack Sidebar in its own module

### LINKED ISSUE #457: Refactor the sidebar to eliminate duplication
As a follow up for #437 

We need to eliminate any unnecessary code duplication introduced by the latest styling of sidebar as mentioned at https://github.com/tarasglek/chatcraft.org/pull/437#pullrequestreview-1881610845

As far as I think, it would be good to package the Sidebar functionality in its own module and expose through an API that does not break existing UI.

Maybe something like this
305206869-bb66379b-a4eb-43a8-a082-2de40e7e37a7

Amnish04:
#437 introduced a lot of duplication for sidebar related styles and logic as we now had different sidebar views for mobile and desktop.

I have tried to minimize that duplication, by creating separate components for mobile and desktop sidebar and exposing through a common `Sidebar` api, taking inspiration from the way `PromptForm` is done.

This way, the places that use the sidebar now look the same as before since all the changes have been abstracted in its own module.

Here's a summary of changes:
1. Renamed the existing `Sidebar` component to `SidebarContent`.
2. Added a `SidebarMobile` component that uses `SidebarContent` and extends it by adding the search field and ChatCraft brand at the top.
3. Similarly, `SidebarDesktop` component wraps `SidebarContent` to apply the new open/close transition effect. This way, we don't have to repeat the `keyframe` definitions at evey place where sidebar is used.
4. These Sidebar components are exposed by the common `Sidebar` in index.tsx based on the screen width using `isMobile` hook.

This fixes #457.

### DIFF
--- a/src/components/Sidebar/SidebarMobile.tsx
+++ b/src/components/Sidebar/SidebarMobile.tsx
@@ -31,7 +31,12 @@ function SidebarMobile({
   const brandColor = useColorModeValue("blue.600", "blue.200");
 
   return (
-    <Drawer isOpen={isSidebarVisible} onClose={handleToggleSidebarVisible} placement="left">
+    <Drawer
+      autoFocus={false}
+      isOpen={isSidebarVisible}
+      onClose={handleToggleSidebarVisible}
+      placement="left"
+    >
       <DrawerOverlay />
       <DrawerContent>
         <DrawerHeader mt={2} p={2}>
--- a/src/components/Sidebar/SidebarMobile.tsx
+++ b/src/components/Sidebar/SidebarMobile.tsx
@@ -1,0 +1,72 @@
+import {
+  Drawer,
+  DrawerBody,
+  DrawerCloseButton,
+  DrawerContent,
+  DrawerHeader,
+  DrawerOverlay,
+  IconButton,
+  Input,
+  InputGroup,
+  Text,
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { TbSearch } from "react-icons/tb";
+import { Form } from "react-router-dom";
+import SidebarContent, { SidebarContentProps } from "./SidebarContent";
+
+type SidebarMobileProps = SidebarContentProps & {
+  searchText?: string;
+  handleToggleSidebarVisible: () => void;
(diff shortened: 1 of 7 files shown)


src/components/Sidebar/SidebarMobile.tsx
```diff
@@ -0,0 +1,72 @@
...
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { TbSearch } from "react-icons/tb";
+import { Form } from "react-router-dom";
+import SidebarContent, { SidebarContentProps } from "./SidebarContent";
+
+type SidebarMobileProps = SidebarContentProps & {
+  searchText?: string;
+  handleToggleSidebarVisible: () => void;
+  isSidebarVisible: boolean;
```
WangGithub0, line 21: `SidebarMobile` and `SidebarDesktop` are both have `isSidebarVisible`, could we put it in `SidebarContent`?

kliu57 (approved):
Did some regression testing on your branch and things are working well

src/components/Sidebar/SidebarMobile.tsx
Amnish04: @WangGithub0 We probably can't have it in `SidebarContent` as its not even using those props.

I was actually thinking of managing sidebar triggers inside the `Sidebar` component which is a common ancestor of both `SidebarDesktop` and `SidebarMobile`.
image

But that came with a problem. Currently, many of the UI elements in the `ChatBase` component are using `isSidebarVisible` managed by the `useDisclosure` hook at the top level of component hierarchy.
image

In other words, those elements are not reading directly from the `Settings` context. And when I tried to encapsulate isSidebarVisible management in `Sidebar.tsx` both of those properties got out of sync and started causing unexpected behaviours with opening and closing of sidebar.

Ideally, I think we should be managing the `sidebarVisible` state with `useSettings` hook considering that its state is meant to be single source of truth.

But since I have not used the `useDisclosure` hook before and am not sure if its serving a higher purpose than my understanding, I decided not to touch it and pass that state as props to components. Again, I am also confused which path I should pick here.

PR #469, opened 2024-02-16
Fix search/share id loading bug

### LINKED ISSUE #446: ChatCraft 'crashes' when accessing Shared Chats or Searchbar
@humphd @Amnish04 The last time this happened, did it also affect Shared Chats and the Searchbar?
I'm also seeing issues accessing Shared Chats:
image

and when using the search bar:
image

_Originally posted by @rjwignar in https://github.com/tarasglek/chatcraft.org/issues/442#issuecomment-1940281321_

humphd:
Fixes #446 

This bug was caused by my refactor of the Share button, moving it out of the chat header, and therefore breaking the connection with the route loader.  The way the data was being passed down from the new loader is different (the old way meant it received a chat id, the new way gets an object).  When we try to use this object with Dexie to get a value by ID, it crashes.

This changes how things work so that we pass the actual chat object to the options, and then can operate on it directly vs. needing to talk to the db.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -452,10 +452,11 @@ function ChatBase({ chat }: ChatBaseProps) {
         <Box maxW="900px" mx="auto" h="100%">
           {chat.readonly ? (
             <Flex w="100%" h="45px" justify="end" align="center" p={2}>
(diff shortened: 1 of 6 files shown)


Amnish04 (approved):
Ohh right, not all route loaders give us the chat id.

This was really a showstopper for many people. Thanks for fixing it!

PR #471, opened 2024-02-16
change google redirect_uri: "https://chatcraft.org/api/login"

WangGithub0:


humphd (approved):
This is my guess as to what it would be.  We probably need @tarasglek to look and tell us if this fails.

Let's try it.

PR #301, opened 2023-11-25
Fix model menu zIndex

Amnish04:
I have assigned the `dropdown level` zIndex to the `MenuList` from extended **Chakra Theme** configured for application.

```jsx
<MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
```

This puts our droplist higher in the stacking order, fixing #273.

PR #477, opened 2024-02-21
"Show More..." button's text is improperly layered and visible through the options menu

### LINKED ISSUE #441: "Show More..." button's text is improperly layered and visible through the options menu
image

image

WangGithub0:
According to 
`<Button zIndex={10} size="sm" variant="ghost" onClick={() => onToggle()}>
                      {isOpen ? "Show Less" : "Show More..."}
                    </Button>` in file `https://github.com/tarasglek/chatcraft.org/tree/b87536cced902bc61292c0c12d438bd6270d152d/src/components/Message/MessageBase.tsx`, set a higher zindex for MenuList

closes #441

### DIFF
--- a/src/components/OptionsButton.tsx
+++ b/src/components/OptionsButton.tsx
@@ -182,7 +182,7 @@ function OptionsButton({
           Options
         </MenuButton>
       )}
-      <MenuList>
+      <MenuList zIndex={11}>
         <MenuItem as={ReactRouterLink} to="/new">
           Clear
         </MenuItem>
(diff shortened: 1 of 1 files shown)


Amnish04 (commented):
@WangGithub0 I guess we should use the zIndices defined in the Chakra theme documented here
https://chakra-ui.com/docs/styled-system/theme#z-index-values

In this case, something like `theme.zIndices.dropdown`?

I used it in #301

WangGithub0:
> @WangGithub0 I guess we should use the zIndices defined in the Chakra theme documented here https://chakra-ui.com/docs/styled-system/theme#z-index-values
> 
> In this case, something like `theme.zIndices.dropdown`?
> 
> I used it in #301

yeah, it's more reasonable, thanks for providing it.

Amnish04 (approved):

PR #479, opened 2024-02-22
Provider refactoring - settings.currentProvider, ChatCraftProvider child classes

### LINKED ISSUE #478: Create settings.currentProvider, OpenAiProvider, OpenRouterProvider
Subtask of issue #401 

- Create `providers/OpenAiProvider.ts`, `providers/OpenRouterAiProvider.ts`
- Copy provider-specific methods `providers/OpenAiProvider.ts`, `providers/OpenRouterAiProvider.ts`
- Remove `settings.apiUrl`, `settings.apiKey` and replace it with `settings.currentProvider` which is ChatCraftProvider object

kliu57:
Closes #478 

### Code changes:

- Created `providers/OpenAiProvider.ts`, `providers/OpenRouterAiProvider.ts`, `providers/index.ts` containing provider-specific methods

- Refactored code to use those methods and removed those methods from `ai.ts`

- Removed `settings.apiUrl` and `settings.apiKey` and replaced with `settings.currentProvider`

- Modified code **throughout the project** to use `settings.currentProvider` instead of the now removed `settings.apiUrl` and `settings.apiKey`

- Handled deserialization of `settings.currentProvider` in `deserializer` function of `settings.ts`
- Handled migration of `settings.apiUrl` and `settings.apiKey` deserializer function `settings.ts`. This means current users of ChatCraft will have their `settings.apiUrl` and `settings.apiKey` values migrated to `settings.currentProvider` and `settings.providers`, and then `settings.apiUrl` and `settings.apiKey` will be deleted afterwards from their localStorage

- the `API URL` dropdown field in `Instructions.tsx` now iterates through the list of `supportedProviders` (missed updating this last time)

### Test API key storage in currentProvider:

Open Chrome Dev Tools
image
Your current apiKey and apiUrl is stored in settings.currentProvider
If you have keys of multiple providers stored, your providers are stored in settings.providers array

### Test Migration:

Paste the following to localStorage so that you will have apiKey and apiUrl in localStorage, then refresh the page to test the migration process. You have to replace YOUR_KEY with a valid OpenAI key.

```
{"model":"gpt-3.5-turbo","apiUrl":"https://api.openai.com/v1","temperature":0,"enterBehaviour":"send","countTokens":false,"sidebarVisible":false,"alwaysSendFunctionResult":false,"announceMessages":false,"apiKey":"YOUR_KEY"}
```

### DIFF
--- a/src/components/Message/AiMessage.tsx
+++ b/src/components/Message/AiMessage.tsx
@@ -96,7 +96,7 @@ function AiMessage(props: AiMessageProps) {
 
   const handleRetryClick = useCallback(
     async (model: ChatCraftModel) => {
-      if (!settings.apiKey) {
+      if (!settings.currentProvider?.apiKey) {
         return;
       }
 
@@ -130,7 +130,7 @@ function AiMessage(props: AiMessageProps) {
         setRetrying(false);
       }
     },
-    [props.chatId, settings.apiKey, message, callChatApi]
+    [props.chatId, settings.currentProvider?.apiKey, message, callChatApi]
   );
 
   // While we're streaming in a new version, use a different display
--- a/src/components/PromptForm/index.tsx
+++ b/src/components/PromptForm/index.tsx
@@ -20,7 +20,7 @@ export default function PromptForm(props: PromptFormProps) {
   const isMobile = useMobileBreakpoint();
 
   // Skip showing anything if we don't have an API Key to use
-  if (!settings.apiKey) {
+  if (!settings.currentProvider?.apiKey) {
     return null;
   }
 
--- a/src/hooks/use-settings.tsx
+++ b/src/hooks/use-settings.tsx
@@ -10,6 +10,8 @@ import {
 import { useLocalStorage } from "react-use";
 
 import { serializer, deserializer, key, defaults, type Settings } from "../lib/settings";
+import { OpenRouterProvider } from "../lib/providers/OpenRouterProvider";
+// import { providerFromUrl } from "../lib/providers";
 
 type SettingsContextType = {
   settings: Settings;
@@ -52,7 +54,15 @@ export const SettingsProvider: FC<{ children: ReactNode }> = ({ children }) => {
         .then((data) => {
           const apiKey = data.key;
           if (apiKey !== undefined) {
-            setSettings({ ...state, apiKey: apiKey });
+            const newProvider = new OpenRouterProvider(apiKey);
+            setSettings({
+              ...state,
+              currentProvider: newProvider,
+              providers: {
+                ...state.providers,
+                [newProvider.apiUrl]: newProvider,
+              },
+            });
             // Strip out openRouter's code from the URL
             searchParams.delete("code");
             const { origin, pathname, hash } = window.location;
@@ -74,8 +84,33 @@ export const SettingsProvider: FC<{ children: ReactNode }> = ({ children }) => {
   }, [openRouterCode]);
 
   useEffect(() => {
-    setState(settings || defaults);
-  }, [settings]);
+    if (!settings) {
+      setState(defaults);
+      return;
+    }
+
+    // No current provider, check if we need to handle migration of deprecated apiKey, apiUrl
+    // const { apiUrl, apiKey, ...restSettings } = settings as any;
+    // if (apiKey && apiUrl) {
+    //   const newProvider = providerFromUrl(apiUrl, apiKey);
+    //   setState({
+    //     ...restSettings,
+    //     currentProvider: newProvider,
+    //     providers: {
+    //       ...restSettings.providers,
+    //       [newProvider.apiUrl]: newProvider,
+    //     },
+    //   });
+
+    //   console.warn("Migrated deprecated apiKey, apiUrl");
+    //   return;
+    // }
+
+    setState(settings);
+
+    // switching providers glitches if settings is set as dependency
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   const updateSettings = useCallback(
     (newSettings: Settings) => {
--- a/src/hooks/use-settings.tsx
+++ b/src/hooks/use-settings.tsx
@@ -11,7 +11,6 @@ import { useLocalStorage } from "react-use";
 
 import { serializer, deserializer, key, defaults, type Settings } from "../lib/settings";
 import { OpenRouterProvider } from "../lib/providers/OpenRouterProvider";
-// import { providerFromUrl } from "../lib/providers";
 
 type SettingsContextType = {
   settings: Settings;
@@ -84,31 +83,7 @@ export const SettingsProvider: FC<{ children: ReactNode }> = ({ children }) => {
   }, [openRouterCode]);
 
   useEffect(() => {
-    if (!settings) {
-      setState(defaults);
-      return;
-    }
-
-    // No current provider, check if we need to handle migration of deprecated apiKey, apiUrl
-    // const { apiUrl, apiKey, ...restSettings } = settings as any;
-    // if (apiKey && apiUrl) {
-    //   const newProvider = providerFromUrl(apiUrl, apiKey);
-    //   setState({
-    //     ...restSettings,
-    //     currentProvider: newProvider,
-    //     providers: {
-    //       ...restSettings.providers,
-    //       [newProvider.apiUrl]: newProvider,
-    //     },
-    //   });
-
-    //   console.warn("Migrated deprecated apiKey, apiUrl");
-    //   return;
-    // }
-
-    setState(settings);
-
-    // switching providers glitches if settings is set as dependency
+    setState(settings || defaults);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/src/lib/ChatCraftProvider.ts
+++ b/src/lib/ChatCraftProvider.ts
@@ -1,6 +1,6 @@
 import { nanoid } from "nanoid";
 
-type ProviderName = "OpenAI" | "OpenRouter.ai";
+export type ProviderName = "OpenAI" | "OpenRouter.ai";
 
 export const OPENAI_API_URL = "https://api.openai.com/v1";
 export const OPENROUTER_API_URL = "https://openrouter.ai/api/v1";
@@ -21,7 +21,7 @@ export type SerializedChatCraftProvider = {
   apiKey?: string;
 };
 
-export class ChatCraftProvider {
+export abstract class ChatCraftProvider {
   id: string;
   name: ProviderName;
   apiUrl: string;
@@ -33,23 +33,4 @@ export class ChatCraftProvider {
     this.apiUrl = url;
     this.apiKey = key;
   }
-
-  // Parses url into instance of ChatCraftProvider
-  static fromUrl(url: string, key?: string) {
-    return new ChatCraftProvider(url, key);
-  }
-
-  // Parse from serialized JSON
-  static fromJSON({ id, name, apiUrl, apiKey }: SerializedChatCraftProvider): ChatCraftProvider {
-    const provider = new ChatCraftProvider(apiUrl, apiKey);
-    provider.id = id;
-    provider.name = name;
-    return provider;
-  }
 }
-
-// Currently supported providers
-export const supportedProviders: ProviderData = {
-  OpenAI: new ChatCraftProvider(OPENAI_API_URL),
-  "OpenRouter.ai": new ChatCraftProvider(OPENROUTER_API_URL),
-};
--- a/src/lib/providers/OpenAiProvider.ts
+++ b/src/lib/providers/OpenAiProvider.ts
@@ -1,0 +1,37 @@
+import { queryModels } from "../ai";
+import {
+  ChatCraftProvider,
+  SerializedChatCraftProvider,
+  OPENAI_API_URL,
+  ProviderName,
+} from "../ChatCraftProvider";
+
+export type SerializedOpenAiProvider = {
+  id: string;
+  name: ProviderName;
+  apiUrl: string;
+  apiKey?: string;
+};
+
+export class OpenAiProvider extends ChatCraftProvider {
+  constructor(key?: string) {
+    super(OPENAI_API_URL, key);
+  }
+
+  // Parse from serialized JSON
+  static fromJSON({ apiKey }: SerializedChatCraftProvider): OpenAiProvider {
+    return new OpenAiProvider(apiKey);
+  }
+
+  static async validateApiKey(apiKey: string) {
+    return !!(await queryModels(apiKey));
+  }
+
+  static logoUrl() {
+    return "/openai-logo.png";
+  }
+
+  static defaultModel() {
+    return "gpt-3.5-turbo";
+  }
+}
--- a/src/lib/providers/OpenRouterProvider.ts
+++ b/src/lib/providers/OpenRouterProvider.ts
@@ -1,0 +1,55 @@
+import {
+  ChatCraftProvider,
+  SerializedChatCraftProvider,
+  OPENROUTER_API_URL,
+  ProviderName,
+} from "../ChatCraftProvider";
+
+export type SerializedOpenRouterProvider = {
+  id: string;
+  name: ProviderName;
+  apiUrl: string;
+  apiKey?: string;
+};
+
+export class OpenRouterProvider extends ChatCraftProvider {
+  constructor(key?: string) {
+    super(OPENROUTER_API_URL, key);
+  }
+
+  // Parse from serialized JSON
+  static fromJSON({ apiKey }: SerializedChatCraftProvider): OpenRouterProvider {
+    return new OpenRouterProvider(apiKey);
+  }
+
+  static async validateApiKey(apiKey: string) {
+    // Use response from https://openrouter.ai/docs#limits to check if API key is valid
+    const res = await fetch(`https://openrouter.ai/api/v1/auth/key`, {
+      method: "GET",
+      headers: {
+        Authorization: `Bearer ${apiKey}`,
+      },
+    });
+
+    if (!res.ok) {
+      throw new Error(`${res.status} ${await res.text()}`);
+    }
+
+    return true;
+  }
+
+  static openRouterPkceRedirect = () => {
+    const callbackUrl = location.origin;
+    // Redirect the user to the OpenRouter authentication page in the same tab
+    location.href = `https://openrouter.ai/auth?callback_url=${encodeURIComponent(callbackUrl)}`;
+  };
+
+  static logoUrl() {
+    // If we don't know, use the OpenAI logo as a fallback
+    return "/openai-logo.png";
+  }
+
+  static defaultModel() {
+    return "openai/gpt-3.5-turbo";
+  }
+}
--- a/src/lib/settings.ts
+++ b/src/lib/settings.ts
@@ -6,18 +6,16 @@
  * when you only need to read something.
  */
 import { ChatCraftModel } from "../lib/ChatCraftModel";
-import { ProviderData } from "../lib/ChatCraftProvider";
+import { ProviderData, ChatCraftProvider } from "../lib/ChatCraftProvider";
+import { providerFromJSON, providerFromUrl } from "./providers";
+import { OpenAiProvider } from "./providers/OpenAiProvider";
 /**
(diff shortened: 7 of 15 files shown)


tarasglek:
These providers have too much copy-pasted code. Maybe make openrouter extend openai?

kliu57:
> These providers have too much copy-pasted code. Maybe make openrouter extend openai?

I could remove the `createClient` function from the two providers and we just use the `createClient` function in [ai.ts](https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ai.ts) which tackles both.

src/lib/providers/OpenAiProvider.ts
```diff
@@ -0,0 +1,69 @@
+import OpenAI from "openai";
+import { ChatCraftProvider, SerializedChatCraftProvider } from "../ChatCraftProvider";
+
+type ProviderName = "OpenAI" | "OpenRouter.ai";
```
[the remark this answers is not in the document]
kliu57: yes I can do that

src/lib/ChatCraftProvider.ts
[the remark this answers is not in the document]
kliu57: I can't make it abstract yet. Right now we are still using ChatCraftProvider and we have not switched to full on using OpenAiProvider and OpenRouterProvider. Once I make it abstract we cannot create ChatCraftProvider objects anymore.

src/lib/providers/OpenAiProvider.ts
```diff
@@ -0,0 +1,69 @@
...
+  constructor(key?: string) {
+    super("https://api.openai.com/v1", key);
+  }
+
+  static createClient = (apiKey: string) => {
+    const headers = undefined;
+    return {
+      openai: new OpenAI({
+        apiKey: apiKey,
+        baseURL: "https://api.openai.com/v1",
```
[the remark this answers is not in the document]
kliu57: I think we should use api url constants defined in ChatCraftProvider.ts instead of hardcoding strings. - Yes I can do that

instead of defining global constants, shouldn't API_URL be a static property of every concrete Provider class? - currently ChatCraftProvider class needs access to these two urls. If I put it in their respective classes and import them into ChatCraftProvider, since they are are child classes I run into a circular dependency issue. To get around that I would need to use dynamic imports. Not sure if this is worth the trouble..

src/lib/ChatCraftProvider.ts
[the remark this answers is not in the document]
kliu57: right now I'm pushing the code for OpenAiProvider and OpenRouterProvider in advance but they are not in use as of yet. Once they are in use then I can make ChatCraftProvider abstract and we wouldn't need fromUrl anymore.

src/lib/providers/OpenRouterProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      return models.map((model: any) => model.id) as string[];
+    } catch (err: any) {
+      throw new Error(err.message ?? `error querying models API`);
+    }
+  }
+
+  static async validateApiKey(apiKey: string) {
+    // Use response from https://openrouter.ai/docs#limits to check if API key is valid
+    const res = await fetch(`https://openrouter.ai/api/v1/auth/key`, {
```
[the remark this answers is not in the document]
kliu57: the full string has to be this path for the api call to work. Is there benefit to just using the OPENROUTER_API_URL for the first part of the string? having to concatenate a string wouldn't make it faster or anything

src/lib/providers/OpenRouterProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+      const models = [];
+      for await (const page of openai.models.list()) {
+        models.push(page);
+      }
+
+      return models.map((model: any) => model.id) as string[];
+    } catch (err: any) {
+      throw new Error(err.message ?? `error querying models API`);
+    }
+  }
```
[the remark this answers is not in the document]
kliu57: ok I can do that

src/lib/providers/OpenRouterProvider.ts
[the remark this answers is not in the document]
Amnish04: It not about making fast, but to limit the hard coded part of the string. Its a common practice to concatenate the base url and pathnames, but up to you

src/lib/providers/OpenAiProvider.ts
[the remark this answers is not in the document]
Amnish04: Makes sense

src/lib/providers/OpenRouterProvider.ts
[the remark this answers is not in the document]
kliu57: actually will just keep the function in its current location in ai.ts for now

src/lib/ChatCraftProvider.ts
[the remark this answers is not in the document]
kliu57: I will make the change to this method so that it returns not ChatCraftProvider objects but specific OpenAiProvider and OpenRouterProvider object

kliu57:
@tarasglek for the duplicated methods such as `createClient` and `queryModels` from `OpenAiProvider`, `OpenRouterProvider` to avoid too much duplicate code and just kept them in [`ai.ts`](https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ai.ts).

@Amnish04 Made the changes you requested:

- removed `createClient`, `queryModels` methods from `OpenAiProvider.ts` and `OpenRouterProvider.ts` since it's better for those methods to stay in [`ai.ts`](https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ai.ts) (less duplicate code)
- replaced hardcoded strings `"https://api.openai.com/v1"`, `"https://openrouter.ai/api/v1"` with `OPENAI_API_URL`, `OPENAI_API_URL` exported from [`ChatCraftProvider.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-163489fb9d67f12a96f71b1a06ec2c3cc41152124ad90c14adcc6ded77869f1a)
- removed `OPENAI_API_URL`, `OPENROUTER_API_URL` from [`lib/settings.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-b5f5521c9795e314efdf6293ad3afda789a1402bbf5e2a304e20efa8134d2a15) and replaced imports to be importing from [`ChatCraftProvider.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-163489fb9d67f12a96f71b1a06ec2c3cc41152124ad90c14adcc6ded77869f1a)
- removed duplicate code `type ProviderName = "OpenAI" | "OpenRouter.ai";` by exporting it from [`ChatCraftProvider.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-163489fb9d67f12a96f71b1a06ec2c3cc41152124ad90c14adcc6ded77869f1a)
- changed `supportedProviders` to [`loadSupportedProviders()`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-163489fb9d67f12a96f71b1a06ec2c3cc41152124ad90c14adcc6ded77869f1a), which now no longer returns associative array of `ChatCraftProvider` objects. It now returns associatitve array of child objects (one `OpenAiProvider` and one `OpenRouterProvider`). Made subsequent changes in [`Instructions.tsx`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-9fcdb80f3814fd658a3bb7dcaf58478d27938a43cde9049d449210652e77434c) and [`PreferencesModel.tsx`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-1025748a16e04005898c20c05840616be48ca0eb4eb4c156248de41a2f3bdd7d) for calling the async function `loadSupportedProviders()` using useEffects etc.

tarasglek:
Code looks much better now. Is there any expected change in behavior or is this a pure refactoring?

kliu57:
> Code looks much better now. Is there any expected change in behavior or is this a pure refactoring?

No change in behaviour. Mostly refactoring and the introduction of the OpenAiProvider and OpenRouterProvider classes as child classes of ChatCraftProvider.

tarasglek:
lgtm

kliu57:
@humphd Need some guidance. Everything works well prior to commit d7d415c11771853de96216efc759401dfd8a481c.

But in commit d7d415c11771853de96216efc759401dfd8a481c I replaced `settings.apiUrl` and `settings.apiKey` with `settings.currentProvider`. This is a little buggy. I believe the reason it is buggy is because I am using useStates and useEffects in multiple files to access `settings.currentProvider`. I couldn't directly access `settings.currentProvider` since it needs to be deserialized (thus I decided to go with a useEffect). With the useEffects it was working well when I has it just in two files, but once I put it in many more files the buggy behaviour started appearing.

kliu57:
[UPDATE] Fixed buggy behaviour by switching from using useEffect and useStates to context. Deserialization of `settings.currentProvider` is now done in one `use-settings` file instead of the previous multiple useEffects. Code ready for testing and review. My tests so far have been successful.

src/components/Message/AiMessage.tsx
```diff
@@ -96,7 +96,7 @@ function AiMessage(props: AiMessageProps) {
 
   const handleRetryClick = useCallback(
     async (model: ChatCraftModel) => {
-      if (!settings.apiKey) {
+      if (!settings.currentProvider || !settings.currentProvider.apiKey) {
```
humphd, line 99: This type of check can use `if (!settings.currentProvider?.apiKey) }` to cover both cases

src/lib/ChatCraftProvider.ts
```diff
@@ -35,21 +35,48 @@ export class ChatCraftProvider {
   }
 
   // Parses url into instance of ChatCraftProvider
-  static fromUrl(url: string, key?: string) {
-    return new ChatCraftProvider(url, key);
+  static async fromUrl(url: string, key?: string) {
+    if (url === OPENAI_API_URL) {
+      const { OpenAiProvider } = await import("./providers/OpenAiProvider");
```
humphd, line 40: Let's not do a dynamic import on these, and then you don't need to have `fromUrl()` be async.  The size saving on these two files is minimal, they can become part of the bundle.

The time when we want to do this is when we're loading something huge (like we saw today in the call with the `python.wasm` file).

src/components/PromptForm/index.tsx
```diff
@@ -20,7 +20,10 @@ export default function PromptForm(props: PromptFormProps) {
   const isMobile = useMobileBreakpoint();
 
   // Skip showing anything if we don't have an API Key to use
-  if (!settings.apiKey) {
+  if (!settings.currentProvider) {
+    return null;
+  }
+  if (!settings.currentProvider.apiKey) {
```
humphd, line 26: Combine this into a single check like I showed above with `?.apikey`

src/hooks/use-chat-openai.ts
```diff
@@ -73,7 +73,7 @@ function useChatOpenAI() {
       let ttsWordsBuffer = "";
       const sentenceEndRegex = new RegExp(/[.!?]+/g);
 
-      const chat = chatWithLLM(messages, {
+      const chat = await chatWithLLM(messages, {
```
humphd, line 76: Why this change?

src/hooks/use-settings.tsx
```diff
@@ -33,6 +34,7 @@ export const SettingsProvider: FC<{ children: ReactNode }> = ({ children }) => {
   });
 
   const [state, setState] = useState<Settings>(settings || defaults);
+  const [isMigrated, setIsMigrated] = useState(false);
```
humphd, line 37: This is legacy code we're not going to need/want later.  A better way to do this is to check for the existence of the old stuff and migrate to the new vs. having this be a permanent part of our API.

humphd (changes requested):
@kliu57 I've done a first pass through this code, and made some suggestions to simplify and improve.  I'll review again once you fix the various things I've highlighted and let me know.

Overall, this is coming along really well.

src/lib/ChatCraftProvider.ts
kliu57: @humphd If I try to do a regular import a the top of the file I run into a circular dependency issue
Any ideas how to overcome that?

Amnish04 (commented):
A few things I noticed

src/hooks/use-settings.tsx
kliu57: @humphd  Where should I do this check? Right now I am checking it in a useEffect in use-settings

src/components/Message/AiMessage.tsx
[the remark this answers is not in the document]
kliu57: I will fix this

src/components/PromptForm/index.tsx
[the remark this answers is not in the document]
kliu57: I will fix this

src/components/Message/AiMessage.tsx
```diff
@@ -130,7 +130,7 @@ function AiMessage(props: AiMessageProps) {
         setRetrying(false);
       }
     },
-    [props.chatId, settings.apiKey, message, callChatApi]
+    [props.chatId, settings.currentProvider, message, callChatApi]
```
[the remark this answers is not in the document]
Amnish04: Nvm that's just a part of useEffect dependencies

src/lib/ChatCraftProvider.ts
humphd: You could move the code for these other providers into this file if you want, since it isn't that much.  We do the same the ChatCraftMessage, etc.

src/components/Message/AiMessage.tsx
kliu57: I will fix this

src/components/PromptForm/index.tsx
kliu57: I will fix this

src/hooks/use-chat-openai.ts
kliu57: Seems I have make a mistake here. chatWithLLM did not need to become async since I handled deserialization in the context. I will fix this.

src/lib/speech-recognition.ts
```diff
@@ -146,7 +146,11 @@ export class SpeechRecognition {
 
     // Turn audio into text via OpenAI and Whisper. If we don't have enough audio
     // for the file to contain any data, return null.
-    return file.size > 0 ? transcribe(file) : null;
+    try {
+      return file.size > 0 ? await transcribe(file) : null;
+    } catch (error) {
+      console.error("Error executing stop", error);
+    }
```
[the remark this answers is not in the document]
kliu57: Seems I have make a mistake here. transcribe did not need to become async since I handled deserialization in the context. I will fix this.

src/lib/ChatCraftProvider.ts
kliu57: @humphd Do we still want to have OpenAiProvider and OpenRouterProvider class in our project? Because if we do then I will need to import them in order for fromUrl to return the appropriate object (either an OpenAiProvider object or OpenRouterProvider object). And I can't get that done without using dynamic import.

humphd: The solution to this problem is to separate the logic for `fromUrl()` from the base class.  For example, you could move the `if/else` logic to the calling code instead of putting it in `fromUrl(), making the `fromUrl()` an `abstract` method that gets implemented by derived classes.

kliu57: @humphd I thought the point of the fromUrl function is we can check a URL and from that string figure out whether we want to return an OpenRouterProvider object or OpenAiProvider object? if fromURL is in OpenAiProvider what use would it have? Instead of calling `OpenAiProvider.fromUrl` I could just call `new OpenAiProvider()` to make a new object.

kliu57: @humphd Sorry I'm still a bit confused. I definitely can make fromUrl abstract and instantiate it in the child classes. But then when the user changes their API URL in the Select dropdown... how would I take that string URL and get a ChatCraftProvider object from it? Right now I am using ChatCraftProvider.fromUrl, but if I make this change I won't be able to any longer.

kliu57:
@humphd @Amnish04 Pushed new code which resolves redundant checks (i.e. using '?' => `currentProvider?.apiKey`)
Also removed unecessary code changes

**These two issues still in discussion:**
- [dynamic import issue](https://github.com/tarasglek/chatcraft.org/pull/479/files/ac3e67df87755ad9f94d9864bdfa323de83bc041#r1514902689)
- [migration code issue](https://github.com/tarasglek/chatcraft.org/pull/479#discussion_r1514906887)

src/hooks/use-settings.tsx
Amnish04: I think what @humphd means is that you don't need to maintain an `isMigrated` state.

You just need to check if the old `apiUrl` and `apiKey` are there, update to the new model.

Which I think you are already doing here
image

I just tried after removing `isMigrated`, manipulating my settings in localStorage to old format, and still works fine meaning that the flag isn't doing anything anyways

src/lib/ChatCraftProvider.ts
kliu57: as per our call, I will move the functions that require dynamic imports to providers/index.ts thus removing the need for dynamic imports

src/hooks/use-settings.tsx
kliu57: as per our call, I will remove the state variable as well as move my code additions out of this file and into settings.ts => deserializer

kliu57:
@humphd 

Solved these two issues as per our call

- dynamic import issue - no more dynamic imports, moved code out of `ChatCraftProvider.ts` and into [`providers/index.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-4c5f3070b0bc8cc1848bbd1824438c5c36d980675884a7a9419c904e865fe81e)
- migration code issue - removed state variable, moved code out of `hooks/use-settings.tsx` and into `deserializer` function of [`lib/settings.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-b5f5521c9795e314efdf6293ad3afda789a1402bbf5e2a304e20efa8134d2a15). The code in the `deserializer` includes deserialization of `currentProvider` and migration code

Everything is working. However I have a few questions for you:

1. Right now the migration code is inside `deserializer` function of [`lib/settings.ts`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-b5f5521c9795e314efdf6293ad3afda789a1402bbf5e2a304e20efa8134d2a15). However it runs many times as you can see by the console.warn messages below. Is this a problem? Once the migration finishes it stop running, but at the very beginning it will run many times. In my old approach when I had the migration code inside the useEffect in [`use-settings.tsx`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-44205fdbbda68a7a732c89f3e7dddb38e6e065a794b4f85eb191ce4e4cc31293) it only runs once. I left the old code commented out in there for now in case you want to compare the two.

image

2. The old code for the [`use-settings.tsx`](https://github.com/tarasglek/chatcraft.org/pull/479/files#diff-44205fdbbda68a7a732c89f3e7dddb38e6e065a794b4f85eb191ce4e4cc31293) file:

image

After I added my code in `settings.ts` to deserialize currentProvider, if I do not change this useEffect's dependency from [settings] to [], then I get glitching whenever I try to swap providers by toggling the dropdown. It seems like by keeping [settings] as the dependency this useEffect is running at the same time that my swap provider code is running and so my code changes the provider, then this useEffect switches it back again. You will notice a "glitching" effect. By settings the dependency to [] I have fixed the problem but I just want to check with you if it is ok to make this change since I am not 100% clear.

src/lib/settings.ts
```diff
@@ -56,6 +55,21 @@ export const deserializer = (value: string): Settings => {
     settings.model = new ChatCraftModel(settings.model);
   }
 
+  if (settings.currentProvider) {
```
humphd, line 58: This all seems good to me.  It doesn't really matter how many times it runs, because eventually it will get serialized back to JSON and stored.

Amnish04 (approved):

src/hooks/use-settings.tsx
```diff
@@ -10,6 +10,8 @@ import {
 import { useLocalStorage } from "react-use";
 
 import { serializer, deserializer, key, defaults, type Settings } from "../lib/settings";
+import { OpenRouterProvider } from "../lib/providers/OpenRouterProvider";
+// import { providerFromUrl } from "../lib/providers";
```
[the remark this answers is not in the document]
kliu57: removed

PR #482, opened 2024-02-23
updated KaTeX rule in default system prompt to address inline KaTeX

### LINKED ISSUE #460: Inline KaTeX Rendering
This message renders partially correct, but the inline bits in the first sentence don't work.  I wonder if we could update our default prompt so they do?

```
To convert bytes to megabytes (MB), you can divide the number of bytes by \(2^{20}\) (since 1 MB is \(2^{20}\) bytes or 1,048,576 bytes).

Given the number of bytes as 20,971,520:

$$\frac{20,971,520 \text{ bytes}}{2^{20}} = \frac{20,971,520}{1,048,576} = 20 \text{ MB}$$

So, 20,971,520 bytes is equal to 20 MB.
```

Screenshot 2024-02-15 at 9 44 55 PM

rjwignar:
This fixes #460

Summary
---
I've updated the system prompt to also wrap inline response elements in double-dollar signs for KaTeX rendering.

Observations
---
It looks like in the original issue, ChatCraft returns in-line math using [LaTeX math mode](https://www1.cmc.edu/pages/faculty/aaksoy/latex/latexthree.html#:~:text=There%20are%20a%20few%20ways,%5Cend%7Bequation%7D%20commands.) 
(e.g., `\(2^{20}\)` instead of `$$(2^{20})$$`)
When working on this issue I found that I had to be very explicit with the math rendering rules for consistent results.
I found that explicitly requesting KaTeX syntax for inline math expressions fixes the issue in #460 by either rendering inline math expressions in KaTeX or avoiding LaTeX math mode syntax.
I'll admit my current rules are verbose, but trying to shorten these rules gave me poor results.

Tests
---
[Replicating the conditions ](https://issue-460.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/hIlmNr-NQCNk8P6Yl3wxD) from #460: 
```
Sure! When converting bytes to megabytes, we divide the number of bytes by 1024 twice because:

1. There are 1024 bytes in a kilobyte (KB).
2. There are 1024 kilobytes in a megabyte (MB).

So, to convert bytes to megabytes, we divide the number of bytes by $$1024^2$$. This calculation gives us the equivalent size in megabytes.
```
image

[Asking ChatCraft to describe a formula](https://issue-460.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/NDF2k7zbEg-RonWoi0kPz):
```
The Michaelis-Menten equation is a mathematical model that describes the rate of enzymatic reactions. It is commonly used in biochemistry to understand how enzymes interact with substrates. The formula is given by:

$$V = \frac{V_{max} \cdot [S]}{K_m + [S]}$$

Where:
- $$V$$ is the initial reaction velocity
- $$V_{max}$$ is the maximum reaction velocity
- $$[S]$$ is the substrate concentration
- $$K_m$$ is the Michaelis constant

The Michaelis-Menten equation shows that the reaction velocity ($$V$$) increases with increasing substrate concentration up to a point where the enzyme becomes saturated (approaching $$V_{max}$$). The Michaelis constant ($$K_m$$) represents the substrate concentration at which the reaction velocity is half of the maximum velocity.

This formula is essential for understanding enzyme kinetics and is widely used in biochemistry and pharmacology.
```
image

[Quadratic formula](https://issue-460.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/20jBW9TyqtDyf7OIX8ilC):\
```
The quadratic formula is a formula used to find the roots of a quadratic equation of the form $$ax^2 + bx + c = 0$$. The roots are the values of $$x$$ that satisfy the equation.

The quadratic formula is given by: 

$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$

where:
- $$a$$, $$b$$, and $$c$$ are coefficients of the quadratic equation $$ax^2 + bx + c = 0$$
- $$\pm$$ indicates that there are two possible solutions, one with a plus sign and one with a minus sign
- $$\sqrt{b^2 - 4ac}$$ is the square root of the discriminant, which determines the nature of the roots:
  - If $$b^2 - 4ac > 0$$, the equation has two distinct real roots
  - If $$b^2 - 4ac = 0$$, the equation has one real root (a repeated root)
  - If $$b^2 - 4ac < 0$$, the equation has two complex roots

By substituting the coefficients $$a$$, $$b$$, and $$c$$ into the formula, you can find the values of $$x$$ that satisfy the quadratic equation.
```
image

### DIFF
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,7 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math markdown, inline or otherwise, I use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs ($$), for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
+- If responding with math markdown, inline or otherwise, I use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs \`($$)\`, for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
 `;
 
 export const defaultSystemPrompt = () => {
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,7 +10,9 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your math Markdown, inline or otherwise, by enclosing it with double-dollar signs rather than parentheses, for example: $$O(n\\log n)$$
+- If responding with math, I follow these rules:
+  - Inline or otherwise, I ALWAYS use KaTeX syntax in math Markdown by ALWAYS enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs ($$), for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
+  - I NEVER wrap mathematical expressions, equations, variables, or formulas in brackets or parentheses. For example, instead of responding with \(V_{max}\), I would respond with $$V_{max}$$
 `;
 
 export const defaultSystemPrompt = () => {
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,7 +10,7 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your Markdown by enclosing it with double-dollar signs rather than parentheses, for example: $$O(n\\log n)$$
+- If responding with math, use KaTeX syntax in your math Markdown, inline or otherwise, by enclosing it with double-dollar signs rather than parentheses, for example: $$O(n\\log n)$$
 `;
 
 export const defaultSystemPrompt = () => {
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -10,9 +10,8 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, I follow these rules:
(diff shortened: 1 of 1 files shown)


tarasglek:
lgtm

humphd (commented):
The one improvement we might make to this is to be clear that we don't want to do this if we're inside a fenced code block; otherwise we might inadvertently have it start mixing code and KaTeX.  Your imperative style here is likely going to force it to do this in unexpected places.  We only want KaTeX syntax when we're doing math, not for code.

src/lib/system-prompt.ts
```diff
@@ -10,7 +10,8 @@ I follow these rules when responding:
 - Format ALL lines of code to 80 characters or fewer
 - Use Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
-- If responding with math, use KaTeX syntax in your Markdown by enclosing it with double-dollar signs rather than parentheses, for example: $$O(n\\log n)$$
+- Inline or otherwise, I ALWAYS use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs ($$), for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
```
mingming-ma, line 13: Looks good me, I found the system prompt will be impacted, maybe we can use `` here

```suggestion
- Inline or otherwise, I ALWAYS use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs `$$`, for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
```

mingming-ma: Oh my bad this will break the code 😢

mingming-ma: Maybe using subitem ?
```suggestion
- Inline or otherwise, I ALWAYS use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs ($$), for example: 
  - $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
```

rjwignar: I find 
```
\`($$)\`
```

seems to work

rjwignar:
> The one improvement we might make to this is to be clear that we don't want to do this if we're inside a fenced code block; otherwise we might inadvertently have it start mixing code and KaTeX. Your imperative style here is likely going to force it to do this in unexpected places. We only want KaTeX syntax when we're doing math, not for code.

@humphd I've been experimenting with ways to do this.
I've tried it by adding a "When responding with math," clause in the rule and it seems to work. Or should I explicitly add something like "I never use KaTeX syntax for code"?

The only issue here is that it breaks syntax formatting when describing scientific equations:
image


However, it's still fine for math and conversion factors (like bytes to megabytes). 
Since ChatCraft is developer-focused, should we ignore this behaviour with scientific formulas for now?

mingming-ma:
> The only issue here is that it breaks syntax formatting when describing scientific equations:

Did you use gpt3 for the test? I found gpt4 understands better and will always give correct format even in describing. It maybe due to model limitation.

humphd (approved):
Let's try it and monitor it.  If it gets out of control and starts doing this in code, we'll have to adjust.

PR #484, opened 2024-02-23
Make AudioPlayer queue global

### LINKED ISSUE #391: Stop playing audio when tts is turned off
Follow up for #357 

Currently, when text to speech is turned off
image

The announcement of response does not stop. 
We need to make sure any tts audio playing is stopped and audio player queue needs to be cleared.

Amnish04:
In #357, I was able to stream audio responses as the LLM response was generated and also optimize the playback.

But there were some problems with it, and follow ups were filed.

This is regarding one of those follow ups - #391.

The problem was that once an audio clip was added to the audio queue, there was no way to stop it from other parts of the application since every invocation to `useAudioPlayer` hook resulted in a fresh audio queue. This was pretty annoying and made the TTS feature unusable as users had to wait for the previous message announcement to finish before it started playing for the next one.

To fix this issue, I have made the Audio Player queue global, by creating and providing an `AudioPlayerContext` at the root of the application.

https://github.com/tarasglek/chatcraft.org/compare/amnish04/global-audio-player?expand=1#diff-1cd8b18798a1a103bfe13bef54354c1f3a3bea29a31c8eea1a0c67a3a839b811

This allowed me to expose another function from the hook called `clearAudioQueue` which can now be called from any application component to pause the current audio clip, and flush the remaining clips in the queue.

Currently, I am using it for 2 scenarios:
1. The audio for previous message will instantly stop playing as soon as the user asks a new question.
2. Any audio clips being played via `useAudioPlayer` will instantly stop as soon as the user disables the TTS setting.

This approach also makes the audio system more robust any we now have a single source of truth for audio clips, and the risk to play multiple audio clips at once is also reduced.

This fixes #391

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -45,6 +45,7 @@ import { ChatCraftCommandRegistry } from "../lib/commands";
 import { ChatCraftCommand } from "../lib/ChatCraftCommand";
 import useMobileBreakpoint from "../hooks/use-mobile-breakpoint";
 import { TbSearch } from "react-icons/tb";
+import useAudioPlayer from "../hooks/use-audio-player";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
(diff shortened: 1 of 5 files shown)


tarasglek:
This is great will get back to you once i had a chance to test it

tarasglek:
There is a bug:when I record a new message it keeps talking while I'm recording a reply. It should stop immediately

Amnish04:
That's right, just pushed the fix!

tarasglek:
works awesome now. lets land

tarasglek (approved):

Issue #312, opened 2023-12-26
Enable python, go, tsc, etc support via wasi


tarasglek:
Currently chatcraft works best with js and typescript(thanks to esbuild wasm). It enables complete idea->prototype workflow including the debug cycle where gpt automagically infers fixes inside the chatcraft. However once you get into go/python, it's a lot less satisfying due to having cut/paste code between chatcraft and editor.

Lets solve this the next couple of languages, see https://snarky.ca/wasi-support-for-cpython-june-2023/ here for more details on python

tarasglek:
Looks like most work was already done https://antonz.org/in-browser-code-playgrounds/

tarasglek:
Looks like for go we should use the official go playground as it supports modules https://go.dev/play/p/XCoNXEjm3TS

humphd:
I think a good first step here would be to try integrating the https://github.com/nalgeon/python-wasi code so we can run arbitrary Python.  See his other `*-wasi` modules too https://github.com/nalgeon?tab=repositories&q=wasi&type=&language=&sort=

humphd:
Here is his online version of the `python-wasi` stuff for experiments: https://codapi.org/embed/?sandbox=python (see `Run` button at bottom to execute).

tarasglek:
Related https://urllib3.readthedocs.io/en/stable/reference/contrib/emscripten.html

WangGithub0:
> https://github.com/nalgeon/python-wasi

Tested this python-wasi, it worked on my computer using this test:
```
const url = "https://unpkg.com/@antonz/python-wasi/dist/python.wasm";

const pythonScript = `
a = 1
b = 2
c = a + b
print(c)
print("Hello World!")
`;

import("@antonz/runno").then(({ WASI }) => {
    WASI.start(fetch(url), {
        args: ["python", "-c", pythonScript],
        stdout: (out) => console.log(out),
        stderr: (err) => console.error(err),
    }).then(result => console.log(`exit code = ${result.exitCode}`));
});
```

```
Output:
3
Hello World!

exit code = 0
```




@humphd  should I combine this in `src/lib/run-code.ts` file to support run a python file in a chat? Should I also add this in `src/lib/ChatCraftFunction.ts` to support add python function?

tarasglek:
yes that seems like the way @WangGithub0

WangGithub0:
[...]

PR #485, opened 2024-03-06
Add python-wasi on run-code

WangGithub0:
Add python-wasi in run-code, and also addressing typeScript type declarations for it.
#312 

image

image

### DIFF
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -16,7 +16,7 @@ import { TbCopy, TbDownload, TbRun, TbExternalLink } from "react-icons/tb";
 
 import { download, formatAsCodeBlock } from "../lib/utils";
 import { useAlert } from "../hooks/use-alert";
-import { isRunnable, isRunnableOnServer, runCode } from "../lib/run-code";
+import { isRunnableInBrowser, isRunnableOnServer, runCode } from "../lib/run-code";
 
 type PreHeaderProps = {
   language: string;
@@ -38,8 +38,9 @@ function CodeHeader({
   const { onCopy } = useClipboard(code);
   const { info, error } = useAlert();
   // Only show the "Run" button for JS code blocks, and only when we aren't already loading
-  const shouldShowRunButton = isRunnable(language) && onPrompt;
-  const shouldShowRunOnServer = isRunnableOnServer(language) && onPrompt;
+  const shouldShowRunButton =
+    (isRunnableInBrowser(language) || isRunnableOnServer(language)) && onPrompt;
+  const shouldShowRunMenuList = isRunnableOnServer(language) && onPrompt;
 
   const handleCopy = useCallback(() => {
     onCopy();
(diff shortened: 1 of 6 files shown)


src/components/CodeHeader.tsx
```diff
@@ -16,7 +16,7 @@ import { TbCopy, TbDownload, TbRun, TbExternalLink } from "react-icons/tb";
 
 import { download, formatAsCodeBlock } from "../lib/utils";
 import { useAlert } from "../hooks/use-alert";
-import { isRunnable, runCode } from "../lib/run-code";
+import { isRunnable, isRunnableOnServer, runCode } from "../lib/run-code";
```
humphd, line 19: I wonder if we should rename `isRunnable` to `isRunnableInBrowser` to match what you're doing here.

src/lib/run-code.ts
```diff
@@ -11,7 +11,9 @@ globalThis.__esbuildWasmLoaded = false;
 
 const supportedJS = ["js", "javascript"];
 const supportedTS = ["ts", "typescript"];
-const supportedLanguages = [...supportedJS, ...supportedTS];
+const supportedPY = ["py", "python"];
+const supportedLanguages = [...supportedJS, ...supportedTS, ...supportedPY];
```
humphd, line 15: `supportedBrowserLanguages` maybe?

src/lib/run-code.ts
```diff
@@ -105,6 +115,41 @@ async function runJavaScript(code: string) {
   }
 }
 
+async function runPython(code: string) {
+  const { WASI } = await import("@antonz/runno");
+  const url = "https://unpkg.com/@antonz/python-wasi/dist/python.wasm";
```
humphd, line 120: Can we get this from node_modules vs. unpkg.com?  Vite allows you to get a URL for a WASM https://vitejs.dev/guide/features#accessing-the-webassembly-module

humphd (changes requested):
This is amazing, such great work.

tarasglek:
thats exciting, lets make the run button run it instead of showing menu when no server option

tarasglek:
love how small this change is

src/lib/run-code.ts
WangGithub0: I change to use node_modules and can build successfully on my local, but fail in [CloudFlare](https://dash.cloudflare.com/?to=/:account/pages/view/chatcraft-org/2ae02daa-e807-47a9-80e1-858c95ad4f84), could you help me check the log on CloudFlare

humphd:
```

22:37:22.763 | Deploying your site to Cloudflare's global network...
-- | --
22:37:26.503 |  
22:37:26.618 | ✘ [ERROR] Error: Pages only supports files up to 26.2 MB in size
22:37:26.619 |  
22:37:26.619 | assets/python-vjebDg3S.wasm is 26.3 MB in size

```

So we're going to have to load it from a CDN.  I guess go back to using unpkg.com.

humphd (approved):

kliu57 (approved):
I tested it and it works on running python code from both user's own python code and chatgpt's reply python code. Only thing I found was it will output an error if you try to run any python code that is asking for user input (because obviously we have no terminal to take in user's input I suppose)

Maybe something for future follow up PRs could be to print more details error messages when such fails occur. Right now it just says something like this

```
Error: Script execution failed

ySe/</</<@https://60283562.console-overthinker-dev.pages.dev/assets/index-1LG397X1.js:579:1095
```

Issue #365, opened 2024-01-25
Investigate bundle size and use more dynamic imports


humphd:
If you run `pnpm build` we generate `stats.html`, which is bundle size info.  Here's what it looks like at the moment:

<img width="630" alt="Screenshot 2024-01-25 at 7 40 10 AM" src="https://github.com/tarasglek/chatcraft.org/assets/427398/a6e997fe-9427-4c23-bf67-7f07a3c6903f">

I've highlighted a few large modules we should consider [importing dynamically](https://vitejs.dev/guide/features#dynamic-import) vs. statically.  We've already done this for a few large modules, see https://github.com/search?q=repo%3Atarasglek%2Fchatcraft.org%20await%20import&type=code.

The modules highlighted above are for features of Mermaid diagrams and Katex math rendering.  Both of these are used in special cases, so we should consider loading the modules only when needed vs. every time.

tarasglek:
love this...would also love to produce this as asset in github actions

humphd:
Sure, we could put this `stats.html` page + Vite test coverage HTML on GitHub Pages or something.

rjwignar:
When attempting to fix this I identified three large modules for dynamic importing:
1. **elkjs** (for Mermaid)
2. **cytoscape** (for Mermaid)
3. **katex** (for Math rendering)

The **elkjs** and **cytoscape** modules are _both_ dependencies of **mermaid**:

The **katex** module is a dependecncy of **two modules**:
- **rehype-katex**:
- and **micromark-extension-math** (not directly imported in code):
...which itself is a dependency of **remark-math** (directly imported in code)

Full explanation available at https://github.com/tarasglek/chatcraft.org/pull/472

My findings:
- **elkjs** and **cytoscape** can't be feasibly dynamically imported. At least with what I've tried, these can only be dynamically imported if mermaid is dynamically imported, but the mermaid package is always required., so it's not useful to dynamically import this. If there is a way to somehow dynamically import these dependencies from within mermaid

- I tried dynamically importing the **rehype-katex** and **remark-math** packages in #472. These changes seem to work locally, but when deployed on the [preview](https://rjwignar-issue-365.console-overthinker-dev.pages.dev/) but there is a workbox-related issue preventing the app from loading. This can be revisited in the future if we can figure out the workbox-related issue.

PR #488, opened 2024-03-08
Add Dynamic Imports for Mermaid

rjwignar:
This is a partial fix for #365

Summary
---

`src/components/MermaidPreview.tsx` **and** `src/components/MessagesView.tsx`: Changed the following module to dynamic import:
- **mermaid** (directly depends on `elkjs` and `cytoscape`)

Explanation
---

#365  identified three large modules for dynamic importing:
1. **elkjs** (for Mermaid)
2. **cytoscape** (for Mermaid)
3. **katex** (for Math rendering)

The **elkjs** and **cytoscape** modules are _both_ dependencies of **mermaid**:
https://github.com/tarasglek/chatcraft.org/blob/ca5ef709661bfc7415638d1708a1e539d7a858aa/pnpm-lock.yaml#L7688-L7701

Dynamically importing the mermaid package will load these packages only when they're needed.

### DIFF
--- a/src/components/MessagesView.tsx
+++ b/src/components/MessagesView.tsx
@@ -1,6 +1,5 @@
 import { useCallback, useLayoutEffect, useMemo } from "react";
 import { Box, useColorMode } from "@chakra-ui/react";
-import mermaid from "mermaid";
 
 import Message from "./Message";
 import NewMessage from "./Message/NewMessage";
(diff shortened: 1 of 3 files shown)


src/components/MessagesView.tsx
```diff
@@ -44,11 +43,15 @@ function MessagesView({
   // Make sure that any Mermaid diagrams use the same light/dark theme as rest of app.
   // Use a layout effect vs. regular effect so it happens after DOM is ready.
   useLayoutEffect(() => {
-    mermaid.initialize({
-      startOnLoad: false,
-      theme: colorMode === "dark" ? "dark" : "default",
-      securityLevel: "loose",
-    });
+    const initializeMermaid = async () => {
+      const mermaid = await import("mermaid");
```
humphd, line 47: I don't think this is going to gain us much, since MessagesView is always going to be included.  We pretty much have to always include Mermaid.

rjwignar: When working on this I noticed that cytoscape and elkjs are dependencies of Mermaid and are never directly called in the code base (at least based on what `git grep` reveals).
From my understanding, the only way I could dynamically import cytoscape and elkjs was to dynamically import the only package that depends on them.

Is there another approach to dynamically import only those dependencies?

PR #492, opened 2024-03-13
Add Ruby in run-code

WangGithub0:
image


tarasglek:
lgtm, love how easy these are to add

src/lib/run-code.ts
```diff
@@ -148,6 +158,39 @@ async function runPython(code: string) {
   return executionResult;
 }
 
+async function runRuby(code: string) {
```
humphd, line 161: This is nearly identical to the python code, can we refactor to share common bits?

src/lib/run-code.ts
```diff
@@ -148,6 +158,39 @@ async function runPython(code: string) {
...
+        stdout: (out) => {
+          console.log(out);
+        },
+        stderr: (err) => {
+          console.error(err);
+        },
+      })
+        .then((result) => {
+          if (result.exitCode === 0) {
+            resolve(); // Resolves the promise with no value
```
humphd, line 179: This comment is not needed.

src/lib/run-code.ts
```diff
@@ -148,6 +158,39 @@ async function runPython(code: string) {
...
+          } else {
+            reject(new Error("Script execution failed")); // Rejects the promise if execution failed
```
humphd, line 181: Same here.

src/lib/run-code.ts
```diff
@@ -148,6 +158,39 @@ async function runPython(code: string) {
...
+          }
+        })
+        .catch((error) => {
+          reject(error); // Reject on error
```
humphd, line 185: And here.  Let the code speak for itself if you have nothing more to add.

humphd (changes requested):
This is really clean.  A few things to figure out.

src/lib/run-code.ts
```diff
@@ -115,15 +125,25 @@ async function runJavaScript(code: string) {
   }
 }
 
-async function runPython(code: string) {
+async function runPythonOrRuby(code: string, language: string) {
```
humphd, line 128: `runInWasi()` would be a better name.

humphd (changes requested):
Looks good, one thing to fix.

kliu57 (approved):
Tested it in browser, it works as expected!

Issue #124, opened 2023-06-22
Consider supporting Dall-E for image generation in chats


humphd:
I can't decide if this is useful or not, but it's doable to support sending prompts to Dall-E for [image generation](https://platform.openai.com/docs/guides/images/image-generation-beta).

The way we'd do it is that you'd enter a prompt, and select "Ask Dall-E" instead of the GPT models.  The code that does the call to OpenAI would need to inspect the model and decide which endpoint to use (normally we use `chat()` but we'd need to call the `openai.createImage()` stuff).

We'd need to create two new things:

1. `ChatCraftImageMessage`, which would store a list of URLs in JSON for the text, or maybe we get the images as data URIs and embed them in the db
2. `ImageMessage` component, which would map the image URLs into `<img>` elements in a grid.

The `ImageMessage` would also need some kind of progress indicator to take the place of the streaming response, while we wait for the model to generate the images and send.

After that you could ask for images and get them back in the current chat.  Dall-E isn't useful for things like diagrams or technical images; however, if you're needing a logo, or sprites, or background graphics to use in an app, this might be useful?

PR #493, opened 2024-03-13
Add: Generate Image by DALL·E 3

mingming-ma:
## Description
The function `generateImage` calls the [`dall-e-3`](https://platform.openai.com/docs/guides/images/image-generation) model api to get results. Add a new command `image` on user side to display the image generated. 

## How to test
Enter `prompt` after command `image` in the input area: `/image ${prompt}`
Example: `/image Happy White Border Collie. Close up`

Fixes: #124

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -401,6 +401,8 @@ type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
 
 export const generateImage = async (
   prompt: string,
+  //You can request 1 image at a time with DALL·E 3 (request more by making parallel requests) or up to 10 images at a time using DALL·E 2 with the n parameter.
+  //https://platform.openai.com/docs/guides/images/generations
   n: number = 1,
   size: dalle3ImageSize = "1024x1024"
 ): Promise<string[]> => {
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -433,9 +433,9 @@ export const generateImage = async (
   try {
     const response = await openai.images.generate({
       model: "dall-e-3",
-      prompt: prompt,
-      n: n,
-      size: size,
+      prompt,
+      n,
+      size,
     });
 
     const imageUrls = response.data.map((img: any) => img.url);
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -411,7 +411,7 @@ export const generateImage = async (
     throw new Error("Missing OpenAI API Key");
   }
 
-  const { openai } = createClient(currentProvider?.apiKey, currentProvider?.apiUrl);
+  const { openai } = currentProvider.createClient(currentProvider.apiKey);
 
   try {
     const response = await openai.images.generate({
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -419,7 +419,6 @@ export const generateImage = async (
       size: size,
     });
 
-    // Assuming the response structure has an array of generated images
     const imageUrls = response.data.map((img: any) => img.url);
     return imageUrls;
   } catch (error: any) {
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -396,3 +396,33 @@ export const textToSpeech = async (message: string): Promise<string> => {
 
   return objectUrl;
 };
+
+type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
+
+export const generateImage = async (
+  prompt: string,
+  n: number = 1,
+  size: dalle3ImageSize = "1024x1024"
+): Promise<string[]> => {
+  const { currentProvider } = getSettings();
+  if (!currentProvider?.apiKey) {
+    throw new Error("Missing OpenAI API Key");
+  }
+
+  const { openai } = createClient(currentProvider?.apiKey, currentProvider?.apiUrl);
+
+  try {
+    const response = await openai.images.generate({
+      model: "dall-e-3",
+      prompt: prompt,
+      n: n,
+      size: size,
+    });
+
+    // Assuming the response structure has an array of generated images
+    const imageUrls = response.data.map((img: any) => img.url);
+    return imageUrls;
+  } catch (error: any) {
+    throw new Error(`Failed to generate image: ${error.message}`);
+  }
+};
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -397,6 +397,23 @@ export const textToSpeech = async (message: string): Promise<string> => {
   return objectUrl;
 };
 
+/**
+ * Only meant to be used outside components or hooks
+ * where useModels cannot be used.
+ */
+export async function isGenerateImageSupported() {
+  const { currentProvider } = getSettings();
+  if (!currentProvider.apiKey) {
+    throw new Error("Missing API Key");
+  }
+
+  return (
+    (await currentProvider.queryModels(currentProvider.apiKey)).filter((model: string) =>
+      model.includes("dall-e-3")
+    )?.length > 0
+  );
+}
+
 type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
 
 export const generateImage = async (
@@ -407,7 +424,7 @@ export const generateImage = async (
   size: dalle3ImageSize = "1024x1024"
 ): Promise<string[]> => {
   const { currentProvider } = getSettings();
-  if (!currentProvider?.apiKey) {
+  if (!currentProvider.apiKey) {
     throw new Error("Missing OpenAI API Key");
   }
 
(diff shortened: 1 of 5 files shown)


tarasglek:
we need to change code so commands added automatically show up in /commands atm this command is impossible to discover 
image

tarasglek:
also really confusing that /dalle3 doesn't work

humphd:
I think it should be `/image ...`

src/lib/ai.ts
```diff
@@ -468,3 +468,32 @@ export const textToSpeech = async (message: string): Promise<string> => {
 
   return objectUrl;
 };
+
+type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
+
+export const generateImage = async (
+  prompt: string,
+  n: number = 1,
+  size: dalle3ImageSize = "1024x1024"
```
rjwignar, line 477: I noticed in the generateImage() call in src/lib/commands/ImageCommand.ts, we only pass the prompt.
I wonder if in the future, we can/should allow the user to select the image size as well.

rjwignar (commented):
One thing I noticed is that once the chat exceeds the screen height, the scrollbar doesn't automatically move down to show the user the generated image:
dalle3Scrollbar

However, this is something to be investigated in a follow-up, and **should not block** this from being merged.

rjwignar (approved):
@mingming-ma These changes look good! It was a good idea to rename the command to "/image".

At the time of my review, @tarasglek and @humphd's suggestions have been addressed and incorporated into this PR:
image

I left ~3 observations/concerns that are either non-issues or can be addressed in follow-up PRs.
They shouldn't prevent this from being merged. @humphd do you agree?

src/lib/ai.ts
```diff
@@ -468,3 +468,32 @@ export const textToSpeech = async (message: string): Promise<string> => {
...
+      model: "dall-e-3",
+      prompt: prompt,
+      n: n,
+      size: size,
+    });
+
+    const imageUrls = response.data.map((img: any) => img.url);
+    return imageUrls;
+  } catch (error: any) {
+    throw new Error(`Failed to generate image: ${error.message}`);
```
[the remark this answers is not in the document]
tarasglek: i think we should do a check to see if dalle3 is among available models and provide a erro message in this PR

tarasglek:
I think we should also try to store the image similarly to how we upload images into chatcraft, so we can then ask gpt vision to analyze it

src/lib/ai.ts
```diff
@@ -468,3 +468,32 @@ export const textToSpeech = async (message: string): Promise<string> => {
...
+  prompt: string,
+  n: number = 1,
+  size: dalle3ImageSize = "1024x1024"
+): Promise<string[]> => {
+  const { currentProvider } = getSettings();
+  if (!currentProvider?.apiKey) {
+    throw new Error("Missing OpenAI API Key");
+  }
+
+  const { openai } = createClient(currentProvider?.apiKey, currentProvider?.apiUrl);
```
Amnish04, line 484: I don't think we need a `.?` here since we are already doing the `currentProvider.apiKey` check above.

I'll fix my `textToSpeech` function as well.

Amnish04 (changes requested):
@mingming-ma This is a pretty cool feature. Just left a few comments after reviewing.

Amnish04:
@mingming-ma This is unrelated to this PR, but I feel like the images should have a `cursor: pointer` for the user to know something happens on clicking them.

Can we make that change in this PR or a follow up?

src/lib/ai.ts
humphd: FYI, @kliu57 is changing how this works right now, too.

mingming-ma:
> I think we should also try to store the image similarly to how we upload images into chatcraft, so we can then ask gpt vision to analyze it(this can be done in followup)

I filed an issue to fix store problem #496 , just want to inform you that at the moment user is able use vision to analyze the generated image

src/lib/commands/ImageCommand.ts
```diff
@@ -0,0 +1,27 @@
...
+export class ImageCommand extends ChatCraftCommand {
+  constructor() {
+    super("image");
+  }
+
+  async execute(chat: ChatCraftChat, user: User | undefined, args?: string[]) {
+    if (!(args && args[0])) {
+      throw new Error("must include a prompt");
+    }
+    const prompt = args.join(" ");
```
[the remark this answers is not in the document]
mingming-ma: Filed https://github.com/tarasglek/chatcraft.org/issues/499 to track

src/lib/ai.ts
mingming-ma: Filed https://github.com/tarasglek/chatcraft.org/issues/499 to track

src/lib/ai.ts
mingming-ma: I'll follow up the changes to see how to fix this

mingming-ma:
> One thing I noticed is that once the chat exceeds the screen height, the scrollbar doesn't automatically move down to show the user the generated image:

@rjwignar I just saw @Amnish04 fixed an issue related the auto scroll, https://github.com/tarasglek/chatcraft.org/pull/495, I don't know it is related or not, let's see how it is after this be merged, sounds good?

Amnish04:
> @rjwignar I just saw @Amnish04 fixed an issue related the auto scroll, #495, I don't know it is related or not, let's see how it is after this be merged, sounds good?
@mingming-ma I just undid that fix in #497 as that was causing nested submenus to break (more details [here](https://github.com/tarasglek/chatcraft.org/pull/497#discussion_r1524153514)). But nonetheless, I don't think that was related to this issue

src/lib/ai.ts
```diff
@@ -468,3 +468,32 @@ export const textToSpeech = async (message: string): Promise<string> => {
 
   return objectUrl;
 };
+
+type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
+
+export const generateImage = async (
```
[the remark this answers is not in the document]
mingming-ma: Thanks for the suggestion, I added a comment to doc this and also let dev know why `n` is `1` which makes more sense.

src/lib/ai.ts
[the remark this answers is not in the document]
mingming-ma: Fixed at [e61b83a](https://github.com/tarasglek/chatcraft.org/pull/493/commits/e61b83ab02f21bfca8b1114116fab0983a06b365)

src/lib/ai.ts
```diff
@@ -396,3 +396,51 @@ export const textToSpeech = async (message: string): Promise<string> => {
...
+  if (!currentProvider.apiKey) {
+    throw new Error("Missing OpenAI API Key");
+  }
+
+  const { openai } = currentProvider.createClient(currentProvider.apiKey);
+
+  try {
+    const response = await openai.images.generate({
+      model: "dall-e-3",
+      prompt: prompt,
```
humphd, line 436: nit: when you have the same key/value names, only include it once:

```ts
{
  model: "dall-e-e",
  prompt,
  n,
  size,
}
```

humphd (approved):

humphd:
In a follow-up, I wonder if we should add a menu item to the Options menu to allow you to do the same thing, but pop-up a little model where you type your prompt?  Or maybe make it so entering a prompt and asking the dall-e model does this too?

mingming-ma:
> In a follow-up, I wonder if we should add a menu item to the Options menu to allow you to do the same thing, but pop-up a little model where you type your prompt? Or maybe make it so entering a prompt and asking the dall-e model does this too?

@humphd Just guessing the main concern here, let user know this feature more easily?

humphd:
> > In a follow-up, I wonder if we should add a menu item to the Options menu to allow you to do the same thing, but pop-up a little model where you type your prompt? Or maybe make it so entering a prompt and asking the dall-e model does this too?
> 
> @humphd Just guessing the main concern here, let user know this feature more easily? Maybe we could also add the `Commands` in the `Options` menu, and list them as submenus.

Lots of users will never use/know about `/slash` commands.  It's pretty advanced.

Amnish04:
@mingming-ma Do we have an issue to track this https://github.com/tarasglek/chatcraft.org/pull/493#issuecomment-1998092723

mingming-ma:
> @mingming-ma Do we have an issue to track this [#493 (comment)](https://github.com/tarasglek/chatcraft.org/pull/493#issuecomment-1998092723)

@Amnish04 Just filed https://github.com/tarasglek/chatcraft.org/issues/504 to track this

mingming-ma:
I think we can merge this PR for now, @humphd @Amnish04 I'll create another issue follow up the  [react-medium-image-zoom](https://www.npmjs.com/package/react-medium-image-zoom)

Amnish04 (approved):

PR #495, opened 2024-03-13
Cap submenu maxHeight to 50vh to prevent weird autoscroll

### LINKED ISSUE #452: Find a better fix for Menu Positioning
We've encountered a critical UI breaking bug a couple of times now, where a Chakra Menu , when placed near the bottom of the screen pushes the entire viewport upwards on any routing operation.

The reason for this happening is documented in the following blog post:
https://github.com/tarasglek/chatcraft.org/issues/442

Recently, it resurfaced in #442 and I fixed it using the `strategy=fixed` hack in #445 .

However, we probably need to look for a better fix so to prevent it from ever happening again.

Eventually, we might have to replace all the Chakra Menus with `React Menu` for submenus, and we're not sure if that also causes this problem. Since even `React Menu` has its own set of issues right now, we might have to find a better alternative if we can't get around those.

So I think its better to wait and watch where it goes, and look into this when we have a stable menu component integrated with the app.

### LINKED ISSUE #454: AutoScroll in Firefox with Long Submenus
When we trigger a submenu with a lot of options, it causes a weird Autoscroll as demonstrated below.

SubmenuAutoscroll

**Note:** We could only reproduce this issue in Firefox.

Amnish04:
This has been an issue for a long time. When we have a lot of items in a submenu, we get a weird autoscroll that tried to focus the first element of the list (which was not able to fit in the available viewport height).

I tried to fix this in #487, but it had a subtle mistake where I set `maxHeight` to `50vw` instead of `50vh` due to which the bug wasn't fixed properly.

I have corrected that mistake and the results are much better now.
image

This fixes #454

### DIFF
--- a/src/components/Menu/SubMenu.tsx
+++ b/src/components/Menu/SubMenu.tsx
@@ -1,18 +1,11 @@
-import { Box } from "@chakra-ui/react";
 import { SubMenu as ReactMenuSubMenu, type SubMenuProps } from "@szhsin/react-menu";
 import React from "react";
 
 const SubMenu: React.FC<SubMenuProps> = (props) => {
   return (
-    <Box>
-      <ReactMenuSubMenu
-        menuStyle={{ maxHeight: "50vw", overflow: "auto" }}
-        align="center"
-        {...props}
-      >
-        {props.children}
-      </ReactMenuSubMenu>
-    </Box>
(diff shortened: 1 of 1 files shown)


WangGithub0 (approved):
Looks good to me

PR #497, opened 2024-03-14
Provide TTS voice options, speak and download features

### LINKED ISSUE #400: Make TTS more flexible
1) Add a Speak submenu to both human and bot messages
2) In the Speak submenu allow one to select voice, which causes message to be spoken,
3) once message is spoke, allow generated message to be downloaded..can add a download to the speak submenu after speech completes. It's ugly but not sure re other options
4) Changing voice in speech menu also changes it for the speak icon when automatic tts is on
5) should move speak icon to the menu on the left and change it to be "options" or 3 vertical dots on mobile

Amnish04:
After finishing up with all the follow ups filed for TTS, I've extended on the existing functionality as requesting in issue #400 

I have:
1. Updated the **settings** model with an additional property of type `TextToSpeechSettings`
```ts
type TextToSpeechSettings = {
  announceMessages?: boolean;
  voice: TextToSpeechVoices;
};

export enum TextToSpeechVoices {
  ALLOY = "alloy",
  ECHO = "echo",
  FABLE = "fable",
  ONYX = "onyx",
  NOVA = "nova",
  SHIMMER = "shimmer",
}
```
2. Added a **Download as Audio** submenu with a list of **available voices** to choose for download.
image
3. Added an  **Audio Submenu** with 2 options - **Speak** and **Select Voice**.
image
The **Speak** option announces the message content in the voice selected in the **Select Voice** setting. The **Select Voice** setting also applies to the automatic TTS streaming that occurs while the response is being generated.

**Important:** The **Select Voice** menu should ideally be a submenu in the **Option Button** in the prompt form. But since we still have to replace it with the new Menu Component to allow for submenus, I have temporarily placed it in Message Base for now.

This closes #400


src/components/Menu/SubMenu.tsx
```diff
@@ -3,7 +3,7 @@ import React from "react";
 
 const SubMenu: React.FC<SubMenuProps> = (props) => {
   return (
-    <ReactMenuSubMenu menuStyle={{ maxHeight: "50vh", overflow: "auto" }} align="center" {...props}>
+    <ReactMenuSubMenu align="center" {...props}>
```
Amnish04, line 6: I realized this doesn't work well with **React Menu** as it considers its height to be an aggregate height of self and any nested submenus breaking the nested menus beyond 1 layer.

humphd:
I think you should move the voice selection to the Settings vs. this sub-menu.  You're not going to change the voice often enough to need this extra complexity.

Amnish04:
@humphd Makes sense, I've moved the voice selection to Preferences modal, and also added a Preview Audio button.
image

src/components/Message/MessageBase.tsx
```diff
@@ -316,6 +348,21 @@ function MessageBase({
...
+  const { clearAudioQueue, addToAudioQueue } = useAudioPlayer();
+
+  const handleSpeakMessage = useCallback(
+    async (messageContent: string) => {
+      // Stop any currently playing audio before starting new
+      clearAudioQueue();
+
+      const { voice } = settings.textToSpeech;
+
+      // Use lighter tts-1 model when not to minimize latency
```
humphd, line 360: This comment is confusing, typo?

humphd (changes requested):
Tested this tonight, works amazing!  The code looks great.

Let's remove this sub-menu of voices, since we have it in the settings.  We only need one place to do it.

Screenshot 2024-03-14 at 8 25 22 PM

src/components/Message/MessageBase.tsx
Amnish04: Yes, that's a typo. Just pushed the fix

Amnish04:
> Let's remove this sub-menu of voices, since we have it in the settings. We only need one place to do it.
> Screenshot 2024-03-14 at 8 25 22 PM

I haven't removed the submenu because users might want to download using a specific voice while on the go. In such cases, constantly navigating back and forth between the settings modal could be cumbersome for the user. What do you think?

Amnish04:
@humphd I realized I wasn't handling errors when calling `textToSpeech`. Just made some updates

humphd:
> I haven't removed the submenu because users might want to download using a specific voice while on the go. In such cases, constantly navigating back and forth between the settings modal could be cumbersome, especially when they just want to use a different voice for a particular download and not affect the normal setting.
> What do you think?

I think you should remove it :). The reason is that you are too deep into this feature and know too much.  Most people are going to set the voice once, and never think about it again.  They won't switch back and forth all the time, and forcing them to pick a voice when all they want is audio over complicates things.

We have a way to change the voice via Settings, we don't then need to sprinkle that throughout the rest of the UI.

tarasglek:
Very cool, I have been wanting this feature forever to generate voice-overs for demos.

I think there is only one case where would wanna switch voice and that's dialog generation, but lets wait for someone to ask us to implement it.

Amnish04:
> I think you should remove it :). The reason is that you are too deep into this feature and know too much. Most people are going to set the voice once, and never think about it again. They won't switch back and forth all the time, and forcing them to pick a voice when all they want is audio over complicates things.
> 
> We have a way to change the voice via Settings, we don't then need to sprinkle that throughout the rest of the UI.

@humphd That also makes sense. Just pushed the changes, now I am using the voice configured in settings for downloads as well

tarasglek:
Is there a limit on length of text one can download?

Amnish04:
> Is there a limit on length of spoken audio one can download?

There isn't. I was also thinking about this, but not sure what's the best way to handle.

tarasglek:
Can we land this, I wanna use this feature to record youtubes :)

Amnish04:
@humphd Are you also okay with shipping this? I've made the changes you requested

humphd:
@Amnish04 absolutely, get this code launched 🚀

PR #498, opened 2024-03-14
Add FreeModelProvider and more provider refactoring

### LINKED ISSUE #401: Do not require chatcraft users to enter a key to use it
If a user does not enter a key, we should use a free model endpoint provided by something like https://taras-free_open_router.web.val.run implemented by https://www.val.town/v/taras/free_open_router .  Should still have a banner prompting for user to enter their key while in limited to free models mode.
This makes use of a new openrouter feature where free models are singled out with :free suffix

kliu57:
Added new provider (free provider for new users to try)
Further providers refactoring

Closes #401

### Code changes:

- Moved `validateApiKey`, `queryModels`, `createClient` out of `ais.ts` and into `ChatCraftProvider.ts`
- Moved `usingOfficialOpenAI`, `usingOfficialOpenRouter` out of `ai.ts` and into `providers/index.ts`
- Added new free provider `providers/DefaultProvider/FreeModelProvider.ts`
- Configured `ModelAvatar.tsx` for `DefaultProvider` so logo will display with the green bg

### Potential future follow up items:

- Make FreeModelProvider the default provider for new users (this will make the Instructions page obsolete so we need to design a new way to instruct new users about ChatCraft)
- Query the first available free provider in the list of providers instead of hardcoding (currently not sure how since requires making `defaultModelForProvider()` async and affecting alot of other files)

### DIFF
--- a/src/Chat/ChatHeader.tsx
+++ b/src/Chat/ChatHeader.tsx
@@ -154,7 +154,7 @@ function ChatHeader({ chat }: ChatHeaderProps) {
                       {title}
                     </Link>
                   </Text>
-                  {settings.currentProvider?.apiKey && (
+                  {settings.currentProvider.apiKey && (
                     <IconButton
                       variant="ghost"
                       size="sm"
@@ -179,7 +179,7 @@ function ChatHeader({ chat }: ChatHeaderProps) {
                 <MenuItem onClick={() => handleCopyChatClick()}>Copy</MenuItem>
                 <MenuItem onClick={() => handleDownloadClick()}>Download</MenuItem>
 
-                {!chat.readonly && settings.currentProvider?.apiKey && (
+                {!chat.readonly && settings.currentProvider.apiKey && (
                   <>
                     <MenuDivider />
                     <MenuItem onClick={onOpen}>Share Chat...</MenuItem>
--- a/src/components/Message/AppMessage/Instructions.tsx
+++ b/src/components/Message/AppMessage/Instructions.tsx
@@ -17,7 +17,7 @@ import { providerFromUrl, providerFromJSON, getSupportedProviders } from "../../
 import { OpenRouterProvider } from "../../../lib/providers/OpenRouterProvider";
 import RevealablePasswordInput from "../../RevealablePasswordInput";
 import { useSettings } from "../../../hooks/use-settings";
-import { validateApiKey } from "../../../lib/ai";
+import { FreeModelProvider } from "../../../lib/providers/DefaultProvider/FreeModelProvider";
 
 const ApiKeyInstructionsText = `## Getting Started with ChatCraft
 
@@ -71,12 +71,13 @@ function Instructions(props: MessageBaseProps) {
 
     // See if this API Key is valid
     setIsValidating(true);
-    validateApiKey(apiKey)
+    settings.currentProvider
+      .validateApiKey(apiKey)
       .then((valid) => {
         if (valid) {
           setIsInvalid(false);
 
-          const newProvider = providerFromUrl(settings.currentProvider?.apiUrl, apiKey.trim());
+          const newProvider = providerFromUrl(settings.currentProvider.apiUrl, apiKey.trim());
 
           setSettings({
             ...settings,
@@ -104,10 +105,18 @@ function Instructions(props: MessageBaseProps) {
         ? providerFromJSON(settings.providers[e.target.value])
         : providerFromUrl(e.target.value);
 
-      setSettings({
-        ...settings,
-        currentProvider: newProvider,
-      });
+      if (newProvider instanceof FreeModelProvider) {
+        // If user chooses the free provider, set the key automatically
+        setSettings({
+          ...settings,
+          currentProvider: new FreeModelProvider(),
+        });
+      } else {
+        setSettings({
+          ...settings,
+          currentProvider: newProvider,
+        });
+      }
     },
     [setSettings, settings]
   );
@@ -123,7 +132,7 @@ function Instructions(props: MessageBaseProps) {
             <FormControl>
               <FormLabel>Provider API URL</FormLabel>
 
-              <Select value={settings.currentProvider?.apiUrl} onChange={handleProviderChange}>
+              <Select value={settings.currentProvider.apiUrl} onChange={handleProviderChange}>
                 {Object.values(supportedProviders).map((provider) => (
                   <option key={provider.apiUrl} value={provider.apiUrl}>
                     {provider.name} ({provider.apiUrl})
@@ -133,7 +142,7 @@ function Instructions(props: MessageBaseProps) {
             </FormControl>
 
             <FormControl isInvalid={isInvalid}>
-              <FormLabel>{settings.currentProvider?.name} API Key </FormLabel>
+              <FormLabel>{settings.currentProvider.name} API Key </FormLabel>
               <Flex gap={4} align="center">
                 <RevealablePasswordInput
                   flex="1"
@@ -148,12 +157,12 @@ function Instructions(props: MessageBaseProps) {
                 </Button>
               </Flex>
               {settings.currentProvider instanceof OpenRouterProvider && (
-                <Button mt="3" size="sm" onClick={OpenRouterProvider.openRouterPkceRedirect}>
+                <Button mt="3" size="sm" onClick={settings.currentProvider.openRouterPkceRedirect}>
                   Get API key from OpenRouter{" "}
                 </Button>
               )}
               <FormErrorMessage>
-                Unable to verify API Key with {settings.currentProvider?.name}.
+                Unable to verify API Key with {settings.currentProvider.name}.
               </FormErrorMessage>
             </FormControl>
           </VStack>
--- a/src/lib/ChatCraftProvider.ts
+++ b/src/lib/ChatCraftProvider.ts
@@ -1,13 +1,17 @@
 import { nanoid } from "nanoid";
+import OpenAI from "openai";
+import { ChatCraftModel } from "./ChatCraftModel";
 
-export type ProviderName = "OpenAI" | "OpenRouter.ai";
+export type ProviderName = "OpenAI" | "OpenRouter.ai" | "Free AI Models";
 
 export const OPENAI_API_URL = "https://api.openai.com/v1";
 export const OPENROUTER_API_URL = "https://openrouter.ai/api/v1";
+export const FREEMODELPROVIDER_API_URL = "https://taras-free_open_router.web.val.run/api/v1";
 
 const urlToNameMap: { [key: string]: ProviderName } = {
   [OPENAI_API_URL]: "OpenAI",
   [OPENROUTER_API_URL]: "OpenRouter.ai",
+  [FREEMODELPROVIDER_API_URL]: "Free AI Models",
 };
 
 export interface ProviderData {
@@ -33,4 +37,48 @@ export abstract class ChatCraftProvider {
     this.apiUrl = url;
     this.apiKey = key;
   }
+
+  createClient(key: string) {
+    return {
+      openai: new OpenAI({
+        apiKey: key,
+        baseURL: this.apiUrl,
+        defaultHeaders: this.clientHeaders,
+        dangerouslyAllowBrowser: true,
+      }),
+      headers: this.clientHeaders,
+    };
+  }
+
+  get clientHeaders() {
+    return {};
+  }
+
+  get logoUrl() {
+    // If we don't know, use the OpenAI logo as a fallback
+    return "/openai-logo.png";
+  }
+
+  async queryModels(key: string) {
+    if (!this.apiUrl) {
+      throw new Error("Missing API Url");
+    }
+
+    const { openai } = this.createClient(key);
+
+    try {
+      const models = [];
+      for await (const page of openai.models.list()) {
+        models.push(page);
+      }
+
+      return models.map((model: any) => model.id) as string[];
+    } catch (err: any) {
+      throw new Error(`error querying models API: ${err.message}`);
+    }
+  }
+
+  abstract validateApiKey(key: string): Promise<boolean>;
+
+  abstract defaultModelForProvider(): ChatCraftModel;
 }
--- a/src/components/Message/AiMessage.tsx
+++ b/src/components/Message/AiMessage.tsx
@@ -96,7 +96,7 @@ function AiMessage(props: AiMessageProps) {
 
   const handleRetryClick = useCallback(
     async (model: ChatCraftModel) => {
-      if (!settings.currentProvider?.apiKey) {
+      if (!settings.currentProvider.apiKey) {
         return;
       }
 
@@ -130,7 +130,7 @@ function AiMessage(props: AiMessageProps) {
         setRetrying(false);
       }
     },
-    [props.chatId, settings.currentProvider?.apiKey, message, callChatApi]
+    [props.chatId, settings.currentProvider.apiKey, message, callChatApi]
   );
 
   // While we're streaming in a new version, use a different display
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -60,10 +60,10 @@ import { useCopyToClipboard } from "react-use";
 // Styles for the message text are defined in CSS vs. Chakra-UI
 import "./Message.css";
 import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
-import { usingOfficialOpenAI } from "../../lib/ai";
 import { ChatCraftChat } from "../../lib/ChatCraftChat";
 import { useLiveQuery } from "dexie-react-hooks";
 import { useUser } from "../../hooks/use-user";
+import { usingOfficialOpenAI } from "../../lib/providers";
 
 export interface MessageBaseProps {
   message: ChatCraftMessage;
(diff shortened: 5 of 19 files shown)


kliu57:
@tarasglek Just wondering if you have ever encountered this CORS issue I'm facing

tarasglek:
Seems to work here, except it doesn't provide full choice of models..

src/lib/providers/FreeOpenRouterProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+  apiKey?: string;
+};
+
+export class FreeOpenRouterProvider extends ChatCraftProvider {
+  constructor(key?: string) {
+    super(FREEOPENROUTER_API_URL, key);
+  }
+
+  get defaultModel() {
+    return "openchat/openchat-7b:free";
```
tarasglek, line 22: should just pick first available model instead of hardcoding

src/components/Message/AppMessage/Instructions.tsx
```diff
@@ -71,7 +71,8 @@ function Instructions(props: MessageBaseProps) {
 
     // See if this API Key is valid
     setIsValidating(true);
-    validateApiKey(apiKey)
+    settings.currentProvider
+      ?.validateApiKey(apiKey)
```
humphd, line 75: Is it possible for us to get rid of the `undefined` in the `currentProvider` type? That is, can we make sure that we always return this free provider if nothing else is configured in the settings, and then not need to deal with `.currentProvider?`

src/lib/ChatCraftProvider.ts
```diff
@@ -1,13 +1,17 @@
 import { nanoid } from "nanoid";
+import OpenAI from "openai";
+import { ChatCraftModel } from "./ChatCraftModel";
 
-export type ProviderName = "OpenAI" | "OpenRouter.ai";
+export type ProviderName = "OpenAI" | "OpenRouter.ai" | "Demo Provider";
```
humphd, line 5: This new name should match what we call this thing.  I'd suggest "Default Provider"

src/lib/ChatCraftProvider.ts
```diff
@@ -33,4 +37,52 @@ export abstract class ChatCraftProvider {
...
+
+    try {
+      const models = [];
+      for await (const page of openai.models.list()) {
+        models.push(page);
+      }
+
+      return models.map((model: any) => model.id) as string[];
+    } catch (err: any) {
+      throw new Error(err.message ?? `error querying models API`);
```
humphd, line 81:
```suggestion
      throw new Error(`error querying models API: ${err.message}`);
```

src/lib/ChatCraftProvider.ts
```diff
@@ -33,4 +37,52 @@ export abstract class ChatCraftProvider {
     this.apiUrl = url;
     this.apiKey = key;
   }
+
+  createClient(key: string) {
```
humphd, line 41: Question: we store the key in this object, can we avoid needing to pass it in here and use `this.apiKey` instead?  Or, can we do:

```ts
createClient(key?: string) {
  key = key ?? this.apiKey;
  if (!key) {
    throw new Error('missing api key');
  }
  ...
}
```

Then you don't need to manage the key externally.

src/lib/providers/FreeOpenRouterProvider.ts
humphd: Your thing returns a list of models?

src/lib/providers/FreeOpenRouterProvider.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async queryModels(key: string) {
```
humphd, line 26: Again, I don't think we need this key arg, since it lives on the instance.

humphd (changes requested):

src/lib/providers/FreeOpenRouterProvider.ts
kliu57: @tarasglek How can I acquire a list of available models? I just tried to use our default open ai queryModels function to get a list of models for it but it just returned an empty array.

src/components/Message/AppMessage/Instructions.tsx
kliu57: Right now the currentProvider is never undefined actually, that question mark is just put in the code to avoid lint issues. Because currentProvider is automatically set to OpenAiProvider (with no key) when the user first visits ChatCraft.

If I set currentProvider to the free provider by default, then they won't encounter the Instructions page where they can input their Open Ai / OpenRouter API key anymore. Any new users will just skip automatically to:

image

src/lib/ChatCraftProvider.ts
```diff
@@ -33,4 +37,52 @@ export abstract class ChatCraftProvider {
...
+  get clientHeaders() {
+    return {};
+  }
+
+  get logoUrl() {
+    // If we don't know, use the OpenAI logo as a fallback
+    return "/openai-logo.png";
+  }
+
+  get defaultModel() {
```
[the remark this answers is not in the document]
kliu57: the getter (get defaultModel) returns a string, while defaultModelForProvider returns a ChatCraftModel object.

src/lib/ChatCraftProvider.ts
kliu57: @humphd The reason I kept the apiKey param is because in Instructions.tsx we only store in the apiKey in settings after it has been validated and not before. If I store the apiKey in settings and then validate it, then when the validation fails I need to add more code to re-set the settings to remove the stored apiKey.

src/components/Message/AppMessage/Instructions.tsx
```diff
@@ -17,7 +17,7 @@ import { providerFromUrl, providerFromJSON, getSupportedProviders } from "../../
 import { OpenRouterProvider } from "../../../lib/providers/OpenRouterProvider";
 import RevealablePasswordInput from "../../RevealablePasswordInput";
 import { useSettings } from "../../../hooks/use-settings";
-import { validateApiKey } from "../../../lib/ai";
+import { FreeOpenRouterProvider } from "../../../lib/providers/FreeOpenRouterProvider";
```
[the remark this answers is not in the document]
kliu57: I have made this change

src/components/Message/AppMessage/Instructions.tsx
```diff
@@ -104,10 +105,18 @@ function Instructions(props: MessageBaseProps) {
...
 
-      setSettings({
-        ...settings,
-        currentProvider: newProvider,
-      });
+      if (newProvider instanceof FreeOpenRouterProvider) {
+        // If user chooses the free provider, set the key automatically
+        setSettings({
+          ...settings,
+          currentProvider: new FreeOpenRouterProvider("mock_key"),
```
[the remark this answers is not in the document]
kliu57: I have made this change.

src/lib/ChatCraftProvider.ts
kliu57: I have made this change

src/lib/ChatCraftProvider.ts
kliu57: I have made this change

src/lib/providers/FreeOpenRouterProvider.ts
tarasglek: https://taras-free_open_router.web.val.run/api/v1/models

it's open source , capabilities are something one can get from source https://www.val.town/v/taras/free_open_router

kliu57:
@humphd Made the following changes according to your reviews:

- Renamed provider to DefaultProvider
- Change the DefaultProvider constructor to not need a key
- Made this code modification: throw new Error(`error querying models API: ${err.message}`);

As for the other unresolved conversations I have replied to them and am awaiting reply

src/components/Message/AppMessage/Instructions.tsx
humphd: But if TypeScript thinks it *could* be undefined, that means it might be at some point (e.g., these types are telling you that there is a bug).  We should make it so that `?.` isn't necessary, then we know it's good.

I think losing the instructions page is OK, since in a follow-up, we can update our logic to show some UI for adding your own provider/key if you are using the default provider.

src/lib/ChatCraftProvider.ts
humphd: Maybe key validation should be `static` and happen without creating the instance?

kliu57: @humphd I can make `validateApiKey` static instead but since the OpenAI `validateApiKey` calls `queryModels` which calls `createClient`, all 3 will have to become static. Is that ok?

kliu57: Also some of these functions require the apiUrl so once I make them static I will have to add back the code we were previously using to get the apiUrl. So maybe keeping it like this is better?

src/components/Message/AppMessage/Instructions.tsx
kliu57: Actually, I think we are good here. I removed the `?` from `currentProvider?.` and TypeScript did not give me any lint errors. We can talk about setting the default provider to the free one for new users and adding new UI after this release.

src/lib/ChatCraftProvider.ts
kliu57: > Maybe key validation should be `static` and happen without creating the instance?

I'm attempting to make `validateApiKey` static in `ChatCraftProvider` but we aren't allowed to make methods static and abstract according to lint. Should I go back to using our previous validateApiKey function (I can use the same code but put it in providers/index.ts intead)?

the code:

```
export async function validateApiKey(apiKey: string) {
  if (usingOfficialOpenAI()) {
    return OpenAiProvider.validateApiKey(apiKey);
  }

  if (usingOfficialOpenRouter()) {
    return OpenRouterProvider.validateApiKey(apiKey);
  }

  if (usingDefaultProvider()) {
    return true;
  }

  // If not either of those providers, something is wrong
  throw new Error("unexpected provider, not able to validate api key");
}
```

src/lib/providers/FreeOpenRouterProvider.ts
kliu57: I have successfully fetched the models

src/lib/providers/FreeOpenRouterProvider.ts
kliu57: I have successfully fetched the models

kliu57:
@tarasglek Just tested all the free models, they all return response except for mythomist-7b:free

image

src/components/Message/AppMessage/Instructions.tsx
```diff
@@ -17,7 +17,7 @@ import { providerFromUrl, providerFromJSON, getSupportedProviders } from "../../
 import { OpenRouterProvider } from "../../../lib/providers/OpenRouterProvider";
 import RevealablePasswordInput from "../../RevealablePasswordInput";
 import { useSettings } from "../../../hooks/use-settings";
-import { validateApiKey } from "../../../lib/ai";
+import { DefaultProvider } from "../../../lib/providers/DefaultProvider";
```
tarasglek, line 20: s/DefaultProvider/FreeModelProvider/

src/lib/providers/DefaultProvider.ts
```diff
@@ -0,0 +1,39 @@
+import { ChatCraftModel } from "../ChatCraftModel";
+import { ChatCraftProvider, DEFAULTPROVIDER_API_URL } from "../ChatCraftProvider";
+
+export class DefaultProvider extends ChatCraftProvider {
+  constructor() {
+    super(DEFAULTPROVIDER_API_URL, "mock_key");
+  }
+
+  get defaultModel() {
+    return "openchat/openchat-7b:free";
```
tarasglek, line 10: again, don't hard code model names, just return first model

src/lib/providers/DefaultProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async queryModels(key: string) {
+    const res = await fetch(`https://taras-free_open_router.web.val.run/api/v1/models`, {
```
tarasglek, line 15: don't hardcode url everywhere, have a baseurl in class and make everything relative to it

src/components/Message/AppMessage/Instructions.tsx
kliu57: @tarasglek I had previously named the new provider file `providers/FreeOpenRouterProvider.ts` but Dave wanted me to change it to `providers/DefaultProvider.ts`. Do you not like the name?

src/lib/ChatCraftProvider.ts
```diff
@@ -1,13 +1,17 @@
...
+export type ProviderName = "OpenAI" | "OpenRouter.ai" | "Default Provider";
 
 export const OPENAI_API_URL = "https://api.openai.com/v1";
 export const OPENROUTER_API_URL = "https://openrouter.ai/api/v1";
+export const DEFAULTPROVIDER_API_URL = "https://taras-free_open_router.web.val.run/api/v1";
 
 const urlToNameMap: { [key: string]: ProviderName } = {
   [OPENAI_API_URL]: "OpenAI",
   [OPENROUTER_API_URL]: "OpenRouter.ai",
+  [DEFAULTPROVIDER_API_URL]: "Default Provider",
```
tarasglek, line 14: Call it "Free AI Models"

src/components/Message/AppMessage/Instructions.tsx
kliu57: @tarasglek Should I change the name to `FreeModelProvider.ts` and put it in the subfolder `DefaultProvider`

kliu57: @tarasglek Should it be `providers/DefaultProvider/FreeModelProvider.ts` or `DefaultProvider/FreeModelProvider.ts` (not under providers)

src/lib/providers/DefaultProvider.ts
kliu57: I have made this fix

src/lib/ChatCraftProvider.ts
kliu57: I have made this fix

src/components/Message/AppMessage/Instructions.tsx
kliu57: I have made this fix

src/lib/providers/DefaultProvider.ts
kliu57: @tarasglek If I user `queryModels` to get the list of models and then get the first in that list, I would need to make `defaultModelForProvider` and potentially make every function that calls it async as well. Is this what we want?

humphd: I took a look at this with @kliu57 and it's going to complicate too many things to dynamically get this model--we rely on the default model logic being synchronous, and querying the models will kill that.

In a follow-up we could think about how to handle provider models, but for now, this is the same thing we're doing for every other provider, so I think it's good.

src/Chat/ChatHeader.tsx
```diff
@@ -154,7 +154,7 @@ function ChatHeader({ chat }: ChatHeaderProps) {
                       {title}
                     </Link>
                   </Text>
-                  {settings.currentProvider?.apiKey && (
+                  {settings.currentProvider.apiKey && (
```
humphd, line 157: Way better with it being required, nice work.

src/lib/ChatCraftProvider.ts
humphd: Yeah, OK, let's leave this.

src/lib/ChatCraftProvider.ts
```diff
@@ -33,4 +37,52 @@ export abstract class ChatCraftProvider {
...
+    return {};
+  }
+
+  get logoUrl() {
+    // If we don't know, use the OpenAI logo as a fallback
+    return "/openai-logo.png";
+  }
+
+  get defaultModel() {
+    return "";
```
humphd, line 63: This seems like a bad idea.  Are you assuming this will get implemented in derived classes?  If so, I would `throw new Error('not implemented');` so we catch it as a bug if we get it wrong.

humphd (commented):
Code changes look good to me, thank you for working so hard on this.  @WangGithub0 might be able to test functionality for you.

I've made a couple of comments, only one of which needs attention here or in a follow-up.

I don't need to read this again. From my point of view, it can land when you're ready.

WangGithub0:
do we need add icon for the free provider?
image

src/lib/ChatCraftProvider.ts
kliu57: I will remove these getters called defaultModel() sicne they aren't being used anywhere except in defaultModelForProvider()

kliu57:
@WangGithub0 I have fixed the white logo issue by setting the logo bg to green for the free models. The logo is there you just couldn't see it since the bg was white and the logo was white. Code I used to do this is inside `ModelAvatar.tsx`

WangGithub0 (approved):
Looks good to me

tarasglek:
omg, we have a chatcraft that one can use without any keys now!

Now if someone wants to figure how to fix my valtown proxy to not buffer SSE before i get to it

PR #508, opened 2024-03-16
enable streaming for free models and pick sane default model

tarasglek:
* switched from valtown to deno, which got me streaming and better perf.
* switched to a better performing model by default


tarasglek:
Hardcoding model names for free provider isn't gonna work as these models cycle in and out. We really need to make default model selection async

PR #512, opened 2024-03-20
Create an index of shared chatcraft chats in R2

### LINKED ISSUE #425: Create an index of shared chatcraft chats in R2
## Approach
We could use a database for this, but I slightly prefer updating a static object in R2 for now.

## Format
We can store the index as atom. See this example https://chatcraft.org/api/share/tarasglek/c2qRonMPsmpCGOukypF1a
This lets us integrate into various feed-oriented flows(like following these in feedly or exporting to mastodon, etc)

## Appearance
https://darekkay.com/blog/rss-styling/ lets the feed look good in a browser.

## Safety
To create the index transactionally/safely, we can use Read-Modify-ConditionalWrite loop:
https://chatcraft.org/api/share/tarasglek/4clRgxqmQfG_ZCOURyu-M I asked mistral medium about this as I wanted newer info, but the code example low quality :)

WangGithub0:
###  After user sharing a chat, the generateAndSaveUserFeed() will automatically get the latest 20 share chats from cloudflare and generated feed file with rss-style in path `${user}/feed.atom`

image

### The generated feed.atom file can be automatically rendered by my chrome RSS Feed Reader:

image

closes #425

### DIFF
--- a/functions/api/share/[[user_id]].ts
+++ b/functions/api/share/[[user_id]].ts
@@ -14,7 +14,7 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
   const { CHATCRAFT_ORG_BUCKET } = env;
   const prefix: string = `${user}/`;
   const { objects } = await CHATCRAFT_ORG_BUCKET.list({ prefix });
-  const xsltUrl = "../../rss-style.xsl";
+  // const xsltUrl = "../../rss-style.xsl";
 
   const feed = new Feed({
     title: `User Feed for ${user}`,
@@ -23,12 +23,12 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
     link: `https://chatcraft.org/api/share/${user}/feed.atom`,
     updated: new Date(),
     feedLinks: {
-      atom: `https://chatcraft.org/api/share/feed/${user}/feed.atom`,
+      atom: `https://chatcraft.org/api/share/${user}/feed.atom`,
     },
     author: {
       name: user,
     },
-    copyright: `Copyright © ${new Date().getFullYear()} by ${user}`,
+    copyright: `Copyright©${new Date().getFullYear()} by ${user}`,
   });
 
   const sortedObjects = objects.sort((a, b) => b.uploaded.getTime() - a.uploaded.getTime());
@@ -39,8 +39,8 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
       const chatContent: string = await chatData.text();
       const $ = load(chatContent);
 
-      const title = $('meta[property="og:title"]').text() || "No Title";
-      const summary = $('meta[name="description"]').attr("content") || "No Summary";
+      const title = $('meta[property="og:title"]').attr("content") || "No Title";
+      const summary = $('meta[property="og:description"]').attr("content") || "No Summary";
       const url = $('meta[property="og:url"]').attr("content") || "No URL";
       const id = url.split("/").pop() || "No ID";
       const preContent = $("pre").text();
@@ -58,6 +58,7 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
     }
   }
 
+  const xsltUrl = "https://daverupert.com/pretty-feed-v3.xsl";
   let feedXml = feed.atom1();
   feedXml =
     `<?xml version="1.0" encoding="UTF-8"?>\n<?xml-stylesheet type="text/xsl" href="${xsltUrl}"?>\n` +
--- a/functions/api/share/[[user_id]].ts
+++ b/functions/api/share/[[user_id]].ts
@@ -17,10 +17,9 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
 
   const feed = new Feed({
     title: `User Feed for ${user}`,
-    description: `This is ${user}'s share chats`,
+    description: `This is ${user}'s shared chats`,
     id: `https://chatcraft.org/api/feed/${user}/feed.atom`,
     link: `https://chatcraft.org/api/feed/${user}/feed.atom`,
-    language: "en",
     updated: new Date(),
     feedLinks: {
       atom: `https://chatcraft.org/api/feed/${user}/feed.atom`,
--- a/functions/api/share/[[user_id]].ts
+++ b/functions/api/share/[[user_id]].ts
@@ -1,4 +1,5 @@
 import { Feed } from "feed";
+import { load } from "cheerio";
 import { errorResponse, createResourcesForEnv } from "../../utils";
 
 interface Env {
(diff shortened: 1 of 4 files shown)


functions/api/share/[[user_id]].ts
```diff
@@ -81,6 +139,9 @@ export const onRequestPut: PagesFunction<Env> = async ({ request, env, params })
       },
     });
 
+    const [user] = user_id;
+    await generateAndSaveUserFeed(env, user);
```
Comment (tarasglek) [inline functions/api/share/[[user_id]].ts:143]: Just pass in request.body instead of getting it from bucket again

tarasglek (changes requested):

functions/api/share/[[user_id]].ts
```diff
@@ -8,6 +8,64 @@ interface Env {
...
+      const urlMatch = metaRefresh?.match(/url=(.+)/i);
+      const url = urlMatch ? urlMatch[1] : "No URL";
+      const idMatch = url.match(/\/c\/[^/]+\/([^/]+)/);
+      const id = idMatch ? idMatch[1] : "No ID";
+      const preContent = doc.querySelector("pre")?.textContent || "";
+      const dateMatch = preContent.match(/date:\s*(.+)/i);
+      const date = dateMatch ? new Date(dateMatch[1]).toISOString() : new Date().toISOString();
+
+      feed += `
+        <entry>
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:44]: Should also include author info

functions/api/share/[[user_id]].ts
Comment (humphd) [reply functions/api/share/[[user_id]].ts:143]: She has to access the bucket itself, so I think this is necessary.

functions/api/share/[[user_id]].ts
```diff
@@ -8,6 +8,64 @@ interface Env {
...
+
+  // Save the feed to R2
+  const feedKey = `${user}/feed/feed.atom`;
+  await CHATCRAFT_ORG_BUCKET.put(feedKey, new TextEncoder().encode(feed), {
+    httpMetadata: {
+      contentType: "application/atom+xml",
+    },
+  });
+
+  return `Feed generated and saved for user ${user}`;
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:66]: Let's not do this.  If you want to do this, make it a log message and return nothing.

humphd (changes requested):

functions/api/share/[[user_id]].ts
```diff
@@ -8,6 +9,75 @@ interface Env {
...
 
+async function generateUserFeed(env: Env, user: string): Promise<void> {
+  const { CHATCRAFT_ORG_BUCKET } = env;
+  const prefix: string = `${user}/`;
+  const { objects } = await CHATCRAFT_ORG_BUCKET.list({ prefix });
+  const xsltUrl = "../../rss-style.xsl";
+
+  const feed = new Feed({
+    title: `User Feed for ${user}`,
+    description: `This is ${user}'s share chats`,
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:20]: shared vs. share

functions/api/share/[[user_id]].ts
```diff
@@ -8,6 +9,75 @@ interface Env {
...
+    id: `https://chatcraft.org/api/feed/${user}/feed.atom`,
+    link: `https://chatcraft.org/api/feed/${user}/feed.atom`,
+    language: "en",
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:23]: This may or may not be true

functions/rss-style.xsl
```diff
@@ -0,0 +1,58 @@
...
+          </div>
+          <div class="py-7">
+            <h1>Atom Feed Preview</h1>
+            <h2><xsl:value-of select="/atom:feed/atom:link[@rel='alternate']/@href"/></h2>
+            <p>
+              <xsl:value-of select="/atom:feed/atom:subtitle"/>
+            </p>
+            <a href="{/atom:feed/atom:link[@rel='alternate']/@href}">Visit Website &#x2192;</a>
+
+            <h2>Recent Posts</h2>
```
humphd, line 31: Recent Shared Chats

functions/rss-style.xsl
```diff
@@ -0,0 +1,58 @@
...
+            <xsl:for-each select="/atom:feed/atom:entry">
+              <div class="pb-7">
+                <div class="text-2 text-offset">
+                  Published on
```
humphd, line 35: Shared on

humphd (changes requested):

humphd (changes requested):
When I try to share a chat and test this, I get this error in the preview version (browser console):

```
Unable to share chat: Unable to share chat: DOMParser is not defined
```

WangGithub0:
> When I try to share a chat and test this, I get this error in the preview version (browser console):
> 
> ```
> Unable to share chat: Unable to share chat: DOMParser is not defined
> ```

Cloudflare Workers don't support DOMParser, I'll revise it using HTMLRewriter

WangGithub0:
@humphd I met this error, could you help me check the log?

Error 1102 Ray ID: 867f62ea7e9e5269 • 2024-03-21 16:41:09 UTC
Worker exceeded resource limits
What happened?
You've requested a page on a website (56878ea8.console-overthinker-dev.pages.dev) that is on the [Cloudflare](https://www.cloudflare.com/5xx-error-landing/) network. An unknown error occurred while rendering the page.

What can I do?
If you are the owner of this website:
you should [login to Cloudflare](https://www.cloudflare.com/login?utm_source=error_100x) and check the error logs for 56878ea8.console-overthinker-dev.pages.dev.

Cloudflare Ray ID: 867f62ea7e9e5269 • Your IP: Click to reveal 89.208.249.207 • Performance & security by [Cloudflare](https://www.cloudflare.com/5xx-error-landing)

humphd:
I can only stream the logs in realtime, but not see old ones.

`Worker exceeded resource limits` implies we've done something to go over the limit (size, calls?)

WangGithub0:
> I can only stream the logs in realtime, but not see old ones.
> 
> `Worker exceeded resource limits` implies we've done something to go over the limit (size, calls?)

I test:
1. if only deal with 10 objects, it works;
2. 50 objects, it doesn't work.

I'm thinking to update an existing feed with new item

WangGithub0:
@tarasglek  @humphd  I can successfully share a chat and run this, but can't get result from the link I used (/api/feed/${user}/feed.atom) and the log, could you help me check it or could you help me grant Cloudflare permission， thanks.

humphd:
Let's debug this together next week some time.  Ping me on Discord and I'll help with logs.

humphd:
> > I can only stream the logs in realtime, but not see old ones.
> > `Worker exceeded resource limits` implies we've done something to go over the limit (size, calls?)
> 
> I test:
> 
> 1. if only deal with 10 objects, it works;
> 2. 50 objects, it doesn't work.
> 
> I'm thinking to get the existing feed and update with the new item.

A feed is often the last N elements in a list vs. everything, so limiting it is fine.  Likely you need to read the objects in pages vs. trying to get them all at once.  So there are probably a few things we can try.

WangGithub0:
> > > I can only stream the logs in realtime, but not see old ones.
> > > `Worker exceeded resource limits` implies we've done something to go over the limit (size, calls?)
> > 
> > 
> > I test:
> > 
> > 1. if only deal with 10 objects, it works;
> > 2. 50 objects, it doesn't work.
> > 
> > I'm thinking to get the existing feed and update with the new item.
> 
> A feed is often the last N elements in a list vs. everything, so limiting it is fine. Likely you need to read the objects in pages vs. trying to get them all at once. So there are probably a few things we can try.

@humphd  how about only showing the latest 20 elements?

humphd:
> @humphd how about only showing the latest 20 elements?

Let's start with that, yes.

functions/api/share/[[user_id]].ts
```diff
@@ -8,6 +10,81 @@ interface Env {
   JWT_SECRET: string;
 }
 
+async function generateUserFeed(env: Env, user: string): Promise<void> {
+  const { CHATCRAFT_ORG_BUCKET } = env;
+  const prefix: string = `${user}/`;
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:15]: tip: you can always skip defining the type when you are assigning a value, since TypeScript can infer what it is from the value.  It feels weird to do this, but it's the norm to leave as many types out as you can, and only help the compiler when it needs it.

package.json
```diff
@@ -22,12 +22,14 @@
     "@szhsin/react-menu": "^4.1.0",
     "@uiw/react-codemirror": "^4.21.21",
     "browser-image-compression": "^2.0.2",
+    "cheerio": "1.0.0-rc.12",
```
humphd, line 25: Published 2 years ago.  Sure feels like this is dead, doesn't it?  Let's do it for now, but I wonder if there is a more well supported option.

humphd (approved):
Code looks good!

package.json
WangGithub0: yeah, it was published 2 years ago and hasn't seen updates since that. I'll do it in follow-up https://github.com/tarasglek/chatcraft.org/issues/532

PR #517, opened 2024-03-21
Set init provider for new users to free model, added info alert banner

### LINKED ISSUE #501: Update on-boarding flow with new Free AI Models provider
With the work in #498 that @kliu57 has done, we need to rethink our on-boarding flow.  Previously, you needed an API Key in order to do anything.  Now, we have a "free tier" that users can use if they don't have a key.

We should think about how to leverage this.  A few options:

1. Default to the free tier and provide some kind of UI decoration that lets the user know that they can add their own API key for more features
2. Update our "instructions" so that the user can easily click a big button to start using the free models if they don't have/don't want to enter a key

kliu57:
- Set initial provider for new users to `FreeModelProvider` (can test by clearing cache)
- New users will not see the Instructions page upon their first visit to chatcraft, they can start talking to FreeModelProvider right away
- Users currently using `FreeModelProvider` will see the following alert banner with link which opens user settings

alert_banner

Closes #501

### Code changes:

Modifications to `settings.ts`:

- Made `FreeModelProvider` the **initial provider** for new users instead of `OpenAiProvider`
- Made the default model of `FreeModelProvider` the **initial model** instead of `OpenAiProvider`'s default model

Modifications to `ChatBase.tsx`:

- Modified `ChatBase.tsx` to show the alert banner to users using `FreeModelProvider` as their currentProvider
- Clicking the link in alert banner opens the User Settings

Modifications to `Instructions.tsx`:

- Made api key field not a required field for `FreeModelProvider`
- Modified form submission handler function so that it will skip key validation for `FreeModelProvider`
- Kept this change even though we are not currently using Instructions page, just in case we need that page in the future

### How to test:

- Either clear your cache (you will automatically be set to free provider) or manually change your provider to the free provider in settings
- Go to the chatcraft main url, you should not see the instructions page and should be able to talk to chatcraft right away
- You should also see the alert banner, which notifies you that you are using free provider
- Click the link in the banner
- The User Settings modal should now open

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -1,5 +1,17 @@
-import { Box, Button, Flex, Grid, GridItem, Text, useDisclosure } from "@chakra-ui/react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import {
+  Alert,
+  AlertDescription,
+  AlertIcon,
+  Box,
+  Button,
+  CloseButton,
+  Flex,
+  Grid,
+  GridItem,
+  Text,
+  useDisclosure,
+} from "@chakra-ui/react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { CgArrowDownO } from "react-icons/cg";
 import { ScrollRestoration } from "react-router-dom";
 
@@ -25,6 +37,8 @@ import {
 } from "../lib/ChatCraftMessage";
 import { ChatCraftCommandRegistry } from "../lib/commands";
 import ChatHeader from "./ChatHeader";
+import { FreeModelProvider } from "../lib/providers/DefaultProvider/FreeModelProvider";
+import PreferencesModal from "../components/PreferencesModal";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -46,6 +60,12 @@ function ChatBase({ chat }: ChatBaseProps) {
   const { error } = useAlert();
   const { user } = useUser();
   const { clearAudioQueue } = useAudioPlayer();
+  const [showAlert, setShowAlert] = useState(true);
+  const {
+    isOpen: isPrefModalOpen,
+    onOpen: onPrefModalOpen,
+    onClose: onPrefModalClose,
+  } = useDisclosure();
 
   // If we can't load models, it's a bad sign for API connectivity.
   // Show an error so the user is aware.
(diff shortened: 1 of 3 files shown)


humphd (changes requested):
Love this!

The new alert banner across the top is 48px high, and our header is 42px.  I wonder if we should make it a bit smaller?  Same with the font size, it seems big.

Make "Click here" the same colour (white) and give it an underline so that it's clear it's a link.

Clicking the "Click here" link while you're already in a chat puts the instructions at the bottom of the chat, and you may not see it.  I wonder how we should deal with that?  We could have it be a link that redirects you to the app with `?show-instructions` on the URL in query string, which could be the trigger for showing this?

When I go to the Instructions, the drop-down for picking a provider shows the Free Model as the selected item, and it wants me to enter an API Key.  This makes it seem like I need to put my API key in for this model, which won't work.  Maybe OpenAI should be the default one we select in the drop-down?

kliu57:
> We could have it be a link that redirects you to the app with `?show-instructions` on the URL in query string, which could be the trigger for showing this?

What do you mean by this? How to achieve this? Also, maybe this can be a follow up issue because there are many places where we have to implement this "scroll fix". The chatcraft behaviour has always had this problem when if you already had some chat history and then you remove your key the instructions will show up underneath your chat history.

---

I thought about changing the provider to OpenAI but I didn't know what to do if the user already had an OpenAi key saved. If we load their key then they won't see Instructions. If we do not load their key then it's weird for them to have the empty field to enter their key when they already have a key saved.

kliu57:
@humphd I have addressed the following:

> The new alert banner across the top is 48px high, and our header is 42px. I wonder if we should make it a bit smaller? Same with the font size, it seems big.

> Make "Click here" the same colour (white) and give it an underline so that it's clear it's a link.

kliu57:
> Clicking the "Click here" link while you're already in a chat puts the instructions at the bottom of the chat

@humphd I have found the fix for this

In `MessagesView.tsx` we have the prevMessages stacked above the instructions as you can see here:

image

In {instructions} was have a check that makes the instructions only display when user has no key.

I just added something similar to the {prevMessages} so they only display when user DOES have a key.

image

tarasglek:
i think the free model message should open options. Showing the old enter a key page is very confusing. I couldnt even figure out how to get out of it

kliu57:
> i think the free model message should open options. Showing the old enter a key page is very confusing. I couldnt even figure out how to get out of it

@tarasglek For sure, I can just open the ChatCraft user settings instead of showing the instructions. I think Dave was thinking that the instructions page would provide more details about how to use ChatCraft and what is ChatCraft. Do you and Dave want to discuss more about this before I make the change?

humphd:
Let's simplify in the name of getting this landed: for now, open the user settings.  Doing proper "instructions" can be a follow-up.

The way we'd do it is to define a new route in `router.ts` that is named something like `/welcome` or something, and it would render the instructions or some other component that explains what ChatCraft is.

Later on, having a proper on-boarding experience seems important.  However, this issue doesn't have to be about that.

Amnish04 (approved):
@kliu57 The changes are pretty clean. 
I have tested the functionality as well. My only concern was that the banner link should rather lead to the **Instructions** page. But going through conversations, I noticed that we are planning to do it in a follow-up, so should be fine.

humphd (approved):
This is great, ship it.

Issue #507, opened 2024-03-16
/import feature successor


tarasglek:
I was thinking oh why the current import feature feels wrong.

1) I can't change it easily. #473 as proposed is an overly fancy fix

2) I don't like /commands cos they are hard to type on mobile and they cause issues like #500 

3) Security aspect restricting /import to logged in users makes for terrible github login UX on mobile browsers/PWA

4) fact that the command disappears after we run it

5) I also dont like that we press `ask gpt` to do something that gpt has nothing to do with.

So I have have the simplest possible solution. Allow one to define POST commands via a yaml textbox in chatcraft options. Eg we would implement current solution by having a default value:
```yaml
handler_url: /proxy
method: GET
match: \/import https?:\/\/\S+
```
this would match current behavior of sending commands that look like 
* `/import https://www.example.com/file.pdf`
* `/import https://www.google.com/file.png`

So upon pressing enter:
1. chatcraft will check if the message matches any of the patterns specified
2. send request with entire message to url specified. `/import?message=/import%20https://www.google.com/file.png`

Response from that url endpoint should then be SSE streamed(can be a followup) back into next message with a clear indicator that it came from chatcraft itself. The endpoint could choose to provide an error message, such as asking for auth.

Since I want a behavior where I can paste urls in without any typing and I wanna use scrape2md, I would extend this with
```yaml
handler_url: /proxy
method: GET
match: \/import https?:\/\/\S+
---
handler_url: https://taras-scrape2md.web.val.run/
match: https?:\/\/\S+
```

We could later consider auth variables like $GITHUB_AUTH_TOKEN to this.
```yaml
handler_url: /proxy
method: POST
match: \/import https?:\/\/\S+
options:
  env:
    - name: GITHUB_AUTH_TOKEN
    - value: $GITHUB_AUTH_TOKEN
```

This seems easy to implement and could be very extensible. yaml should get validated and not let user save changes unless it matches some schema. This could also be done in a followup.

tarasglek:
Cool thing is that we could also allow these patterns to match ai responses...eg provide a run-custom-handler button to validate chatcraft output.

```yaml
handle_url: https://some/url/with-dockerfile/validator
match: "^```docker"
match:
  multiline: true
  model_output: true
```

We could move all run-on-server stuff to this

tarasglek:
Obviously one could build a complex ui over this, but yaml gets us going sooner

tarasglek:
[...]

PR #519, opened 2024-03-21
Implement initial version of web handlers

Amnish04:
This is my initial attempt in implementing the idea of what we can call web handlers as discussed in #507 .

I have:
1. Written a `WebHandler` class registering/searching web handlers and executing their logic i.e. sending request to `handler url` based on request config (so far request method)
2. Added logic to check if the prompt matches with **match pattern** of one of the registered Web Handlers. The search is sequential as mentioned in the issue, and the first matched web handler processes the prompt url. If there is no match, the prompt is released to the regular control flow (slash command or sent to AI).
3. Currently, I have hard coded 2 Web Handlers as mentioned in issue, but we can later create UI to configure as many as user wants in `localStorage`.

**Demo:**

Prompt: `https://github.com/tarasglek/chatcraft.org/pull/519`

image

What I don't like is how we are currently allowing prompt like `import <url>` to be registered as match patterns. This causes ambiguity and I had to write specialized logic to handle edge cases like
```ts
url = extractFirstUrl(url) ?? ""; // When the input is not a url itself
```
as an example.

I feel like we should restrict web handler match patterns to just urls, and let import command serve its own purpose.

 Maybe I am wrong and don't understand fully where we are going with this.

@humphd @tarasglek Please let me know what you think.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -211,9 +211,9 @@ function ChatBase({ chat }: ChatBaseProps) {
       }
 
       // If we have a web handler registered for this url
-      if (prompt && WebHandler.getMatchingHandler(prompt)) {
-        const handler = WebHandler.getMatchingHandler(prompt);
+      const handler = WebHandler.getMatchingHandler(prompt ?? "");
 
+      if (prompt && handler) {
         try {
           const result = await handler!.executeHandler(prompt); // We know handler is always there
 
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -215,7 +215,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
       if (prompt && handler) {
         try {
-          const result = await handler!.executeHandler(prompt); // We know handler is always there
+          const result = await handler.executeHandler(prompt);
 
           chat.addMessage(new ChatCraftHumanMessage({ user, text: result }));
           forceScroll();
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -35,6 +35,7 @@ import {
   ChatCraftFunctionResultMessage,
   ChatCraftHumanMessage,
 } from "../lib/ChatCraftMessage";
+import { WebHandler } from "../lib/WebHandler";
 import { ChatCraftCommandRegistry } from "../lib/commands";
 import ChatHeader from "./ChatHeader";
 import { FreeModelProvider } from "../lib/providers/DefaultProvider/FreeModelProvider";
@@ -207,6 +208,26 @@ function ChatBase({ chat }: ChatBaseProps) {
       // Special-case for "help", to invoke /help command
       if (prompt?.toLowerCase() === "help") {
         prompt = "/help";
+      }
+
+      // If we have a web handler registered for this url
+      if (prompt && WebHandler.getMatchingHandler(prompt)) {
+        const handler = WebHandler.getMatchingHandler(prompt);
+
+        try {
+          const result = await handler!.executeHandler(prompt); // We know handler is always there
+
+          chat.addMessage(new ChatCraftHumanMessage({ user, text: result }));
+          forceScroll();
+        } catch (err: any) {
+          error({
+            title: "Error running Web Handler",
+            message: err.message,
+          });
+        }
+
+        setLoading(false);
+        return;
       }
 
       // If this is a slash command, execute that instead of prompting LLM
--- a/src/lib/WebHandler.ts
+++ b/src/lib/WebHandler.ts
@@ -17,15 +17,15 @@ export class WebHandler {
     this.matchPattern = matchPattern;
   }
 
-  isMatchingHandler(url: string) {
-    return this.matchPattern.test(url);
+  isMatchingHandler(message: string) {
+    return this.matchPattern.test(message);
   }
 
-  async executeHandler(url: string): Promise<string> {
+  async executeHandler(message: string): Promise<string> {
     const requestUrl = new URL(this.handlerUrl);
 
     const params = new URLSearchParams();
-    params.append("url", url);
+    params.append("url", message);
 
     requestUrl.search = params.toString();
 
@@ -36,14 +36,16 @@ export class WebHandler {
     // const type = guessType(response.headers.get("Content-Type"));
     const content = (await response.text()).trim();
 
-    const messageHeader = `**Web Handler**: [${this.handlerUrl}](${this.handlerUrl})?url=[${url}](${url})`;
-    const text = `${messageHeader}\n\n` + content;
+    const resultHeader = `**Web Handler**: [${this.handlerUrl}](${this.handlerUrl})?url=[${message}](${message})`;
+    const text = `${resultHeader}\n\n` + content;
 
     return text;
   }
 
-  static getMatchingHandler(url: string): WebHandler | null {
-    return this.getRegisteredHandlers().find((handler) => handler.isMatchingHandler(url)) ?? null;
+  static getMatchingHandler(message: string): WebHandler | null {
+    return (
+      this.getRegisteredHandlers().find((handler) => handler.isMatchingHandler(message)) ?? null
+    );
   }
 
   static getRegisteredHandlers(): WebHandler[] {
--- a/src/lib/WebHandler.ts
+++ b/src/lib/WebHandler.ts
@@ -10,11 +10,11 @@ export class WebHandler {
   }: {
     handlerUrl: string;
     method: HttpMethod;
-    matchPattern: RegExp;
+    matchPattern: RegExp | string;
   }) {
     this.handlerUrl = handlerUrl;
     this.method = method;
-    this.matchPattern = matchPattern;
+    this.matchPattern = matchPattern instanceof RegExp ? matchPattern : new RegExp(matchPattern);
   }
 
   isMatchingHandler(message: string) {
@@ -48,17 +48,15 @@ export class WebHandler {
   }
 
   static getRegisteredHandlers(): WebHandler[] {
-    // TODO: Fetch from localStorage
-
-    const supportedHandlers = [
-      new WebHandler({
-        handlerUrl: "https://taras-scrape2md.web.val.run/",
-        method: HttpMethod.GET,
-        matchPattern: /^https:\/\/\S+$/,
-      }),
-    ];
+    const supportedHandlers = JSON.parse(
+      localStorage.getItem(webHandlersLocalStorageKey) ?? ""
+    ).map((handler: WebHandler) => new WebHandler(handler));
 
     return supportedHandlers;
+  }
+
+  isValidHandler(): boolean {
+    return !!this.handlerUrl && !!this.method && !!this.matchPattern;
   }
 }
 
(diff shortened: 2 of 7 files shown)


tarasglek:
Prompt: https://github.com/tarasglek/chatcraft.org/pull/519 <-- seems like wrong url?

src/Chat/ChatBase.tsx
```diff
@@ -156,6 +157,26 @@ function ChatBase({ chat }: ChatBaseProps) {
         prompt = "/help";
       }
 
+      // If we have a web handler registered for this url
+      if (prompt && WebHandler.getMatchingHandler(prompt)) {
```
tarasglek, line 161: does `prompt` here mean current chat `message`?

src/Chat/ChatBase.tsx
```diff
@@ -156,6 +157,26 @@ function ChatBase({ chat }: ChatBaseProps) {
...
+        const handler = WebHandler.getMatchingHandler(prompt);
```
tarasglek, line 162: should do const handler above if and then reuse it

src/lib/WebHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+    const supportedHandlers = [
+      new WebHandler({
+        handlerUrl: "/api/proxy",
+        method: HttpMethod.GET,
+        matchPattern: /^\/import https?:\/\/\S+/,
+      }),
+      new WebHandler({
+        handlerUrl: "https://taras-scrape2md.web.val.run/",
+        method: HttpMethod.GET,
+        matchPattern: /^https:\/\/\S+/,
```
tarasglek, line 74: you should fix match pattern to have $ at end to match entire message only

tarasglek: in both

humphd:
Not critical, but I'd vote to get rid of `/proxy ...` style regexes as you wrote above, and only do this for URLs.  Leave `/slash` commands for built-in things, and this is something different.

`WebHandler` is a nice name for this.  I'm excited to see this happen so quickly.

src/Chat/ChatBase.tsx
Amnish04: Yes, I think `promptText` is a better name. But it's already in use later in the same function

src/lib/WebHandler.ts
Amnish04: Thanks for noticing, I've made the change.

Amnish04:
@humphd Agreed, I've removed the `/proxy` handler. Maybe we can have further validation for regex patterns in a future PR where we implement UI for handler registration

src/Chat/ChatBase.tsx
```diff
@@ -156,6 +157,26 @@ function ChatBase({ chat }: ChatBaseProps) {
         prompt = "/help";
       }
 
+      // If we have a web handler registered for this url
+      const handler = WebHandler.getMatchingHandler(prompt ?? "");
+
+      if (prompt && handler) {
+        try {
+          const result = await handler!.executeHandler(prompt); // We know handler is always there
```
tarasglek, line 165: would just do const executedHandlerResult =  WebHandler.executeHandler(prompt ?? "");

tarasglek: s/prompt/message/

src/lib/WebHandler.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const type = guessType(response.headers.get("Content-Type"));
+    const content = (await response.text()).trim();
+
+    const command = `**Web Handler**: [${this.handlerUrl}](${this.handlerUrl})?url=[${url}](${url})`;
+    const text =
+      `${command}\n\n` +
+      // If it's already markdown, dump it into the message as is;
+      // otherwise, wrap it in a code block with appropriate type
+      (type === "markdown" ? content : `\`\`\`${type}\n${content}` + `\n\`\`\``);
```
tarasglek, line 53: i think text/plain is also ok to show directly

src/Chat/ChatBase.tsx
tarasglek: eg should just execute first matching handler

tarasglek: i think you also wanna set some sort of network spinner given that you doing network io

src/lib/WebHandler.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.method = method;
+    this.matchPattern = matchPattern;
+  }
+
+  isMatchingHandler(url: string) {
+    return this.matchPattern.test(url);
+  }
+
+  async executeHandler(url: string): Promise<string> {
+    url = extractFirstUrl(url) ?? ""; // When the input is not a url itself
```
tarasglek, line 27: i don't understand what this code is doing. needs a comment re algo here

src/lib/WebHandler.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  static getRegisteredHandlers(): WebHandler[] {
+    // TODO: Fetch from localStorage
+
+    const supportedHandlers = [
+      new WebHandler({
+        handlerUrl: "https://taras-scrape2md.web.val.run/",
+        method: HttpMethod.GET,
+        matchPattern: /^https:\/\/\S+$/,
```
tarasglek, line 69: i think you also need to make this pattern do /^(https:\/\/\S+)$/

tarasglek: so the algo is you apply regexp..then if it has a match group, eg ()..pass that to url. if it doesn't just pass whole message(tho that should only work for post i think)

tarasglek:
I would love it if you could expose this as a  webhandler section in prefs(maybe hidden by default under advanced), so i could enter yaml directly

tarasglek:
Until it's editable there isn't too much value in specifying handlers this way

src/Chat/ChatBase.tsx
tarasglek: i see when i run it, there is a spinner..don't understand why :)

tarasglek:
if i put 
```
https://github.com/tarasglek/chatcraft.org/pull/519/files some trash
```
into a message. then it's handled as a call to llm.

if i do edit, cut it down to 
```
https://github.com/tarasglek/chatcraft.org/pull/519/files
```
and press re-ask it's also passed to llm. Should be webhandled

src/Chat/ChatBase.tsx
Amnish04: > would just do const executedHandlerResult = WebHandler.executeHandler(prompt ?? "");

That will work, but I would still have to have a function to know if the prompt text matches a web handler.

src/lib/WebHandler.ts
```diff
@@ -0,0 +1,98 @@
...
+    handlerUrl: string;
+    method: HttpMethod;
+    matchPattern: RegExp;
+  }) {
+    this.handlerUrl = handlerUrl;
+    this.method = method;
+    this.matchPattern = matchPattern;
+  }
+
+  isMatchingHandler(url: string) {
```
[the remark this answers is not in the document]
Amnish04: Sorry I didn't get that :)

src/lib/WebHandler.ts
Amnish04: Yes, I noticed it doesn't make any difference if I skip this step. Will make the change

src/lib/WebHandler.ts
Amnish04: This was for handling non-url prompts. But now that we have decided to not allow those, we can remove this

src/lib/WebHandler.ts
Amnish04: > i think you also need to make this pattern do /^(https://\S+)$/

Do you mean wrapping in parenthesis?

Amnish04:
> I would love it if you could expose this as a webhandler section in prefs(maybe hidden by default under advanced), so i could

I was actually thinking of making it similar to how Functions are done.

We could have an accordion item labelled "Web Handlers"
image

And each web handler item would lead to a separate page with a web handler config form anf all those configs could be stored in user's localStorage.

Amnish04:
> if i put
> 
> ```
> https://github.com/tarasglek/chatcraft.org/pull/519/files some trash
> ```
> 
> into a message. then it's handled as a call to llm.
> 
> if i do edit, cut it down to
> 
> ```
> https://github.com/tarasglek/chatcraft.org/pull/519/files
> ```
> 
> and press re-ask it's also passed to llm. Should be webhandled

@tarasglek Nice catch, we should fix it. Can we do it in a follow up?

src/lib/WebHandler.ts
Amnish04: @tarasglek Just pushed the changes

src/lib/WebHandler.ts
Amnish04: > so the algo is you apply regexp..then if it has a match group, eg ()..pass that to url. if it doesn't just pass whole message(tho that should only work for post i think)

Yes, I go through each **registered** Web Handler, and try to match the prompt against match pattern of the handler. If one matches, I invoke that handler's configured `handlerUrl` which takes the prompt text as `url` query param for now.

Maybe we can make how the prompt is sent with request configurable too in the future

src/lib/WebHandler.ts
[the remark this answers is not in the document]
Amnish04: Ohhh, you mean I should rename `url` to `message`

Didn't realize you were talking in `sed` command 😅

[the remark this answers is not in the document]
Amnish04: Just pushed https://github.com/tarasglek/chatcraft.org/pull/519/commits/9cf7f2da72fd0831f7f79dca715722951fc2aaed

tarasglek:
> > I would love it if you could expose this as a webhandler section in prefs(maybe hidden by default under advanced), so i could
> 
> I was actually thinking of making it similar to how Functions are done.
> 
> We could have an accordion item labelled "Web Handlers" image
> 
> And each web handler item would lead to a separate page with a web handler config form anf all those configs could be stored in user's localStorage.

I am not super excited about ui for this. I would like the web handlers to be exposed as a yaml configuration so I could copy/paste it between devices, share it with friends. The UI is less practical than text here. If we had a ui as a higher level layer to generate that config, that's fine, but then you have to worry about round-trip editing, etc. Seems not worth it.

tarasglek:
We can delay re-ask to followup, but webhandlers(agree this is a cool name) should land with config editing.

Amnish04:
@tarasglek @humphd Just pushed some new changes.

The user can now enter **YAML** configurations for Web Handlers (as Taras suggested) with a modal that a pops up by clicking the new **Web Handlers** menu item in our profile image drop menu.
image

**Configuration Modal:**
image

This config is persisted in **localStorage**
image

with the help of a `use-web-handlers` hook I wrote for its management.
image

Now, we can configure as many patterns we want to and the first match is executed.

Based on my config, if I enter a **Youtube Url** the request is redirected to the api of one of my projects (I have intentionally set CORS headers to *).

image

**Result:**
image

humphd:
Can you change the font in that textarea to be monospaced?

Another idea (follow-up) would be to use the code I already have for the CodeMirror editor that we use in functions, so we get a better dev experience.

Amnish04:
@humphd Agreed, I should have used that in the first place. Just filed a [follow up](https://github.com/tarasglek/chatcraft.org/issues/540).

Also, I've changed the `fontFamily` of TextArea to `monospace`.

Please let me know if you have any more suggestions

Amnish04:
> We can delay re-ask to followup, but webhandlers(agree this is a cool name) should land with config editing.

@tarasglek I've been thinking about the Re-ask issue you brought up. The problem is when the Web Handler response has been generated, the message looks like this
image

If we are to resend the request to `handlerUrl` in this case, should we extract all the required data by parsing the message header (first line) or maybe build a custom message component for Web Handler responses that has all that data in its state.

Amnish04:
@tarasglek Thanks for reviewing! I am trying to avoid having it in the `User Settings` as the modal is going to be too long. And since Web Handlers could potentially grow in complexity, I would prefer it be separate thing :)

Also, we were disusing the weird Markdown today in the meeting. Not sure, but we're probably adding redundant backtics somewhere. I'll file a follow up for this

PR #520, opened 2024-03-21
Hide text to speech options when not supported

### LINKED ISSUE #518: Text to Speech operations fail when TTS not supported
In #497, I forgot to make a check whether TTS is supported before redering **Speak** and **Download as Audio** options.
Screenshot 2024-03-21 at 12 35 53 PM

When the user clicks on those options with an unsupported provider, the operation fails
image

Amnish04:
In #497, I added **Speak** and **Download as Audio** options to MessageBase menu.

However, those options are even visible when TTS is not supported.

In this PR, I have added checks to hide those options when that feature is not supported.

**Switching to free model:**
image

This fixes #518

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -115,6 +115,9 @@ function MessageBase({
   const [, copyToClipboard] = useCopyToClipboard();
   const { id, date, text, imageUrls } = message;
   const { models } = useModels();
+  const isTtsSupported = useMemo(() => {
+    return !!models.filter((model) => model.id.includes("tts"))?.length;
+  }, [models]);
   const { onCopy } = useClipboard(text);
   const { info, error } = useAlert();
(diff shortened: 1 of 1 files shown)


kliu57 (approved):
Looks good. I tested it on chrome and firefox and I am only seeing the speak and download audio options when I am using OpenAI provider, which is what we want. 

For future it would be nice if there was a way to stop the talking in the middle in case it's speaking a very long paragraph and the user wants it stopped. But I guess we would need a design idea first.

Amnish04:
@kliu57 Thanks for reviewing. 

> For future it would be nice if there was a way to stop the talking in the middle in case it's speaking a very long paragraph and the user wants it stopped. But I guess we would need a design idea first.

Filed #521 for this

mingming-ma (approved):
LGTM! Tested and works well as expected.

Amnish04:
@mingming-ma Thanks for reviewing!

PR #526, opened 2024-03-26
Add Nomnoml Renderer

### LINKED ISSUE #466: Add nomnoml renderer
I found https://github.com/skanaar/nomnoml works with llms and is easier to write than mermaid for SQL type stuff and renders better

rjwignar:
This fixes #466 

Summary of Changes
---
This adds Nomnoml rendering support to ChatCraft using [skanaar/nomnoml](https://github.com/skanaar/nomnoml).

This feature was heavily based on [Mermaid rendering support](https://github.com/tarasglek/chatcraft.org/pull/48/files) using [mermaid-js](https://github.com/mermaid-js/mermaid). For example, Nomnoml rendering is handled by [NomnomlPreview.tsx](https://github.com/tarasglek/chatcraft.org/blob/bb7c55737566be0231fdb525ec725bb8b8409680/src/components/NomnomlPreview.tsx) similar to how Mermaid rendering is handled by MermaidPreview.tsx, and similar to mermaid, the rendering is done in Markdown.tsx when a nomnoml code block is detected:
https://github.com/tarasglek/chatcraft.org/blob/bb7c55737566be0231fdb525ec725bb8b8409680/src/components/Markdown.tsx#L108-L115

Differences from Mermaid rendering
---
Although this feature is modelled off of Mermaid rendering, due to the differences between the nomnoml and mermaid-js libraries I had to implement Nomnoml rendering differently:
### No Error Message in the Nomnoml Preview
Whenever mermaidjs fails to render mermaid code, it renders an image in the preview, but
Nomnoml currently does not render an error image, so failed renders are rendered as a blank diagram instead:
image

### No SVG ID value support
Unlike the [mermaid-js render method](https://github.com/mermaid-js/mermaid/blob/8e95c4db55c10b152e262d6fb271415f86f616d6/packages/mermaid/src/mermaid.ts#L359-L406), nomnoml's [renderSvg method](https://github.com/skanaar/nomnoml/blob/89a1ac4470f27f5f042c156ec53e7a328f27e507/src/nomnoml.ts#L60-L71) does not support an ID parameter, so in NomnomlPreview.tsx an ID value is not generated or assigned to the rendered SVG.

### No config method/no way to directly modify default styling
Nomnoml currently (as of **1.6.2**) **does not** expose a config method like [mermaid.initialize](https://mermaid.js.org/config/usage.html#using-the-mermaidapi-initialize-mermaid-initialize-call).
This means there's currently no way to programatically modify the default styling of nomnoml diagrams.
However, nomnoml has [directives](https://github.com/skanaar/nomnoml?tab=readme-ov-file#directives) that modify styling

#### Alternative Styling Modification
In the future, I see two ways users can reliably indirectly modify the styling:
- The user can update the system prompt with instructions for nomnoml styling
- Within a ChatCraft conversation, the user iteratively ask the LLM to modify the styling

However, both of these methods are heavily dependent on the LLM.
Even GPT-4 seems to [hallucinate on nomnoml syntax](https://issue-466.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/AkWRUWWG5zYmNsg-Z2e9R) when asked to modify styling, unless a [directive example is provided](https://issue-466.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/UW08evlCIutmdi_v1poCZ)

### DIFF
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -8,7 +8,7 @@ I follow these rules when responding:
 - Use GitHub flavored Markdown
 - ALWAYS include the programming language name (js) or type of data (csv) at the start of Markdown code blocks
 - Format ALL lines of code to 80 characters or fewer
-- Use Mermaid diagrams when discussing visual topics
+- Use nomnoml or Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
 - If responding with math markdown, inline or otherwise, I use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs \`($$)\`, for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
 `;
--- a/src/lib/system-prompt.ts
+++ b/src/lib/system-prompt.ts
@@ -8,7 +8,7 @@ I follow these rules when responding:
 - Use GitHub flavored Markdown
 - ALWAYS include the programming language name (js) or type of data (csv) at the start of Markdown code blocks
 - Format ALL lines of code to 80 characters or fewer
-- Use nomnoml or Mermaid diagrams when discussing visual topics
+- Use Nomnoml or Mermaid diagrams when discussing visual topics
 - If using functions, only use the specific functions I have been provided with
 - If responding with math markdown, inline or otherwise, I use KaTeX syntax in math Markdown by enclosing EVERY mathematical expression, equation, variable, and formula with double-dollar signs \`($$)\`, for example: $$O(n\\log n)$$, $$1024 * 1024 = 1048576$$, $$1024^2$$, $$X$$
 `;
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "lodash-es": "^4.17.21",
     "mermaid": "^10.8.0",
     "nanoid": "^5.0.5",
+    "nomnoml": "^1.6.2",
     "openai": "^4.26.1",
     "react": "^18.2.0",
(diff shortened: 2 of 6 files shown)


humphd (approved):
Code looks good to me.

We should consider an update to the default system prompt to use this.

rjwignar:
Hi @tarasglek, @humphd, I've updated the default system prompt following your suggestions.
I've also added new changes to the PR and updated my PR summary. I'm interested to see if you have any thoughts or suggestions on my observations when making this PR.

humphd (approved):
Love it!

Screenshot 2024-03-26 at 7 11 19 PM

tarasglek:
> Love it!
> 
> Screenshot 2024-03-26 at 7 11 19 PM

This is fantastic diagram to use in a blog :)

rjwignar:
Recently updated this PR with an error message that renders in the Nomnoml preview if the nomnoml code cannot be rendered:
image

Copying the preview card still copies the actual error:
nomnomlCopyError

src/lib/system-prompt.ts
```diff
@@ -8,7 +8,7 @@ I follow these rules when responding:
 - Use GitHub flavored Markdown
 - ALWAYS include the programming language name (js) or type of data (csv) at the start of Markdown code blocks
 - Format ALL lines of code to 80 characters or fewer
-- Use Mermaid diagrams when discussing visual topics
+- Use nomnoml or Mermaid diagrams when discussing visual topics
```
WangGithub0, line 11: should you use a capital N for "Nomnoml" ?

WangGithub0 (approved):
It's so cool!  Looks good to me!

PR #527, opened 2024-03-27
/ - Keyboard shortcut to focus on current chat prompt input

### LINKED ISSUE #511: Keyboard shortcut to focus the input bar
Is there no keyboard shortcut to focus the input bar? I find myself wishing for this all the time, and I'm surprised that a tool meant for programmers doesn't have it.

Other sites that DO have such keyboard shortcut:

- Perplexity: `Cmd-/`
- Phind: `/`
- Google search: `/`
- Gmail: `/`
- YouTube: `/`
- Jira: `/`
- Github: `/`

Others don't have it and it drives me mad: Google Gemini, (wow, many sites have it!)

Anyways, this is obviously becoming a pattern on the web and I think ChatCraft should implement it as well. I'm not a web developer so I'm not sure how to do this, otherwise I'd make a PR for it.

rjwignar:
This fixes #511 

Summary of Changes
---
This adds a `/` shortcut that sets focus on the current Chat prompt entry.
This adds two methods to the ChatBase component:
`handleChatInputFocus()` - sets focus on Chat prompt entry
`handleForwardSlash` - useEffect that listens for '/' key entry when user is not focused on a text input box

I initially tried to do this by extending [useKeyDownHandler](https://github.com/tarasglek/chatcraft.org/blob/9208edb962839a3cce0524efc4e36beb6fedf652/src/hooks/use-key-down-handler.ts) with a method to handle forward slashes and adding a **useKeyDownHandler** instance to the ChatBase component, but the listener would only trigger when inside a text area (like the search bar or chat prompt entry).

If there's a better way to implement this please let me know.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -11,7 +11,7 @@ import {
   Text,
   useDisclosure,
 } from "@chakra-ui/react";
-import { FormEvent, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { CgArrowDownO } from "react-icons/cg";
 import { ScrollRestoration } from "react-router-dom";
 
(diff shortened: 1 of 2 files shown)


humphd:
@chadparker do you want to test this fix out for us?  You can use the preview URL at https://8a118347.console-overthinker-dev.pages.dev/

chadparker:
I tried it out and the functionality seems great, thank you!

(I don't know React, but the code seems sensible. I'll let somebody closer to the project officially review it, though.)

humphd (approved):
Looks good, thanks for fixing this so fast.

PR #529, opened 2024-03-27
Vite Config: Explicitly provide null navigateFallback url

### LINKED ISSUE #528: Workbox: non-precached-url [{"url":"index.html"}]
I've noticed this for a while but never thought to create an Issue on it until now.
This doesn't seem to break ChatCraft, whenever I load ChatCraft I see this error in the Dev Tools:
image

This seems to be a workbox related issue, based on this [discussion](https://github.com/vite-pwa/vite-plugin-pwa/issues/120) the issue might be related to the workbox settings in vite.config.ts:
https://github.com/tarasglek/chatcraft.org/blob/62da804d21c6b92950e7febe87e0d1c7fa10cd45/vite.config.ts#L65-L78

rjwignar:
This fixes #528 

I copied the code for this fix from this [comment](https://github.com/vite-pwa/vite-plugin-pwa/issues/120#issuecomment-1202579983) which gets rid of the non-precached-url error.

I don't think this will have any unforeseen consequences, as the `navigateFallback` parameter [was ostensibly already null](https://github.com/vite-pwa/vite-plugin-pwa/blob/4abcd5462f7ce030c7418303ba5bc6d5dd5b2634/docs/guide/development.md?plain=1#L69-L75):

image

But I'd like additional opinions.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -75,6 +75,9 @@ export default defineConfig({
           "**/assets/*.{js,json,css,ico,png,svg}",
           ...buildLanguageGlobPatterns("**/assets/"),
         ],
+        // Don't fallback on document based (e.g. `/some-page`) requests
+        // Even though this says `null` by default, I had to set this specifically to `null` to make it work
+        navigateFallback: null,
       },
     }),
   ],


humphd (approved):
Thank you!  I hated that error.

Issue #504, opened 2024-03-15
Add loading/warning info when image is being generated


mingming-ma:
Follow up PR for https://github.com/tarasglek/chatcraft.org/pull/493#issuecomment-1998092723

Same reason for the https://github.com/tarasglek/chatcraft.org/pull/531#issuecomment-2026212932

PR #531, opened 2024-03-27
Support size settings in the image command

### LINKED ISSUE #499: Support size settings in the image command
#493 needs followup PR to support configuration the `size` or user friendly words such as "horizontal", "vertical" and "normal" (default), in order to solve this, we should probably look for a " delimited prompt rather than concatenating all args.

mingming-ma:
Using regex to find matched size options which staring with `@`
  
Examples:

`/image @landscape dog`
`/image dog @portrait`
`/image dog @1792x1024`
`/image @1024x1792 dog`

Fixes #499

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -419,7 +419,7 @@ export async function isGenerateImageSupported() {
   );
 }
 
-type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
+export type dalle3ImageSize = "1024x1024" | "1792x1024" | "1024x1792";
 
 export const generateImage = async (
   prompt: string,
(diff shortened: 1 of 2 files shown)


humphd:
Can you do `/image [landscape|portrait|widthxheight] prompt` and not require the `@` syntax?  That is, have this arg go second, and be optional.

humphd:
Agree, I think using `size=[portrait|landscape]` or `layout=[portrait|landscape]` as optional first arg is smart.

mingming-ma:
@humphd  I think `layout` make more sense, another thought, how about also support the `p` and `l`, like `layout=[p|portrait|l|landscape]`

WangGithub0:
when I test:     `/image layout=1 cat`   it still return me (DALL·E 3 result [layout square] of the prompt: cat), should we give some warning/info message?

humphd:
> @humphd I think `layout` make more sense, another thought, how about also support the `p` and `l`, like `layout=[p|portrait|l|landscape]`

I like it

mingming-ma:
@WangGithub0 I realized the info part can be done in #504, as the functionality is highly overlapping, will make a PR later today. I'll add this in that issue.

WangGithub0 (approved):
It looks good to me!

PR #535, opened 2024-03-27
Add the Providers selection Menu Group to easy switch between providers

### LINKED ISSUE #510: Make it easier to switch between providers
Currently, we assume that the user only ever wants to use a single provider. For example, the "Ask" menu shows models for the current provider:

Screenshot 2024-03-17 at 5 22 20 PM

However, @kliu57 has been doing awesome work to add support for multiple providers.  Now that we support 3 providers, and soon arbitrary providers (see #440), it would be interesting if the "Ask" menu allowed you to easily switch to another of your configured providers and use those models.

I wonder if this would be a good UX:

```
+--------------+
| model-1      |
| model-2      |
| ...          |
|--------------|
| ✔️ Provider 1 |
|   Provider 2 |
|   Provider 3 |
+--------------+
Ask
```

In the wireframe above, the models for the current provider (i.e., "Provider 1") are shown.  In addition, the other providers are also available, and clicking on "Provider 2" or "Provider 3" would cause the list of models to change.

There might be a better UX for this, but that's one option.

Rachit1313:
Closes #510 

Tasks Completed:
- [x] Enable switching between providers from the ask menu for Desktop Prompt
- [x] Enable switching between providers from the ask menu for Mobile Prompt
- [x] Use MenuGroup from Chakra UI to group models and providers distinctly

Files Changed:
PromptForm.tsx

Screenshots:
image

### DIFF
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -38,7 +38,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
 
   return (
     <ButtonGroup variant="outline" isAttached>
-      <Menu placement="top" strategy="fixed" offset={[-90, 0]}>
+      <Menu placement="top" strategy="fixed" closeOnSelect={false} offset={[-90, 0]}>
         <IconButton
           type="submit"
           size="lg"
@@ -102,7 +102,11 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
             {models
               .filter((model) => !usingOfficialOpenAI() || model.id.includes("gpt"))
               .map((model) => (
-                <MenuItem key={model.id} onClick={() => setSettings({ ...settings, model })}>
+                <MenuItem
+                  closeOnSelect={true}
+                  key={model.id}
+                  onClick={() => setSettings({ ...settings, model })}
+                >
                   {model.prettyModel}
                 </MenuItem>
               ))}
@@ -185,7 +189,7 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
           </Button>
         </Tooltip>
       )}
-      <Menu placement="top" strategy="fixed">
+      <Menu placement="top" strategy="fixed" closeOnSelect={false}>
         <MenuButton
           as={IconButton}
           size="sm"
@@ -198,7 +202,11 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
             {models
               .filter((model) => !usingOfficialOpenAI() || model.id.includes("gpt"))
               .map((model) => (
-                <MenuItem key={model.id} onClick={() => setSettings({ ...settings, model })}>
+                <MenuItem
+                  closeOnSelect={true}
+                  key={model.id}
+                  onClick={() => setSettings({ ...settings, model })}
+                >
                   {model.prettyModel}
                 </MenuItem>
               ))}
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -11,7 +11,6 @@ import {
   MenuGroup,
 } from "@chakra-ui/react";
 import { TbChevronUp, TbSend } from "react-icons/tb";
-import { FREEMODELPROVIDER_API_URL } from "../../lib/ChatCraftProvider";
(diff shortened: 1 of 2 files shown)


humphd:
We've got the same alignment issue here that you fixed elsewhere.  All of the providers should be indented the same amount (left margin) and the icon only showing beside the selected provider.  It shouldn't indent the text when the checkbox is beside it (e.g., the text should already be at that position).

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -26,6 +30,7 @@ type PromptSendButtonProps = {
 function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
+  const supportedProviders = settings.providers;
```
kliu57, line 33: You can just use settings.providers directory no need to set it to a variable. We don't want to confused it with supportedProviders which is a different array.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -95,13 +100,37 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
...
+          </MenuGroup>
+          <MenuGroup title="Providers">
+            {Object.values({
+              ...supportedProviders,
+              [FREEMODELPROVIDER_API_URL]: new FreeModelProvider(),
+            }).map((provider) => (
+              <MenuItem
+                key={provider.apiUrl}
+                onClick={() => {
+                  const newProvider = providerFromJSON({
```
kliu57, line 120: My hotfix from pr #537 will handle deserializing the provider objects inside settings.providers. So you can directly set the currentProvider to the item in the array. You do not need to call providerFromJSON.

`setSettings({ ...settings, currentProvider: settings.providers[provider.name] });`

I am using provider.name as the key here because **after** the hotfix is merged you will need to use the name as key.

kliu57 (changes requested):
Some minor changes requested. Also you have to make the change from settings.provider using the apiUrl as the key to using name as the key once the hotfix #537 is merged, thanks.

Rachit1313:
@kliu57 Do you want me change key based on name now or after the PR has been merged. I believe that it would be better to do it after merge as I would be able to test what I'm implementing. Or maybe.. I get this merged and get a PR ready to fix stuff which could be merged right after your PR.

@humphd, What would be the best approach for this?

Could you please also review the indentation changes based on your feedback?

kliu57:
> @kliu57 Do you want me change key based on name now or after the PR has been merged. I believe that it would be better to do it after merge as I would be able to test what I'm implementing. Or maybe.. I get this merged and get a PR ready to fix stuff which could be merged right after your PR.
> 
> @humphd, What would be the best approach for this?
> 
> Could you please also review the indentation changes based on your feedback?

You cannot make the change until after my hotfix is merged.

Rachit1313:
> I'm not loving the checkmark icon. I wonder if there is something better, maybe an arrow that points at the one
> 
> Maybe we should use a right-hand pointing chevron? https://react-icons.github.io/react-icons/search/#q=chevron
> 
> Something like:
> 
> Screenshot 2024-03-28 at 6 30 56 PM
> Another thing I notice: when you change providers, it closes the menu, which is surprising, since you are probably going to want to pick a model from the new list. Can we make it so that choosing the provider doesn't close the menu?
> 
> Screenshot 2024-03-28 at 6 26 44 PM

I changed the icon based on your feedback but not sure how can we achieve the second thing? I tried. using disclosure but even that didn't work. Maybe it has to close the menu after switching providers because it has to re render the model based on the change.

humphd:
What if you call `e.preventDefault()` when it's a click on the provider?

Rachit1313:
> e.preventDefault()

Nope doesn't work

Rachit1313:
> > e.preventDefault()
> 
> Nope doesn't work

I found a solution for keeping it open i.e we can add [closeOnSelect={false} ](https://github.com/chakra-ui/chakra-ui/pull/3762) but the problem is this applies to the whole menu and it wont even close when we select model. Is this something acceptable?

kliu57 (approved):
Thanks for doing the [refactoring](https://github.com/tarasglek/chatcraft.org/pull/535/commits/2dd9addbb845d2263cf0b25b7a90d6a658ceba2e) I asked for, everything looks good to me.

PR #537, opened 2024-03-28
Migration fix for settings.providers

kliu57:
Closes #536

#### Part 1 - Changing settings.providers to use name as the key

Files:
[`src/components/Message/AppMessage/Instructions.tsx`](https://github.com/tarasglek/chatcraft.org/pull/537/files#diff-9fcdb80f3814fd658a3bb7dcaf58478d27938a43cde9049d449210652e77434c)
[`src/components/PreferencesModal.tsx`](https://github.com/tarasglek/chatcraft.org/pull/537/files#diff-1025748a16e04005898c20c05840616be48ca0eb4eb4c156248de41a2f3bdd7d)
[`src/lib/settings.ts`](https://github.com/tarasglek/chatcraft.org/pull/537/files#diff-b5f5521c9795e314efdf6293ad3afda789a1402bbf5e2a304e20efa8134d2a15)

#### Part 2 - Deserializing settings.providers in [`settings.ts`](https://github.com/tarasglek/chatcraft.org/pull/537/files#diff-b5f5521c9795e314efdf6293ad3afda789a1402bbf5e2a304e20efa8134d2a15)

To test this you can set your settings json to the following (replace YOUR_OPENAI_KEY with your key). Using my json you start with two openai providers and two openrouter providers and after navigating to or refreshing chatcraft the deserializer will change it to having two.

```
[...]

PR #538, opened 2024-03-28
Hotfix for bug introduced in pr 537

kliu57:
Accidentally introduced a bug in this morning's merged PR #537 

Took a statement out of an if statement in `settings.ts`, which may cause some users to be unable to load ChatCraft.

The users that are affected are users who have not logged into ChatCraft since last month.
They have a `settings` JSON in their localStorage, but do not have `settings.currentProvider`. (Because the last time they logged in was last month, back before I introduced `settings.currentProvider`)

Also if a user went to localStorage and manually set settings to {}, they would also be affected.

**The code from PR 537 which caused the bug:**

image

**Fix**

Restoring the if statement in this hotfix fixes the issue.

**How to test:**

1. Go to localStorage and remove the currentProvider part of the settings json, or just clear the entire settings json and put {}
2. Load ChatCraft
3. Result - (it should have an error in production, but not in this PR's preview)

### DIFF
--- a/src/lib/settings.ts
+++ b/src/lib/settings.ts
@@ -79,17 +79,19 @@ export const deserializer = (value: string): Settings => {
   }
 
   // Handle migration of deprecated apiKey, apiUrl
-  if (settings.apiKey && settings.apiUrl) {
-    const newProvider = providerFromUrl(settings.apiUrl, settings.apiKey);
-    settings.currentProvider = newProvider;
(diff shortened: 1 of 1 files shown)


rjwignar (approved):
I followed your testing steps for both the production website and your branch preview and this change fixes the issue.

mingming-ma (approved):
I tested and I can see in this pr after reload, the json will be populated not null value and load correct, which is failed in current prod. Code changes LGTM

PR #539, opened 2024-03-29
Add loading info when image is being generated

### LINKED ISSUE #504: Add loading/warning info when image is being generated

mingming-ma:
## Description

As the `command` is non-component, I added an alert function in `utils` which uses the [Standalone Toasts](https://chakra-ui.com/docs/components/toast/usage#standalone-toasts) of chakra-ui to create toasts from outside of React Components

## Result

image

Fixes #504

### DIFF
--- a/src/lib/commands/ImageCommand.ts
+++ b/src/lib/commands/ImageCommand.ts
@@ -18,22 +18,23 @@ export class ImageCommand extends ChatCraftCommand {
     if (!(args && args[0])) {
       throw new Error("must include a prompt");
     }
+
     const prompt = args.join(" ");
     let imageUrls: string[] = [];
     const text = `(DALL·E 3 result of the prompt: ${prompt})`;
 
-    const now = new Date();
+    const alertId = loading({
+      title: `Generating image, please wait.`,
+    });
+
     try {
-      loading({
-        id: now.toISOString(),
-        title: `Generating image, please wait.`,
-      });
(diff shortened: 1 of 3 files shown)


src/lib/commands/ImageCommand.ts
```diff
@@ -19,12 +22,18 @@ export class ImageCommand extends ChatCraftCommand {
     let imageUrls: string[] = [];
     const text = `(DALL·E 3 result of the prompt: ${prompt})`;
 
+    const now = new Date();
     try {
+      loading({
+        id: now.toISOString(),
```
Amnish04, line 28: Maybe you can generate an id inside your `utilizeAlert` function and make passing an id optional.

Amnish04 (changes requested):
@mingming-ma I was already looking for a way to show alert without that hook. Thanks to you, I can later use it in my PR :D

Looks good for the most part, just left a suggestion for toast id generation

src/lib/commands/ImageCommand.ts
mingming-ma: Agree, let me fix that

Amnish04 (approved):
Looks good to me!

PR #543, opened 2024-03-30
Replace "Cancel Adding Provider" SVG with an IconButton

### LINKED ISSUE #551: User settings providers table - new row close icon html fix
Proposed by @Amnish04 and discussed during Apr 3 team meeting.

Use `<IconButton>` instead of `<Box>`

Change from using a `<Box>` to `<IconButton>` for the new provider cancel icon, while maintaining the look of the icon. 

image

Preferred alignment (The icon should line up correctly with the checkboxes underneath):

image

Amnish04:
As discussed in https://github.com/tarasglek/chatcraft.org/pull/530#discussion_r1545481317, we should not use Svg's directly as an alternative of a button since it is semantically incorrect.

https://www.reddit.com/r/webdev/comments/r72zq6/comment/hmwvobm/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

I tried preserving the look of the button, and was successful for the most part.

**Normal:**
image

**Active:**
image

This fixes #551


Amnish04:
The diffs have become too messy. Continued in #554

PR #549, opened 2024-04-03
Fix image generated not available after some time

### LINKED ISSUE #496: Fix image generated not available after some time
The image generated will be not available due to URL lifetime ( #493 )

image

Need to figure out how to store so user can back check them later.

[DALL-E API Image URL lifetime](https://community.openai.com/t/dall-e-api-image-url-lifetime/53672)

mingming-ma:
## Description
In the DALL·E 3 generations, if response returned as URL, it will expire after [an hour](https://platform.openai.com/docs/guides/images/example-dall-e-3-generations)

This update change the response format to `b64_json` so we can keep the image's Data

Fixes #496

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -449,9 +449,13 @@ export const generateImage = async ({
       prompt,
       n,
       size,
+      response_format: "b64_json",
     });
 
-    const imageUrls = response.data.map((img: any) => img.url);
+    const imageUrls = response.data.map((item) => {
+      const base64ImageData = item.b64_json;
+      return `data:image/jpeg;base64,${base64ImageData}`;
+    });
     return imageUrls;
   } catch (error: any) {
     throw new Error(error);
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -452,10 +452,7 @@ export const generateImage = async ({
       response_format: "b64_json",
     });
 
-    const imageUrls = response.data.map((item) => {
-      const base64ImageData = item.b64_json;
-      return `data:image/jpeg;base64,${base64ImageData}`;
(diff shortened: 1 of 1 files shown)


mingming-ma:
I want to share a chat to prove the image can be keep, but I found I can't (even no images) in this branch, maybe cloudflare side?

I tested share generated image chat on lated prod is ok.


image

As we can see the `src` now is base64 string, so it should keep.

image

kliu57 (approved):
Looks good to me. I tested it and the image is still there after more than 15 hours.

image

Amnish04 (approved):
@mingming-ma Works as expected!

Amnish04:
One drawback I see to this approach is that we are now storing the image data locally.

If indexed db has a max storage limit, its going to fill up quickly if the user primarily works with images. And if there's a way to bypass that limit, it will fill up user's device storage.

I am not sure if this is actually an issue and if it is, what would be the best approach to solve this. 
Should we be maintaining a db in cloud for non-text media types? That would probably make things unnecessarily complicated.

humphd:
@Amnish04 I understand your concern.  I do wonder, though, that since these images are ones that the user has explicitly opted into creating, if us storing them does make sense?  If I make a cool image, and a week later want to try and find it again, but it's gone, that sucks.

IndexedDB can store a lot of data without issue.

Amnish04:
> @Amnish04 I understand your concern. I do wonder, though, that since these images are ones that the user has explicitly opted into creating, if us storing them does make sense? If I make a cool image, and a week later want to try and find it again, but it's gone, that sucks.
> 
> IndexedDB can store a lot of data without issue.

Right, makes sense.

PR #554, opened 2024-04-03
Replace "Cancel Adding Provider" SVG with an IconButton

### LINKED ISSUE #551: User settings providers table - new row close icon html fix

Amnish04:
As discussed in https://github.com/tarasglek/chatcraft.org/pull/530#discussion_r1545481317, we should not use Svg's directly as an alternative of a button since it is semantically incorrect.

https://www.reddit.com/r/webdev/comments/r72zq6/comment/hmwvobm/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

I tried preserving the look of the button, and was successful for the most part.

**Normal:**
image

**Active:**
image

This fixes #551

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -695,7 +695,7 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
                         return (
                           <Tr key={provider.name}>
                             <Td>
-                              <Flex justifyContent={"center"}>
+                              <Flex width={6} justifyContent={"center"}>
                                 <Checkbox
                                   onChange={() => handleSelectedProviderChange(provider)}
                                   isChecked={selectedProvider?.name === provider.name}
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -572,34 +572,36 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
                     {newCustomProvider && (
                       <Tr>
                         <Td sx={{ textAlign: "center" }}>
-                          <Box
-                            as={MdCancel}
+                          <IconButton
+                            aria-label="Cancel adding new provider"
+                            icon={<MdCancel />}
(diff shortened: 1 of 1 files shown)


Amnish04:
@kliu57 I have fixed the alignment issue you mentioned

kliu57:
@Amnish04 it's causing the column headings as well as the data underneath the new row to shift to the right the moment we add the new row. I think it is due to the padding/margin of the iconbutton being greater than the padding/margin around the checkboxes. Can you fix it so there is no shift?

Amnish04:
@kliu57 Just pushed the fix for that

src/components/PreferencesModal.tsx
```diff
@@ -571,36 +571,38 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
+                            }}
+                            _focus={{
                               _focus: {
                                 outline: "none",
                                 boxShadow: "0 0 0 3px rgba(66, 153, 225, 0.6)",
                                 borderColor: "blue.300",
                               },
                             }}
+                            _active={{
+                              backgroundColor: "none",
```
kliu57, line 604: Can we remove this?

kliu57 (changes requested):

src/components/PreferencesModal.tsx
Amnish04: I did it to make it look similar to yours.

Without that style:
image

kliu57: oh ok I see. that's fine the way it is then

kliu57 (approved):
looks good to me

Amnish04:
Thanks for reviewing

PR #557, opened 2024-04-03
Hide voice selection option in prefs menu when tts is not supported

### LINKED ISSUE #523: Hide Voice Selection setting when TTS is not supported
I partially fixed this issue in #520, but forgot to add a check for voice selection 🤦🏻‍♂️

image

We need to hide this setting too when TTS is not supported

Amnish04:
This is to make sure that the voice selection options in Preferences Modal are hidden when TTS is not supported by the current provider.

**With OpenAI:**

image

**With Free Provider:**

image

Fixes #523

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -10,6 +10,7 @@ import {
   FormLabel,
   IconButton,
   Input,
+  InputGroup,
   Kbd,
   Link,
   Modal,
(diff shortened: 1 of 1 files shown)


kliu57 (approved):
Tested this with OpenAi and a custom new openAI provider and works well for showing the voice section. And when tested with free model or openrouter it works well for hiding the voice section. LGTM

Amnish04:
@kliu57 Thanks for reviewing!

PR #559, opened 2024-04-04
Updated regex to ignore backslash

### LINKED ISSUE #427: /import command is sensitive to trailing backslash
Compare the results of doing both of these:

1. `/import https://github.com/tarasglek/chatcraft.org/pull/421/` with trailing backslash
2. `https://github.com/tarasglek/chatcraft.org/pull/421` without trailing backslash

I think the issue here is that we don't want to mess with a URL like https://github.com/tarasglek/chatcraft.org/pull/421#issue-2124528458 (comment).

The code in question is in https://github.com/tarasglek/chatcraft.org/blob/main/functions/lib/rewriters/github-rewriter.ts

Rachit1313:
Closes #427 

### Changes Made:
Updated regex to ignore backslash in github commands.

### How to test:
Both the links with backslash at the end or without would recieve diff when a link of github PR is imported using /import command.

### DIFF
--- a/functions/lib/rewriters/github-rewriter.ts
+++ b/functions/lib/rewriters/github-rewriter.ts
@@ -9,7 +9,7 @@ export class GitHubRewriter extends DefaultRewriter {
   async rewriteUrl(url: URL) {
     // Blob URLs - https://github.com/<owner>/<repo>/blob/<branch>/<path>
     if (url.origin === "https://github.com" && /\/(.*)\/blob\/(.*)\/?$/.test(url.pathname)) {
-      // Adjusted to ignore trailing slash
+      // https://github.com/<owner>/<repo>/blob/<branch>/<path> ->
       return new URL(
         url.href.replace(
           /^https:\/\/github\.com\/(.*)\/blob\/(.*)\/?$/,
@@ -20,7 +20,7 @@ export class GitHubRewriter extends DefaultRewriter {
 
     // Gist URLs - https://gist.github.com/<owner>/<gist-id>
     if (url.origin === "https://gist.github.com") {
-      // Adjusted to ignore trailing slash
+      // https://gist.github.com/<owner>/<gist-id> -> https://gist.githubusercontent.com/<owner>/<gist-id>/raw
       return new URL(
         url.href.replace(
           /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)\/?$/,
@@ -31,10 +31,13 @@ export class GitHubRewriter extends DefaultRewriter {
 
     // PR URLs - https://github.com/<owner>/<repo>/pull/<number>
     if (url.origin === "https://github.com" && /\/(.*)\/pull\/(.*)\/?$/.test(url.pathname)) {
+      // TODO: we need to add more logic to deal with a specific comment in a PR via the GitHub API
+      // Example: https://github.com/tarasglek/chatcraft.org/pull/370#issuecomment-1916037856
       if (url.hash.startsWith("#issuecomment-")) {
         return url;
       }
 
+      // https://github.com/<owner>/<repo>/pull/<number> -> https://patch-diff.githubusercontent.com/raw/<owner>/<repo>/pull/<number>.patch
       return new URL(
         url.href.replace(
           /^https:\/\/github\.com\/([^/]+)\/([^/]+)\/pull\/(\d+)\/?$/,
--- a/functions/lib/rewriters/github-rewriter.ts
+++ b/functions/lib/rewriters/github-rewriter.ts
@@ -8,11 +8,11 @@ export class GitHubRewriter extends DefaultRewriter {
 
   async rewriteUrl(url: URL) {
     // Blob URLs - https://github.com/<owner>/<repo>/blob/<branch>/<path>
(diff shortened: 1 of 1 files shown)


kliu57 (changes requested):
```
Behvaiour is different for /import https://github.com/tarasglek/chatcraft.org/pull/421

Than /import http://github.com/tarasglek/chatcraft.org/pull/421
```

```
Moreover, behaviour is different from production for /import https://github.com/tarasglek/chatcraft.org/pull/421/ and all others

Output stops at line 83
```
image

kliu57:
Cancelling this PR since the change was made in commit https://github.com/tarasglek/chatcraft.org/commit/bcdd39ca56ae7a6b73f1516833f3f1488ed6dbf3

functions/lib/rewriters/github-rewriter.ts
```diff
@@ -8,11 +8,11 @@ export class GitHubRewriter extends DefaultRewriter {
 
   async rewriteUrl(url: URL) {
     // Blob URLs - https://github.com/<owner>/<repo>/blob/<branch>/<path>
-    if (url.origin === "https://github.com" && /\/(.*)\/blob\/(.*)$/.test(url.pathname)) {
-      // https://github.com/<owner>/<repo>/blob/<branch>/<path> -> https://raw.githubusercontent.com<owner>/<repo>/<branch>/<path>
+    if (url.origin === "https://github.com" && /\/(.*)\/blob\/(.*)\/?$/.test(url.pathname)) {
```
humphd, line 11:
```
# Works
/import https://github.com/tarasglek/chatcraft.org/blob/main/.eslintignore

# Fails
/import https://github.com/tarasglek/chatcraft.org/blob/main/.eslintignore/
```

The latter returns a 400 from the server

functions/lib/rewriters/github-rewriter.ts
```diff
@@ -23,14 +23,14 @@ export class GitHubRewriter extends DefaultRewriter {
...
-          /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)$/,
+          /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)\/?$/,
           "https://gist.githubusercontent.com/$1/$2/raw"
         )
       );
     }
 
     // PR URLs - https://github.com/<owner>/<repo>/pull/<number>
-    if (url.origin === "https://github.com" && /\/(.*)\/pull\/(.*)$/.test(url.pathname)) {
+    if (url.origin === "https://github.com" && /\/(.*)\/pull\/(.*)\/?$/.test(url.pathname)) {
```
humphd, line 33: This one seems to work well for me.

functions/lib/rewriters/github-rewriter.ts
```diff
@@ -23,14 +23,14 @@ export class GitHubRewriter extends DefaultRewriter {
       // https://gist.github.com/<owner>/<gist-id> -> https://gist.githubusercontent.com/<owner>/<gist-id>/raw
       return new URL(
         url.href.replace(
-          /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)$/,
+          /^https:\/\/gist\.github\.com\/([a-zA-Z0-9_-]+)\/([a-f0-9]+)\/?$/,
```
humphd, line 26: This is working well for me.

humphd (changes requested):

functions/lib/rewriters/github-rewriter.ts
Rachit1313: I just verified that this is something we have in production as well

humphd: Do you want to fix it while you're working on this or split it off into a future issue?

Rachit1313: I would like to work on debugging this, but as this PR fixes the problem for PRs and gist, i would like to get this merged and create another issue and PR for that one.

Rachit1313:
And for the second part i.e difference between https and http, I believe this is because we are only looking for https in the origin of the URL to confirm if we want to rewrite this with github-rewriters.ts. Everywhere in our code we used https to verify.
image
I am not sure, just guessing, @humphd would be able to answer this.

humphd:
> @kliu57 , for me both the import command on PR 412 without backslash returns the same number of lines i.e 83 in production as well in my branch preview URL. image image

this is likely due to it being a diff, with only ~6 lines of context (so it naturally cuts if off), which is normal.

humphd:
> @kliu57 And for the second part i.e difference between https and http, I believe this is because we are only looking for https in the origin of the URL to confirm if we want to rewrite this with github-rewriters.ts. Everywhere in our code we used https to verify. image I am not sure, just guessing, @humphd would be able to answer this.

I think requiring HTTPS is fine.

mingming-ma (commented):
Here is comparing results:

| | Prod | This PR |
| ------------- | ------------- | ------|
| import https, has end slash| image  | image  |
| import https, no end slash  | image| image |
| import http, has end slash| image  |image  |
| import http, no  end slash  | image| image |

Rachit1313:
I think we're good to land this. I'll fix for 400 error of blob as a separate issue.

rjwignar (approved):
These changes LGTM.
I think the Blob URL issue can be resolved in a followup.

PR #560, opened 2024-04-04
Configure CodeMirror editor for Web Handlers configuration

### LINKED ISSUE #540: Use CodeMirror editor for Web Handlers configuration
As a follow-up for #519, we should be using a dedicated code editor for Web Handlers configuration that is written in YAML.

We are already using [CodeMirror](https://github.com/uiwjs/react-codemirror) for ChatCraft functions, which should be good for this purpose as well.
image

Amnish04:
In this PR, I have:

1. Replaced the [TextArea](https://chakra-ui.com/docs/components/textarea/usage) that was being used to configure **Web Handlers** with [CodeMirror](https://github.com/uiwjs/react-codemirror) code editor to ensure a good developer experience.

image

2. Also added some onboarding instructions so users can easily understand how this feature works.
This instructions can be hidden/shown with a switch I've added right above the editor.

image

3. We, of course, also get all the functionalities `CodeHeader` component implements - including **copying**, **downloading** and opening the code in a new browser tab.

This closes #540

### DIFF
--- a/src/components/WebHandlersConfigModal.tsx
+++ b/src/components/WebHandlersConfigModal.tsx
@@ -1,9 +1,6 @@
 import {
   Box,
   Button,
-  Flex,
-  FormControl,
-  FormLabel,
   Heading,
   Modal,
   ModalBody,
@@ -12,7 +9,6 @@ import {
   ModalFooter,
   ModalHeader,
   ModalOverlay,
-  Switch,
   Text,
   VStack,
   useColorMode,
@@ -29,7 +25,6 @@ import CodeHeader from "./CodeHeader";
 
 import { yaml } from "@codemirror/lang-yaml";
 import { MdSignalCellularAlt } from "react-icons/md";
-import { useSettings } from "../hooks/use-settings";
 import useMobileBreakpoint from "../hooks/use-mobile-breakpoint";
 
 type WebHandlersConfigModalProps = {
@@ -41,11 +36,9 @@ type WebHandlersConfigModalProps = {
 function WebHandlersConfigModal({ isOpen, onClose, finalFocusRef }: WebHandlersConfigModalProps) {
   const { webHandlers, registerHandlers } = useWebHandlers();
   const { success, error } = useAlert();
-  const { settings, setSettings } = useSettings();
 
-  const getWebHandlersYaml = useCallback(
-    (webHandlers: WebHandlers) => {
-      const onBoardingInstructions = `##############################################################################
+  const getWebHandlersYaml = useCallback((webHandlers: WebHandlers) => {
+    const onBoardingInstructions = `##############################################################################
 ## You can configure "match patterns" for certain types
 ## of URLs, that send an HTTP request to your
 ## configured "handler url".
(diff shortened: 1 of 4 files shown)


Amnish04:
I just added a new pref to our Settings called `showWebHandlersInstructions`.

I am storing it in local storage so familiar users don't have to turn it off on each refresh.

tarasglek:
I don't think you need to hide instructions and ui for that. Just have instructions in default value

src/components/WebHandlersConfigModal.tsx
```diff
@@ -79,30 +116,84 @@ function WebHandlersConfigModal({ isOpen, onClose, finalFocusRef }: WebHandlersC
...
+  }, [error, registerHandlers, success, webHandlerConfig]);
+
+  const configDownloadFilename = "WebHandlersConfig.yaml";
+  const { colorMode } = useColorMode();
+  const editorHeight = "350px";
+  const isMobile = useMobileBreakpoint();
 
   return (
-    <Modal isOpen={isOpen} onClose={onModalClose} size="lg" finalFocusRef={finalFocusRef}>
+    <Modal isOpen={isOpen} onClose={onModalClose} size="2xl" finalFocusRef={finalFocusRef}>
```
humphd, line 127: Doest he modal have a prop that allows resizing?

src/components/WebHandlersConfigModal.tsx
```diff
@@ -79,30 +116,84 @@ function WebHandlersConfigModal({ isOpen, onClose, finalFocusRef }: WebHandlersC
...
-                placeholder="Please enter the configurations of your handlers in YAML"
-              />
-
-              <FormHelperText>
-                The first handler with a successful match pattern is executed
-              </FormHelperText>
-            </FormControl>
+            <Text textAlign={isMobile ? "justify" : "start"}>
+              {/* eslint-disable-next-line react/no-unescaped-entities */}
+              If you want to extend ChatCraft's functionality by plugging external services your own
```
humphd, line 138: "If you want to extend ChatCraft's default URL handling by adding your own external services..."

humphd (changes requested):

Amnish04:
> I don't think you need to hide instructions and ui for that. Just have instructions in default value

@tarasglek I added that since I found it a little unnecessary to have to scroll below all that info even when you don't need it.

But maybe its an overkill. Just removed it.

src/components/WebHandlersConfigModal.tsx
Amnish04: Just fixed the message

src/components/WebHandlersConfigModal.tsx
Amnish04: The modal is resizable by default. I think this property is to specify the max width. Since the addition of comments was demanding more space to be user friendly, I increased it.

humphd:
Can we reduce the width of the instructions comment, so it doesn't cause horizontal scrolling?


https://github.com/tarasglek/chatcraft.org/assets/427398/14a86748-81c9-4ab5-a058-ea6a1d44b2d5

Amnish04:
@humphd Weird, there is not horizontal scrolling on my side 🤔
image

Amnish04:
@humphd What's the width of your modal in pixels?

Amnish04:
> 672px

Huh, its same for me. I wonder what could be causing it?

humphd:
If you remove some of the extra `###` characters, which aren't needed, it will be fine.

Amnish04:
> If you remove some of the extra `###` characters, which aren't needed, it will be fine.

Just pushed. Should work well now
image

Amnish04:
@humphd I was also thinking we should probably store the default web handler in memory instead of local storage.

```yaml
- handlerUrl: https://taras-scrape2md.web.val.run/
  method: GET
  matchPattern: ^https:\/\/\S+$
```

And add it at the end of the list so it takes the least precedence.

Right now, if the user deletes it, it's gone forever until they clear their local storage and the hook add it back for them

tarasglek:
can also put the docs under config :)

humphd (approved):

Amnish04:
> can also put the docs under config :)

@tarasglek I just added the sample handler at the end as a comment. I am not sure if that's how you wanted but we can always have a follow up issue to make any changes.

humphd:
@Amnish04 I would go with this and improve in follow-ups. What you have is awesome imho.

PR #562, opened 2024-04-04
Logo bg color fix for all models that don't have logo

kliu57:
**Bug:**

Currently some logos in OpenRouter and custom providers appear as white, since they have a default white logo on a white bg.

**Fix:**

Set the default Avatar to have green bg color.

---
code change:

image


src/components/ModelAvatar.tsx
```diff
@@ -12,14 +12,12 @@ export default function ModelAvatar({ model, size }: { model: ChatCraftModel; si
   if (id.includes("gpt-3.5-turbo")) {
```
tarasglek, line 12: change this to else if

tarasglek:
lgtm

src/components/ModelAvatar.tsx
kliu57: done

kliu57:
> lgtm

Made the change for the else if, will merge this now.

PR #565, opened 2024-04-04
Add logo for non-OpenAI models

### LINKED ISSUE #564: Don't use OpenAI logo for non-OpenAI models
When we first started ChatCraft, we only used OpenAI and our fallback logo was the OpenAI logo.  Now, we can use arbitrary models from any provider.  When we aren't using OpenAI, we need some generic logo for an AI model, and should apply a colour based on hash of the provider/model name or something.

mingming-ma:
## Description

Using the  [Avatar Fallbacks](https://chakra-ui.com/docs/components/avatar#avatar-fallbacks)  behaviour of `chakra-ui`to set the name with simple hash function to set the background colour for different models

## Result

image

Fixes #564

### DIFF
--- a/src/components/ModelAvatar.tsx
+++ b/src/components/ModelAvatar.tsx
@@ -3,7 +3,7 @@ import { Avatar } from "@chakra-ui/react";
 import { ChatCraftModel } from "../lib/ChatCraftModel";
 
 export default function ModelAvatar({ model, size }: { model: ChatCraftModel; size: "sm" | "xs" }) {
-  const { id, logoUrl, prettyModel } = model;
+  const { id, logoUrl, logoBg, prettyModel } = model;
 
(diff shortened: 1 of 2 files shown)


kliu57:
> @humphd @kliu57
> 
> Since model 'auto' has no slash, currently code logic set the mode `name` fallback to `openai` because OpenAI uses `gpt-3.5-turbo` with no vendor info.
> 
> ```ts
>     this.vendor = parts.length > 1 ? parts[0] : "openai";
> 
>     if (vendor === "openai") {
>       return "/openai-logo.png";
>     }
> ```
> 
> This lead to still use open ai logo in the free model
> 
> image
> Is it ok to change `auto` to something like `provider/auto`?

responded to you in discord

humphd (approved):

PR #567, opened 2024-04-05
Update ChatCraftCommand::isCommand() criteria for slash function

### LINKED ISSUE #500: Chatcraft commands conflict with code errors
parsting below block
```
/workspaces/DeepStructure/misc/getsetup/add-to-notion.ts:181
        throw new Error(`Notion API call failed: ${await response.text()}`);
              ^


Error: Notion API call failed: {"object":"error","status":400,"code":"validation_error","message":"body failed validation: body.children[0].heading_1.rich_text should be defined, instead was `undefined`.","request_id":"f73a1254-1204-4610-9820-76f6890821dc"}
    at notionApiCall (/workspaces/DeepStructure/misc/getsetup/add-to-notion.ts:181:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at createHeadingWithText (/workspaces/DeepStructure/misc/getsetup/add-to-notion.ts:137:5)
    at updateOrAppendAiSummary (/workspaces/DeepStructure/misc/getsetup/add-to-notion.ts:113:5)
    at main (/workspaces/DeepStructure/misc/getsetup/add-to-notion.ts:192:5)

Node.js v20.8.1
```
causes chatcraft to just spin the loading icon. We should probably not assume all /things are commands

rjwignar:
This fixes #500.
The issue encountered above is due to ChatBase interpreting a non-slash command (in this case, a filepath in an error message) as a slash command:
image

`isCommand()` currently treats every prompt starting with a forward slash as a potential slash command, then either tries to execute the slash command or parse the invalid command to notify the user (i.e **[invalid command] is not a valid command!**).

This PR reuses the **parseCommand() Regex** in **isCommand()** so things like filepaths aren't treated as a potential slash command:
https://github.com/tarasglek/chatcraft.org/blob/f7b28038b3820d617a14b346536377977f4379d0/src/lib/ChatCraftCommand.ts#L10-L12

How to Test
---
- Test every existing slash command in the preview and confirm they still work
- Copy/paste the codeblock from the Issue description into a new chat and confirm it is sent as a message analyzed by the LLM.

Potential Issue
---
I think this is a step in the right direction, but the Regex in the new **isCommand()** is not perfect.
If the user enters a slash command containing a non-word (special) character, it will be sent to the LLM:
image

### DIFF
--- a/src/lib/ChatCraftCommand.ts
+++ b/src/lib/ChatCraftCommand.ts
@@ -18,6 +18,8 @@ export abstract class ChatCraftCommand {
 
   // Checks if a string is a command.
   static isCommand(input: string): boolean {
-    return input.startsWith("/");
+    // Check if there's something resembling a command (no non-word characters in the portion after forward slash)
+    const match = input.match(/^\/(\w+)(?:\s+(.*))?$/);
+    return !!match;
   }
 }
--- a/src/lib/ChatCraftCommand.ts
+++ b/src/lib/ChatCraftCommand.ts
@@ -19,7 +19,6 @@ export abstract class ChatCraftCommand {
   // Checks if a string is a command.
   static isCommand(input: string): boolean {
     // Check if there's something resembling a command (no non-word characters in the portion after forward slash)
-    const match = input.match(/^\/(\w+)(?:\s+(.*))?$/);
-    return !!match;
+    return /^\/(\w+)(?:\s+(.*))?$/.test(input);
   }
 }


src/lib/ChatCraftCommand.ts
```diff
@@ -18,6 +18,8 @@ export abstract class ChatCraftCommand {
 
   // Checks if a string is a command.
   static isCommand(input: string): boolean {
-    return input.startsWith("/");
+    // Check if there's something resembling a command (no non-word characters in the portion after forward slash)
+    const match = input.match(/^\/(\w+)(?:\s+(.*))?$/);
```
humphd, line 22: If you don't need the match groups, use test:

```ts
return /^\/(\w+)(?:\s+(.*))?$/.test(input)
```

This gives you a boolean.

humphd (approved):

src/lib/ChatCraftCommand.ts
rjwignar: Thank you! I was looking for a better way to test if the input string has a regex match.
I've updated my PR to use [test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test)

humphd (approved):
ship it

PR #574, opened 2024-04-08
Using linkedom to parse HTML on cloudflare

### LINKED ISSUE #532: Try to find other package to parse HTML on cloudflare
In this [PR](https://github.com/tarasglek/chatcraft.org/pull/512), using cheerio to parse HTML on cloudflare, but it was published 2 years ago and hasn't seen updates since that. 
Try to find a better one.

WangGithub0:
closes #532

### DIFF
--- a/functions/api/share/[[user_id]].ts
+++ b/functions/api/share/[[user_id]].ts
@@ -1,5 +1,5 @@
 import { Feed } from "feed";
-import { load } from "cheerio";
+import { DOMParser } from "linkedom";
 import { errorResponse, createResourcesForEnv } from "../../utils";
 
 interface Env {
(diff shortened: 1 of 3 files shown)


functions/api/share/[[user_id]].ts
```diff
@@ -37,13 +37,21 @@ async function generateUserFeed(env: Env, user: string): Promise<void> {
     const chatData = await CHATCRAFT_ORG_BUCKET.get(object.key);
     if (chatData) {
       const chatContent = await chatData.text();
-      const $ = load(chatContent);
+      const document = new DOMParser().parseFromString(chatContent, "text/html");
+
+      const getTitle = document.querySelector('meta[property="og:title"]');
+      const title = getTitle ? getTitle.getAttribute("content") : "No Title";
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:43]: Another pattern you can use here is:

```ts
const title = getTitle.getAttribute("content") ?? "No Title";
```

Comment (WangGithub0) [reply functions/api/share/[[user_id]].ts:43]: there are some document don't have title sometimes, I tried this way, it threw "Unable to share chat: Cannot read properties of null (reading 'getAttribute')"

humphd (approved):

functions/api/share/[[user_id]].ts
Comment (tarasglek) [reply functions/api/share/[[user_id]].ts:43]: `const title = getTitle?.getAttribute("content") ?? "No Title"; ` would've worked for this

PR #575, opened 2024-04-09
Disable message editing on shared chats

### LINKED ISSUE #570: Disable message editing on shared chats
I often get into a problem where I open a shared chat..and start modifying it BEFORE pressing duplicate...These changes are then all lost when I duplicate :(

mingming-ma:
## Description 

Due to the special behavior of the JavaScript language, the previous logic `message.readonly && disableEdit` will be `undefined` if `disableEdit` is `undefined`, so `!disableEdit` still `true` and display the edit button in shared chat.

To fix the issue, and given `message.readonly` comes from`SerializedChatCraftMessage`, we can give `message.readonly` higher priority, as `message.readonly ?? disableEdit`

## Result

Only display the copy button in the
[shared chat](https://mingming-sharedchats-570.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/3Y5Hrv-sIB2GSfxV-_32s)

image

Fixes #570

### DIFF
--- a/src/components/Message/AiMessage.tsx
+++ b/src/components/Message/AiMessage.tsx
@@ -159,7 +159,7 @@ function AiMessage(props: AiMessageProps) {
       onRetryClick={retrying ? undefined : handleRetryClick}
       onDeleteClick={retrying ? undefined : props.onDeleteClick}
       disableFork={retrying}
-      disableEdit={retrying}
+      disableEdit={message.readonly ?? retrying}
     />
   );
 }
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -430,7 +430,7 @@ function MessageBase({
                       onClick={() => onEditingChange(!editing)}
                     />
                   )}
-                  {onDeleteClick && (
+                  {!disableEdit && onDeleteClick && (
                     <IconButton
                       variant="ghost"
                       icon={<TbTrash />}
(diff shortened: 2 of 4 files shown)


kliu57 (approved):
Just tested, looks good to me. I only see the copy button now when I visit a shared chat and not the others. The code is simple and looks fine as well.

rjwignar (approved):
I reviewed the code and the changes and it looks good! All possible message types appear to be accounted for (their edit buttons are disabled if on a shared chat).

When I initially saw the issue, one solution I thought of was persisting changes to the shared chat, though that ideas has some issues (e.g. would other users see those changes?).

I think what you did here is a better solution. It fixes #570 by requiring users to duplicate a chat first in order to edit and experiment with the chat.

tarasglek:
This is fantastic. While at it, can you replace options button with duplicate to completely fix this UI?...none of the other options make sense and this will be less confusing

mingming-ma:
@tarasglek I prefer to keep the options button. For me I always use the `New window` to have a quick new chat, some thoughts inspired by the shared chat (or a total different one) that does not need the history. I think there are some users like me. Can we keep it 😄

tarasglek:
> @tarasglek I prefer to keep the options button. For me I always use the `New window` to have a quick new chat, some thoughts inspired by the shared chat (or a total different one) that does not need the history. I think there are some users like me. Can we keep it 😄

for now

PR #581, opened 2024-04-11
Fix Web Handlers not executing with Re-ask option

### LINKED ISSUE #541: Clicking Re-ask button should check if the updated text needs to be handled by a Web Handler
As mentioned by @tarasglek in https://github.com/tarasglek/chatcraft.org/pull/519#issuecomment-2015725543, if the user edits a message and clicks on re-ask, we should be checking if the message matches a Web Handler and execute that instead of sending message to LLM.
image

Amnish04:
In #519, I implemented initial support for Web Handlers, which extend ChatCraft's URL handling capabilities.

However, @tarasglek noticed that when a non-matching prompt is edited to a valid url pattern that is configured in Web Handlers configuration, prompt is still sent to LLM instead of invoking the corresponding Web Handler.

I researched a bit and found that the reason for this to happen was that the `Resubmit` handler function for the message form was not passing the prompt text  to the `onPrompt` function.

This is why the handler wasn't able to determine any specifics about the prompt. I am now passing the text to that handler so `onPrompt` in `ChatBase` knows that it needs to be passed to a Web Handler.

**Before:**
```ts
onResubmitClick={async () => {
  await deleteMessages(message.id, "after");
  onPrompt();
}}
```

**After:**
```ts
onResubmitClick={async (promptText?: string) => {
  await deleteMessages(message.id, "after");
  onPrompt(promptText);
}}
```

This fixes #541

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -80,7 +80,7 @@ export interface MessageBaseProps {
   isLoading: boolean;
   hidePreviews?: boolean;
   onPrompt?: (prompt?: string) => void;
-  onResubmitClick?: (promtText?: string) => void;
+  onResubmitClick?: (promptText?: string) => void;
   onDeleteBeforeClick?: () => void;
   onDeleteClick?: () => void;
   onDeleteAfterClick?: () => void;
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -80,7 +80,7 @@ export interface MessageBaseProps {
   isLoading: boolean;
   hidePreviews?: boolean;
   onPrompt?: (prompt?: string) => void;
-  onResubmitClick?: () => void;
+  onResubmitClick?: (promtText?: string) => void;
   onDeleteBeforeClick?: () => void;
   onDeleteClick?: () => void;
   onDeleteAfterClick?: () => void;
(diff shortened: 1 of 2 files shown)


src/components/Message/MessageBase.tsx
```diff
@@ -80,7 +80,7 @@ export interface MessageBaseProps {
   isLoading: boolean;
   hidePreviews?: boolean;
   onPrompt?: (prompt?: string) => void;
-  onResubmitClick?: () => void;
+  onResubmitClick?: (promtText?: string) => void;
```
rjwignar, line 83: Should this be promptText instead?
I see only one instance when I run `git grep promtText`:
image

rjwignar (changes requested):
@Amnish04 The fix looks great and I can confirm it works:
Production Bug (re-asking with a valid PR URL sends message to LLM):
webHandlerBugProd

Observed Fix (re-asking with a valid PR URL prompts web handler):
webHandlerBugFix

Before I approve this, can you address my comment?

src/components/Message/MessageBase.tsx
Amnish04: Oh right that's a typo, will fix that now

rjwignar (approved):

rjwignar:
@Amnish04 Thanks for clearing that up!
LGTM!

Amnish04:
@rjwignar Great, thanks for reviewing!

PR #582, opened 2024-04-12
Refactor PasswordInput component to follow Chakra's design system

### LINKED ISSUE #553: PasswordInput component refactor
For discussion

Right now we are taking a single inputSize param in `PasswordInput` component.

When it is "sm", then we make the padding/sizes/colors to match the small password input field that we currently use in the providers table.

When the param is not specified, we make the padding/sizes/color to match our regular password input field we use in instructions page.

Proposed by @Amnish04 that instead of hardcoding all the prop values like size, padding, color, etc, that we all all those props to the PasswordInput component so that it can be dynamically altered.

Amnish04:
In #530, @kliu57 added support for custom AI providers and in implementing the UI for providers selection, she made changes to the existing `RevealablePasswordInput` component to make the appearance ideal for her use case.

However, I felt that the code had some issues.

1. Programatically determining styles based on other styles and hard coding values inside component caused tight coupling between different styles making it hard to reuse in other cases in the future. (Would need to add more if statements every time a new appearance was required)
2. In a way, it overrode Chakra's in-built styling system which is not good.

In this PR, I have tried to refactor that part to make use of Chakra's own design system as much as possible turning it into a more generic wrapper around normal Input. `PasswordInput` now behaves more like a component provided by Chakra itself since most of the styles/props are being mapped directly.

This fixes #553

### DIFF
--- a/src/components/Message/AppMessage/Instructions.tsx
+++ b/src/components/Message/AppMessage/Instructions.tsx
@@ -156,6 +156,7 @@ function Instructions(props: MessageBaseProps) {
               <Flex gap={4} align="center">
                 <PasswordInput
                   flex="1"
+                  size={"md"}
                   type="password"
                   name="openai-api-key"
                   bg="white"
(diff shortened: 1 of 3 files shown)


kliu57 (approved):
I just tested it and it looks and behaves the same as it used to. So your refactoring works well. If possible could you and @rjwignar merge your PRs for the password field sometime soon because I need to put in some hotfixes and that would allow me to not deal with so many merge conflicts.

PR #584, opened 2024-04-12
Fix for minor bugs in instructions page and prefs modal

### LINKED ISSUE #583: Minor bugs in instructions and prefs modal
I discovered some bugs that arisen after the custom provider implementation

**Bug 1** - Instructions page provider dropdown does not include the current provider when it is a custom-made provider

Steps to reproduce:
1. Create custom provider
2. Set custom provider to current provider
3. Clear the custom provider's api key
4. Close Prefs Modal, you should now see instructions page (if you don't scroll to bottom)

Expected Result: In instructions page, the dropdown should include and be currently set to the current provider which is the custom provider. And if you enter the correct api key and press Save button it should work.

Actual Result: The dropdown is not set to the current provider which is custom provider. And the Save button does not work since the password field is mapped to the custom provider but the provider dropdown won't show it.

**Bug 2** - When switching to a provider with a saved key

1. Store valid keys in two of your providers
2. Set one of them as your current provider.
3. Clear the key
4. Close Prefs Modal and see Instructions page (if you don't scroll to bottom)
5. In Instructions page, use the dropdown to select the other provider that has a saved key

Expected Result: Just show the key but don't load the next page right away

Actual Result: Goes to the next page right away without validating the key

**Bug 3** - Password field still focused after closing and reopening the Prefs Modal, validation issues

1. Clear the api key of current provider
7. Close the Prefs Modal
8. Reopen the Prefs Modal
9. Paste the correct api key

Expected Result: Validation should automatically occur and invalid key message should disappear. Also the field should not be focused when we reopen the modal.

Actual Result: Validation did not occur and invalid key message still there. Field is focused the moment we reopen the modal.

**Bug 4** - We need to ban browsers from asking to save the key

This is important since in the prefs table we have many api key fields and if user use Chrome to save their key it won't get repopulated in the correct field. Also, we are storing the key in localStorage so there is no need for users to save their key to their browser.

kliu57:
Fixes #583 

See the above issue for details on how to reproduce each bug in production.

**How each bug was fixed:**

**Bug 1:**

Fixed by having the dropdown display a list that is made up of `supportedProviders` and `settings.providers` merged together (without duplicates). The merge results in `providersList`. (Prior to this change, the dropdown displays only `supportedProviders` which is wrong since that does not include our custom created providers. The merge of the two lists is done the same way as @Rachit1313 did in `PromptSendButton.tsx`.

**Bug 2:**

Fixed by introducing state variable `selectedProvider` in `Instructions.tsx` so we can keep track of the selected provider's api key. This allows us to delay the setting of the currentProvider's api key until after we click the Save button. Before this change, the moment we switch to a provider with a saved key, we leave the instructions page immediately (due to the fact that we set `currentProvider.apiKey` and instructions only appears when that is blank). This is not ideal, since the key could be invalid. So we do not want to leave the instructions page yet, and this state variable allows us to delay setting `settings.currentProvider.apiKey`.

**Bug 3:**

Fixed by one line addition to `PreferencesModel.tsx`

**Bug 4:**

Fixed by banning browsers from saving api keys as password. This is done by removing `type={show ? "text" : "password"}` because when `type` is `password` it prompts browsers to use their password manager. Instead of using `type = "password"` to disguise the password as asterix, I am now using:

```
        sx={{
          WebkitTextSecurity: show ? "none" : "disc",
        }}
```

### DIFF
--- a/src/components/Message/AppMessage/Instructions.tsx
+++ b/src/components/Message/AppMessage/Instructions.tsx
@@ -1,4 +1,4 @@
-import { FormEvent, memo, useState, useCallback, ChangeEvent } from "react";
+import { FormEvent, memo, useState, useEffect } from "react";
 import {
   Button,
   Box,
@@ -13,11 +13,12 @@ import {
 
 import MessageBase, { type MessageBaseProps } from "../MessageBase";
 import { ChatCraftAppMessage } from "../../../lib/ChatCraftMessage";
-import { nameToUrlMap, providerFromUrl, supportedProviders } from "../../../lib/providers";
+import { providerFromUrl, supportedProviders } from "../../../lib/providers";
 import { OpenRouterProvider } from "../../../lib/providers/OpenRouterProvider";
 import PasswordInput from "../../PasswordInput";
 import { useSettings } from "../../../hooks/use-settings";
 import { FreeModelProvider } from "../../../lib/providers/DefaultProvider/FreeModelProvider";
+import { ProviderData } from "../../../lib/ChatCraftProvider";
 
 const ApiKeyInstructionsText = `## Getting Started with ChatCraft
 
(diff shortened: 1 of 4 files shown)


Rachit1313:
The above change looks good to me!. I tested and it works fine but I believe we have another bug we should fix, maybe in a separate PR i.e 
Steps:
1. save key for openAI or any other provider
2. switch to openrouter as current provider
3. edit the API key for openAI to lets say 1 or 2 char / break the key
4. we still see option in ask menu for the provider and see an error when trying to switch to it

I believe we should implement validation onChange of api key.

Rachit1313 (approved):

WangGithub0 (approved):
It looks good to me!

mingming-ma (approved):
LGTM!

PR #589, opened 2024-04-14
Add Accordion to hide Image Compression section

### LINKED ISSUE #522: Advanced Image Compression settings should be hidden by default
#513 added settings in Preferences Modal for configuring image compression.

This makes the modal too long.
image

We should have those settings in an **Expansion Bar** so the user can reveal those settings if needed

menghif:
Fixes #522 

This PR moves the Image Compression section of the User Settings into an Accordion that is hidden by default.

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -38,6 +38,11 @@ import {
   Tooltip,
   Tr,
   VStack,
+  Accordion,
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel,
+  AccordionIcon,
 } from "@chakra-ui/react";
 import { ChangeEvent, RefObject, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useDebounce } from "react-use";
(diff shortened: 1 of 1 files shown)


mingming-ma (approved):
LGTM!

src/components/PreferencesModal.tsx
```diff
@@ -953,91 +958,107 @@ function PreferencesModal({ isOpen, onClose, finalFocusRef }: PreferencesModalPr
...
-                    <FormControl>
-                      <FormLabel>
-                        Maximum image dimension: {settings.maxImageDimension} (px)
+              <FormControl>
+                <Accordion allowToggle>
+                  <AccordionItem>
+                    <AccordionButton>
+                      <FormLabel as="legend" flex="1">
+                        Image Compression
                       </FormLabel>
```
Amnish04, line 967: We should probably not wrap the `Accordion` in a `FormControl`.
And since we do not have a `fieldset` anymore, we should not have a `legend` either. 

Invalid HTML

We could do something like this as suggested in [Chakra Docs](https://v2.chakra-ui.com/docs/components/accordion#usage)

```tsx
<AccordionButton>
  <Box as='span' flex='1' textAlign='left'>
    Section 1 title
  </Box>
  <AccordionIcon />
</AccordionButton>
```

menghif:
Closing this in favour of the bigger changes in #592

PR #591, opened 2024-04-15
Fix the loading spinner does not disappear when image generation fails

### LINKED ISSUE #590: The loading spinner does not disappear when image generation fails
When trying to generate images and an error occurs, the alert with loading spinner does not disappear falsely suggesting something is still going on in the background.
image

mingming-ma:
Fixes #590 

Make sure `closeLoading(alertId);` will always be called, so the loading spinner can disappear whether an error is thrown or not.


Amnish04 (approved):
@mingming-ma Works well. Thanks for the fix!

PR #577, opened 2024-04-10
adjusted toast message width for mobile experience

rjwignar:
This fixes #569 
I've been working with ChakraUI's [Toast props](url) and I've found a way to make toast messages fit the viewport when mobile is detected (screen width <= 480px) , but it seems the second end of my ternary operator overrides the default width we currently use for desktop.

PR #595, opened 2024-04-16
Revert "adjusted toast message width for mobile experience"

### LINKED ISSUE #586: Error alert toast bug
A new bug was introduced from commit 654ee24 from pr #577 

### Commits in order by date:
- d8466be (has bug)
- 62cd016 (has bug)
- 7cc1e2f (has bug)
- 1863245 (has bug)
- 70181df (has bug)
- 654ee24 (has bug) <---
- adcfd2d (does not have bug)
- 267558a (does not have bug)

Tested these listed commits using `git checkout commitCode` and then `pnpm dev`
Using the preview url is not reliable (perhaps because the preview url only had the branch code and not the merged main code)

### Steps to reproduce bug:

1. Open User Settings Modal
2. Set OpenAI api key to a valid key and also set it to your current provider
3. Close User Settings Modal by clicking out
4. Open User Settings Modal
5. Change to api key to something invalid like, "a"
6. Wait until validation is done, telling you its invalid
7. Close User Settings Modal by clicking out
8. Open User Settings Modal and change the key to a valid key

**Expected Result:**

Key validation is successful. And when you send a message to chatgpt it is fine. No odd behaviour.

**Actual result:**

You get two error alert toasts before the validation even completes. Then the validation completes successfully. 
And when you send a message to chatgpt, the message sends fine and chatgpt replies fine, but you also get many error alert toasts saying your api key is wrong and your api key is "a".

rjwignar:
Reverts tarasglek/chatcraft.org#577

This code currently introduces a bug described in #586.
After attempting a bugfix in #593, additional bugs were still found.

Removing this code removes these bugs

This fixes #586

### DIFF
--- a/src/hooks/use-alert.ts
+++ b/src/hooks/use-alert.ts
@@ -1,6 +1,6 @@
 import { useCallback } from "react";
 import { useToast } from "@chakra-ui/react";
-import useMobileBreakpoint from "../hooks/use-mobile-breakpoint";
+
 export type AlertArguments = {
   // Use `id` if you want to avoid duplicate alerts showing
   id?: string;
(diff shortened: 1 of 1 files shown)


kliu57 (approved):
Tested and confirmed that reverting this code fixes the bug.

PR #597, opened 2024-04-17
Enhancement - only save keys to settings.providers if validated

### LINKED ISSUE #596: Only valid keys should be stored in settings.providers array
@Rachit1313 brought this enhancement up to me

Currently we are storing the new api key as soon as it's changed, even if it is invalid.

We should only store the providers with validated api keys to settings.providers

This makes it so in the dropdown below we will only be listing useable providers and not including any provider that has an invalid key.

image

kliu57:
Fixes #596 

**Summary:** No longer saving providers or their api keys to `settings.providers` array unless the api key is validated. If an invalid or blank is entered, the provider is REMOVED from `settings.providers`.

**Important coding points:**

- Removed the placeholder text in the api key field when you create a new custom provider row (placeholder text was showing up as asterix)
- Removed the `useDebounce` code, it wasn't working and is not worth the trouble to fix. `useDebounce` is sort of like a `useEffect`, but which allows us to wait until the user stops typing to validate the key. However, we do not expect anyone to be typing out a 32 character api key anyways, so opted to remove.
- The keys are stored in `tableProviders` and only when they are validated they are stored to `settings.providers`
- In these scenarios, `tableProviders` is synced with `settings.providers`: (when "syncing" happens, those invalid and blank api keys will disappear)
   - When modal first opens
   - When modal closes
   - When current provider is set (need to sync to update the checkmark)
   - When a custom provider is deleted (need to sync to show the row deleted)
   - When a custom provider is added (need to sync to show the new row)
- **Note:** If you create a custom provider and then later set the key to something invalid, when you close the modal the custom provider row will disappear since we only save it if it has a valid key.

I put in some `console.log` comments for @rjwignar ; those will be removed before I merge to main

### DIFF
--- a/src/components/PreferencesModal.tsx
+++ b/src/components/PreferencesModal.tsx
@@ -40,7 +40,6 @@ import {
   VStack,
 } from "@chakra-ui/react";
 import { ChangeEvent, RefObject, useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { useDebounce } from "react-use";
 
 import { capitalize } from "lodash-es";
 import { FaCheck } from "react-icons/fa";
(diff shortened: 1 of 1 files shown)


kliu57:
@rjwignar @Rachit1313 Sorry not yet ready for review. Will comment again when it is.

kliu57:
@rjwignar @Rachit1313 Now it is ready for review

WangGithub0:
I found the error color changed, is this the expected?
production:
image

priview
image

kliu57:
> I found the error color changed, is this the expected? production: image
> 
> priview image

The color is different because the message is different. The orange one happens when you try to add a new custom provider, but you put an invalid key. So it says "Provider not added" I used a warning instead of error so it can automatically disappear, I feel it is more convenient. The red one is when you change your current provider key to an invalid one.

Rachit1313 (approved):
Looks good to me.

tarasglek:
I think Error was more correct here..as it's fairly fatal that you can't validate your key

kliu57:
> I think Error was more correct here..as it's fairly fatal that you can't validate your key

@tarasglek Ok, I modified the code so that users trying to add a new custom provider but gave an invalid url or invalid key will get the "error" type alert instead of the warning

kliu57:
@WangGithub0 I changed the alert type when you create new custom providers with erroneous info to error instead of warning. 

Also I found an issue where the placeholder text of the `PasswordInput` was displaying as asterix and I fixed it

kliu57:
@WangGithub0 also I found an issue where the placeholder text was showing up as asterix after I made the change in last hotfix PR to the `PasswordInput` component to stop browsers from asking users to save their password. Could not find an easy fix for this so for now I removed the "API Key" placeholder text for the password field.

WangGithub0 (approved):
It looks good to me.

PR #598, opened 2024-04-17
Disable edit summary for shared chats

### LINKED ISSUE #160: Disable edit/delete for shared chats
I was looking at https://chatcraft.org/c/tarasglek/cbo8qmybUrF9MwTIt8XBa, a shared chat I don't own, and noticed that the edit and delete buttons/icons are still there.  They don't do anything, since the chat is `readonly` (all shared chats are loaded from JSON vs. the db, so are read-only), but it's weird to have the user think that they can use them.

Places that need a fix:

1. Edit icon beside the chat's summary text
2. Edit and Delete icons when hovering a message
3. Edit and Delete menu options for a message

Screenshot 2023-07-10 at 2 17 42 PM

WangGithub0:
before:
image

after:
image

closes #160

### DIFF
--- a/src/Chat/ChatHeader.tsx
+++ b/src/Chat/ChatHeader.tsx
@@ -154,7 +154,7 @@ function ChatHeader({ chat }: ChatHeaderProps) {
                       {title}
                     </Link>
                   </Text>
-                  {settings.currentProvider.apiKey && (
+                  {!chat.readonly && settings.currentProvider.apiKey && (
                     <IconButton
                       variant="ghost"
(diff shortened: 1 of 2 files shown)


kliu57 (changes requested):
Looks good, but maybe you could also get rid of the divider

image

If it is difficult to do, then it probably is not worth it, then can just skip this

kliu57 (approved):
LGTM

Amnish04 (approved):
@WangGithub0 Looks good to me!

PR #600, opened 2024-04-17
Added icon to point towards currently selected model and incremental search

### LINKED ISSUE #502: Show currently selected model, add incremental search
It would be nice to have a checkbox near current model
image

It would also be nice to have incremental search in that menu, it's crazy finding a model on openrouter these days, esp with all the variation in naming

Rachit1313:
Closes #502

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll": true
+    "source.fixAll": "explicit"
   },
   "files.eol": "\n",
   "files.insertFinalNewline": true
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll": "explicit"
+    "source.fixAll": true
   },
   "files.eol": "\n",
   "files.insertFinalNewline": true
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -21,7 +21,7 @@ import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
 import { MdVolumeUp, MdVolumeOff, MdOutlineChevronRight } from "react-icons/md";
-import { useEffect, useMemo, useRef, useState, type KeyboardEvent } from "react";
+import { useMemo, useRef, useState, type KeyboardEvent } from "react";
 import useAudioPlayer from "../../hooks/use-audio-player";
 import { usingOfficialOpenAI } from "../../lib/providers";
 import { useDebounce } from "react-use";
@@ -50,10 +50,6 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
     600,
     [searchQuery]
   );
-
-  useEffect(() => {
-    inputRef.current?.focus();
-  }, []);
 
   const providersList = {
     ...settings.providers,
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -210,6 +210,11 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
                   key={model.id}
                   onClick={() => setSettings({ ...settings, model })}
                 >
+                  {settings.model.id === model.id ? (
+                    <MdOutlineChevronRight style={{ marginRight: "4px" }} />
+                  ) : (
+                    <span style={{ width: "24px", display: "inline-block" }} />
+                  )}
                   {model.prettyModel}
                 </MenuItem>
               ))}
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -12,7 +12,6 @@ import {
   Input,
   InputGroup,
   InputLeftElement,
-  useDisclosure,
 } from "@chakra-ui/react";
 import { TbChevronUp, TbSend, TbSearch } from "react-icons/tb";
 import { FreeModelProvider } from "../../lib/providers/DefaultProvider/FreeModelProvider";
@@ -36,35 +35,11 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
   const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("");
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
-  const { isOpen, onOpen, onClose } = useDisclosure();
   const inputRef = useRef<HTMLInputElement>(null);
 
   const isTtsSupported = useMemo(() => {
     return !!models.filter((model) => model.id.includes("tts"))?.length;
   }, [models]);
-
-  useEffect(() => {
-    if (isOpen) {
-      // Optionally, add a slight delay if needed to ensure the input is ready to be focused
-      setTimeout(() => {
-        inputRef.current?.focus();
-      }, 100);
-    }
-  }, [isOpen]);
-
-  const onStartTyping = (e: KeyboardEvent<HTMLElement>) => {
-    // Check if the inputRef is current and the input is not already focused
-    if (inputRef.current && document.activeElement !== inputRef.current) {
-      // Don't handle the keydown event more than once
-      e.preventDefault();
-      // Make sure we are focused on the input element
-      inputRef.current.focus();
-      // Ignore control keys
-      const char = e.key.length === 1 ? e.key : "";
-      // Set the initial character in the input so we don't lose it
-      setSearchQuery(char);
-    }
-  };
 
   const { clearAudioQueue } = useAudioPlayer();
 
@@ -72,7 +47,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
     () => {
       setDebouncedSearchQuery(searchQuery);
     },
-    500,
+    600,
     [searchQuery]
   );
 
@@ -87,14 +62,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
 
   return (
     <ButtonGroup variant="outline" isAttached>
-      <Menu
-        placement="top"
-        strategy="fixed"
-        closeOnSelect={false}
-        offset={[-90, 0]}
-        onOpen={onOpen}
-        onClose={onClose}
-      >
+      <Menu placement="top" strategy="fixed" closeOnSelect={false} offset={[-90, 0]}>
         <IconButton
           type="submit"
           size="lg"
@@ -153,12 +121,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
           title="Choose Model"
           icon={<TbChevronUp />}
         />
-        <MenuList
-          maxHeight={"70vh"}
-          overflowY={"auto"}
-          zIndex={theme.zIndices.dropdown}
-          onKeyDownCapture={onStartTyping}
-        >
+        <MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
           <MenuGroup title="Models">
             <InputGroup>
               <InputLeftElement pointerEvents="none">
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -61,7 +61,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
     () => {
       setDebouncedSearchQuery(searchQuery);
     },
-    250,
+    500,
     [searchQuery]
   );
 
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -41,6 +41,20 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
     return !!models.filter((model) => model.id.includes("tts"))?.length;
   }, [models]);
 
+  const onStartTyping = (e: KeyboardEvent<HTMLElement>) => {
+    // Check if the inputRef is current and the input is not already focused
+    if (inputRef.current && document.activeElement !== inputRef.current) {
+      // Don't handle the keydown event more than once
+      e.preventDefault();
+      // Make sure we are focused on the input element
+      inputRef.current.focus();
+      // Ignore control keys
+      const char = e.key.length === 1 ? e.key : "";
+      // Set the initial character in the input so we don't lose it
+      setSearchQuery(char);
+    }
+  };
+
   const { clearAudioQueue } = useAudioPlayer();
 
   useDebounce(
@@ -121,7 +135,12 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
           title="Choose Model"
           icon={<TbChevronUp />}
         />
-        <MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
+        <MenuList
+          maxHeight={"70vh"}
+          overflowY={"auto"}
+          zIndex={theme.zIndices.dropdown}
+          onKeyDownCapture={onStartTyping}
+        >
           <MenuGroup title="Models">
             <InputGroup>
               <InputLeftElement pointerEvents="none">
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -18,7 +18,7 @@ import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
 import { MdVolumeUp, MdVolumeOff, MdOutlineChevronRight } from "react-icons/md";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import useAudioPlayer from "../../hooks/use-audio-player";
 import { usingOfficialOpenAI } from "../../lib/providers";
 
@@ -141,6 +141,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
 }
 
 function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
+  const [searchQuery, setSearchQuery] = useState("");
   const { settings, setSettings } = useSettings();
   const { models } = useModels();
   const isTtsSupported = useMemo(() => {
@@ -202,8 +203,18 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
         />
         <MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
           <MenuGroup title="Models">
+            <input
+              type="text"
+              placeholder="Search models..."
+              value={searchQuery}
+              onChange={(e) => setSearchQuery(e.target.value)}
+              style={{ margin: "10px", padding: "5px", width: "calc(100% - 20px)" }}
+            />
             {models
               .filter((model) => !usingOfficialOpenAI() || model.id.includes("gpt"))
+              .filter((model) =>
+                model.prettyModel.toLowerCase().includes(searchQuery.toLowerCase())
+              )
               .map((model) => (
                 <MenuItem
                   closeOnSelect={true}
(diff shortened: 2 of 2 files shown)


kliu57 (changes requested):
If you type something in the search field and then erase it with backspace, before you finish erasing it the field will lose focus. Can you do something so it doesn't keep losing focus?

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll": true
+    "source.fixAll": "explicit"
```
rjwignar, line 8: Do we also want to change this setting?

Rachit1313: No, Thanks for pointing out. I'll just remove it

rjwignar (changes requested):
I like these features @Rachit1313.
I like that this shows which model is currently selected, and I like the incremental search too.

I just have the same concern as @kliu57
It seems this bug only occurs after you narrow your possible entries, and then expand the list of possible matches by pressing **Backspace** a few times.

Notice how as I narrow my possible matches by typing out `gpt-4-1106-` the text entry remains in focus, but as soon as I hit **Backspace** until `gpt-4-`, the list of possible matches increases and the text entry loses focus
600_incrementalSearchBug

rjwignar:
@Rachit1313 I found this Stackoverflow post which might be helpful:
https://stackoverflow.com/questions/42573017/in-react-es6-why-does-the-input-field-lose-focus-after-typing-a-character

Rachit1313:
> I like these features @Rachit1313. I like that this shows which model is currently selected, and I like the incremental search too.
> 
> I just have the same concern as @kliu57 It seems this bug only occurs after you narrow your possible entries, and then expand the list of possible matches by pressing **Backspace** a few times.
> 
> Notice how as I narrow my possible matches by typing out `gpt-4-1106-` the text entry remains in focus, but as soon as I hit **Backspace** until `gpt-4-`, the list of possible matches increases and the text entry loses focus 600_incrementalSearchBug [ 600_incrementalSearchBug ](https://private-user-images.githubusercontent.com/78163326/323426475-c36eb888-e841-49be-b3e5-729fb117c405.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTM0MDA4MjAsIm5iZiI6MTcxMzQwMDUyMCwicGF0aCI6Ii83ODE2MzMyNi8zMjM0MjY0NzUtYzM2ZWI4ODgtZTg0MS00OWJlLWIzZTUtNzI5ZmIxMTdjNDA1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDE4VDAwMzUyMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRhODJjZTVjMmMzYTI4MGFkYWZkZWIxMGM5NGM2OTA4NmZkNTkwNWU0NTZjM2Y1MmUxZjlmZDM2MjRlMzhiNGUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.UMvlJeMqloAVKwWHIffdve9P98pilXDak4l5eYu8JHg) [ ](https://private-user-images.githubusercontent.com/78163326/323426475-c36eb888-e841-49be-b3e5-729fb117c405.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTM0MDA4MjAsIm5iZiI6MTcxMzQwMDUyMCwicGF0aCI6Ii83ODE2MzMyNi8zMjM0MjY0NzUtYzM2ZWI4ODgtZTg0MS00OWJlLWIzZTUtNzI5ZmIxMTdjNDA1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDE4VDAwMzUyMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRhODJjZTVjMmMzYTI4MGFkYWZkZWIxMGM5NGM2OTA4NmZkNTkwNWU0NTZjM2Y1MmUxZjlmZDM2MjRlMzhiNGUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.UMvlJeMqloAVKwWHIffdve9P98pilXDak4l5eYu8JHg)

Yeah, I see that this behavior is getting caused by re rendering of the Menu when we find more/ less possible matches.

Rachit1313:
> @Rachit1313 I found this Stackoverflow post which might be helpful: https://stackoverflow.com/questions/42573017/in-react-es6-why-does-the-input-field-lose-focus-after-typing-a-character

tried many solutions, nothing works for me.
Tried
1. use disclosure method
2. switching the input with ChakraUI's Input
3. autoFocus property
4. useEffect to manually focus after setting inputRef

Rachit1313:
@humphd/ @Amnish04 do you know anything that might help here?

tarasglek:
a) This is so freaking awesome!
b) You want to debounce like elsewhere in code to not search while user is typing..eg give user 100ms or so to finish their text before searching
c) you likely lose focus on backspace likely because react is destroying and re-creating the menu..hacky solution is to do setTimeout, get reference to raw element and focus that when it's recreated...but really talk to chatcraft re your approach nd this problem

kliu57:
@Rachit1313 I know you need a bit of help so here is the code to do the useDebounce since I am familiar with it. Already tested on my local so it works.

```
import { useDebounce } from "react-use";

function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
  // Do the same here as I did in DesktopPromptSendButton
}

function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
  const [searchQuery, setSearchQuery] = useState("");
  const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("");

  // I used 1000 milliseconds just as a test to make sure it is working. You can change it to 500 or something I think it may be enough time for a user to finish typing, I actually think 100 milliseconds was not enough time but do some tests and see
  useDebounce(
    () => {
      setDebouncedSearchQuery(searchQuery);
    },
    1000,
    [searchQuery]
  );
  
  ...
  // notice I changed searchQuery to debouncedSearchQuery below in the {models... } part
            <input
              type="text"
              placeholder="Search models..."
              value={searchQuery}
              onChange={(e) => setSearchQuery(e.target.value)}
              style={{ margin: "10px", padding: "5px", width: "calc(100% - 20px)" }}
            />
            {models
              .filter((model) => !usingOfficialOpenAI() || model.id.includes("gpt"))
              .filter((model) =>
                model.prettyModel.toLowerCase().includes(debouncedSearchQuery.toLowerCase())
              )
              .map((model) => (
                <MenuItem
                  closeOnSelect={true}
                  key={model.id}
                  onClick={() => setSettings({ ...settings, model })}
                >
                  {settings.model.id === model.id ? (
                    <MdOutlineChevronRight style={{ marginRight: "4px" }} />
                  ) : (
                    <span style={{ width: "24px", display: "inline-block" }} />
                  )}
                  {model.prettyModel}
                </MenuItem>
              ))}
  ...
}
```

More information I uncovered while testing the backspace problem. It is not a backspace problem but a general overall focus problem for the input field. Try clicking on the input search field. It now has focus. Now hover your mouse away from that area and hover on to one of the models. It immediately loses focus. This is the root of the problem. We thought the problem was from the backspace, but actually all the backspace does was change the search query which changed the size of the dropdown, which made your mouse hover over one of the models or hover on something else and the search field lost focus immediately.

I don't know why the search field is losing focus when you hover away, gpt does not seem to know. Maybe look into what code is triggered when you hover over the models? Maybe that code is shifting the focus to something else and thus the focus of the search input field is lost?

tarasglek:
If the problem is mouseovers, this is probably an artifact of how react-menu works

Rachit1313:
> If the problem is mouseovers, this is probably an artifact of how react-menu works

Yeah I see, Even the solution @kliu57 gave it not helping us to solve the problem completely. If its a problem of how ChakraMenu is setup, how do we fix this?

Rachit1313:
The above method would help us stay in focus until we fully erase the text we want while backspacing and not remove focus just after a single match found but I believe this still have some issues as it delays the response time if that matters.

humphd:
I love this, great work @Rachit1313!  I pulled it locally and made some fixes, which you can use:

```diff
diff --git a/src/components/PromptForm/PromptSendButton.tsx b/src/components/PromptForm/PromptSendButton.tsx
index e9ddd5b..2dae5e9 100644
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -9,8 +9,11 @@ import {
   Tooltip,
   MenuDivider,
   MenuGroup,
+  Input,
+  InputGroup,
+  InputLeftElement,
 } from "@chakra-ui/react";
-import { TbChevronUp, TbSend } from "react-icons/tb";
+import { TbChevronUp, TbSend, TbSearch } from "react-icons/tb";
 import { FreeModelProvider } from "../../lib/providers/DefaultProvider/FreeModelProvider";
 
 import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
@@ -18,7 +21,7 @@ import { useSettings } from "../../hooks/use-settings";
 import { useModels } from "../../hooks/use-models";
 import theme from "../../theme";
 import { MdVolumeUp, MdVolumeOff, MdOutlineChevronRight } from "react-icons/md";
-import { useMemo, useState } from "react";
+import { useMemo, useRef, useState, type KeyboardEvent } from "react";
 import useAudioPlayer from "../../hooks/use-audio-player";
 import { usingOfficialOpenAI } from "../../lib/providers";
 import { useDebounce } from "react-use";
@@ -106,7 +109,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
         />
         <MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
           <MenuGroup title="Models">
-            <input
+            <Input
               type="text"
               placeholder="Search models..."
               value={searchQuery}
@@ -165,15 +168,30 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
   const isTtsSupported = useMemo(() => {
     return !!models.filter((model) => model.id.includes("tts"))?.length;
   }, [models]);
+  const inputRef = useRef<HTMLInputElement>(null);
 
   useDebounce(
     () => {
       setDebouncedSearchQuery(searchQuery);
     },
-    500,
+    250,
     [searchQuery]
   );
 
+  const onStatTyping = (e: KeyboardEvent<HTMLElement>) => {
+    // Check if the inputRef is current and the input is not already focused
+    if (inputRef.current && document.activeElement !== inputRef.current) {
+      // Don't handle the keydown event more than once
+      e.preventDefault();
+      // Make sure we are focused on the input element
+      inputRef.current.focus();
+      // Ignore control keys
+      const char = e.key.length === 1 ? e.key : "";
+      // Set the initial character in the input so we don't lose it
+      setSearchQuery(char);
+    }
+  };
+
   const { clearAudioQueue } = useAudioPlayer();
 
   const providersList = {
@@ -227,15 +245,26 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
           title="Choose Model"
           icon={<TbChevronUp />}
         />
-        <MenuList maxHeight={"70vh"} overflowY={"auto"} zIndex={theme.zIndices.dropdown}>
+        <MenuList
+          maxHeight={"70vh"}
+          overflowY={"auto"}
+          zIndex={theme.zIndices.dropdown}
+          onKeyDownCapture={onStatTyping}
+        >
           <MenuGroup title="Models">
-            <input
-              type="text"
-              placeholder="Search models..."
-              value={searchQuery}
-              onChange={(e) => setSearchQuery(e.target.value)}
-              style={{ margin: "10px", padding: "5px", width: "calc(100% - 20px)" }}
-            />
+            <InputGroup>
+              <InputLeftElement pointerEvents="none">
+                <TbSearch />
+              </InputLeftElement>
+              <Input
+                ref={inputRef}
+                type="text"
+                variant="ghost"
+                placeholder="Search models..."
+                value={searchQuery}
+                onChange={(e) => setSearchQuery(e.target.value)}
+              />
+            </InputGroup>
             {models
               .filter((model) => !usingOfficialOpenAI() || model.id.includes("gpt"))
               .filter((model) =>
```

The main changes:

- add a key down handler to listen for the user typing, and when they do, add the first character to the input and then let your exiting code take over
- reduce the debounce so updates happen sooner
- use Chakra's controls to make it a bit more integrated

Rachit1313:
I believe for mobile, we can have debounce itself of 600 as probably we would want the keyboard to close after the new models come so that we can select another model based on our search

Rachit1313:
> I believe for mobile, we can have debounce itself of 600 as probably we would want the keyboard to close after the new models come so that we can select another model based on our search

@humphd / @kliu57 / @tarasglek

.vscode/settings.json
humphd, line 8: I hate how VSCode constantly changes this

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -9,33 +9,48 @@ import {
...
 import { usingOfficialOpenAI } from "../../lib/providers";
+import { useDebounce } from "react-use";
 
 type PromptSendButtonProps = {
   isLoading: boolean;
 };
 
 function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
+  const [searchQuery, setSearchQuery] = useState("");
+  const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("");
```
humphd, line 35: It seems odd to me to have this twice.  Can we do it with a single state value, and debound the `set*` method once?

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -141,11 +177,36 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
 }
 
 function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
+  const [searchQuery, setSearchQuery] = useState("");
+  const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("");
```
humphd, line 181: Same comment here

src/components/PromptForm/PromptSendButton.tsx
humphd: If this is too hard to make work now, let's not block on it.  File a follow-up to do a better fix later.

Rachit1313:
As discussed with dave, will work on trying to refactor the code to store searchQuery in 1 useState() function in a follow up later as trying to fix that now is causing another bugs.

src/components/PromptForm/PromptSendButton.tsx
Rachit1313: Done. #603

src/components/PromptForm/PromptSendButton.tsx
rjwignar, line 1: I noticed that, while typing in keywords to search, if there's content in the prompt entry and you press Enter, it sends the prompt to the LLM:
enterDuringSearch.

I'm not sure if this is the behaviour we want. I'd like to hear more opinions, but if it's something we don't want I think we iterate on this in a follow-up

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -141,11 +181,36 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
...
   }, [models]);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useDebounce(
+    () => {
+      setDebouncedSearchQuery(searchQuery);
+    },
+    250,
+    [searchQuery]
+  );
```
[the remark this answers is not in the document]
rjwignar: In a similar vein, but I think it's more noticeable:
If you're pressing "backspace" after entering a keyword, and the list re-renders to show more/less matches, the next time you press "backspace" the entire keyword will be cleared:
backspaceDuringSearch

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -104,14 +119,35 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
...
                   key={model.id}
                   onClick={() => setSettings({ ...settings, model })}
                 >
+                  {settings.model.id === model.id ? (
+                    <MdOutlineChevronRight style={{ marginRight: "4px" }} />
+                  ) : (
+                    <span style={{ width: "24px", display: "inline-block" }} />
+                  )}
                   {model.prettyModel}
                 </MenuItem>
```
rjwignar, line 152: I briefly tested the debounce on mobile (Safari) and noticed the model search feature is a little unwieldy, at least on an iPhone 8.
It's nothing that will block these changes. We can update the mobile version in followup issues if required.

rjwignar (changes requested):
@Rachit1313 The new changes look good.
I kept note of things I think can be improved, probably in followups.
If we can iterate on these in followups I'd love to approve this

src/components/PromptForm/PromptSendButton.tsx
Rachit1313: I think this is because its triggering the enter of chat message as there's no enter function on the search bar.

src/components/PromptForm/PromptSendButton.tsx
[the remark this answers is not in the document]
Rachit1313: @humphd , The above described unusual behavior is what I mentioned you in the chat that I faced after I made the changes to refactor debounce. The difference is that it is more frequent if we make the changes.

src/components/PromptForm/PromptSendButton.tsx
Rachit1313: Could you please detail a bit n this?

kliu57:
@rjwignar @Rachit1313 @humphd 
Hey Roy, I found out the root of the problem for your backspace issue. It is not anything to do with backspace.

The issue is still to do with hovering your mouse over one of the models in the models list, however now we do not lose focus, which is good. However we do lose the previous content we typed.

**Steps to reproduce the issue:**

1. Get ready to type a long word for example I will type openchat (open then chat)
2. Type open
3. Wait a second or two so the search results load
4. Hover your mouse over a model
5. Now type chat
6. Your search will be wiped and you will only see chat

It does not matter if you backspaced or you typed a new letter, the previous typed content will be wiped.

@humphd not sure if this is something we can fix? But it does look a lot better than before since we do not lose focus. However the user has to restart typing from scratch if they did not manage to finish typing everything before the useDebounce period. Perhaps we should increase the useDebounce from 250 to something more? If I don't finish typing in under 250 I will need to restart typing from scratch.

humphd (changes requested):
I played with this some more and I think that we need 2 fixes:

1. Don't erase everything in the search input when starting to type, in case there is already something there
2. When the menu closes, reset the search input

I tried implementing both and it seemed to work well:

```diff
diff --git a/src/components/PromptForm/PromptSendButton.tsx b/src/components/PromptForm/PromptSendButton.tsx
index ee458f5..ab700ec 100644
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -204,7 +204,7 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
       // Ignore control keys
       const char = e.key.length === 1 ? e.key : "";
       // Set the initial character in the input so we don't lose it
-      setSearchQuery(char);
+      setSearchQuery(searchQuery + char);
     }
   };

@@ -253,7 +253,12 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
           </Button>
         </Tooltip>
       )}
-      <Menu placement="top" strategy="fixed" closeOnSelect={false}>
+      <Menu
+        placement="top"
+        strategy="fixed"
+        closeOnSelect={false}
+        onClose={() => setSearchQuery("")}
+      >
         <MenuButton
           as={IconButton}
           size="sm"
```

tarasglek:
> 2. When the menu closes, reset the search input

I actually really like the behavior of keeping state in there. I put 22B in and was able to just have various mistral finetunes in the list to compare between them.

humphd:
Let's get this in before we ship 2.0 please.

kliu57 (approved):
LGTM! Not getting any of the previous issues anymore

src/components/PromptForm/PromptSendButton.tsx
rjwignar: Agreed. However, do we want to prevent this from happening if the search bar is open?

rjwignar (approved):

rjwignar:
It's perfect. These changes fix all the issues I previously had

humphd (approved):
@Rachit1313 land this when you're ready

PR #601, opened 2024-04-18
Add custom error toast message width on mobile

### LINKED ISSUE #569: Impossible to work with errors on mobile
I have an error on mobile about my API key, and it's not possible to close it.  We should fix the max-width on this so it is:

rjwignar:
Similar to #577, this fixes #569 by making toast error messages fit the viewport when mobile is detected.

However, instead of adding the container style within the **useAlert**, the container styling is applied to the error **useCallback**.

This appears to avoid the bugs caused by #577 and reported in #586, but I'd like more opinions.

If this does fix the issues previously encountered, do we also want to apply this change to the rest of the useCallbacks?

### DIFF
--- a/src/hooks/use-alert.ts
+++ b/src/hooks/use-alert.ts
@@ -54,7 +54,7 @@ export function useAlert() {
           width: isMobile ? "90vw" : "initial",
         },
       }),
-    [toast]
+    [toast, isMobile]
   );
 
   const success = useCallback(
--- a/src/hooks/use-alert.ts
+++ b/src/hooks/use-alert.ts
@@ -1,5 +1,6 @@
 import { useCallback } from "react";
 import { useToast } from "@chakra-ui/react";
+import useMobileBreakpoint from "./use-mobile-breakpoint";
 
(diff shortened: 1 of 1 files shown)


kliu57 (approved):
Looks good to be, not experiencing any of the issues from before.

kliu57:
@rjwignar I think this toast width fix you did here could be applied to the other toasts as well (info, warning, success). What do you think @humphd

kliu57:
@rjwignar @humphd Actually, I think for now we just need to apply this "toast width fix" to the "error type toast", since we are only getting the exceed screen width issue with very long messages. (And we don't have any long messages for success, info, or warning). Unless we want to apply it as a precaution for future cases.

WangGithub0:
> @WangGithub0 @Rachit1313 Could one of you test this toast width on Safari? The test is just that the toast width when you enter an invalid key for your current provider will not exceed the length of the page. (The red toast alert)

I tested on Safari, looks good.

WangGithub0 (approved):

PR #602, opened 2024-04-18
Allow downloading and re-announcing messages of infinite length

### LINKED ISSUE #525: TTS throws error when more than 4096 characters
I realized that Text to Speech only allows 4096 characters at max. @tarasglek also mentioned it before

image

Not sure what's the best way to handle this. 
Should we hide the TTS options when the message is too long or let the error show as its current state?

I think hiding the options can be confusing for users.

Amnish04:
A few weeks ago, I tried to download a podcast imported from youtube using web handlers, as **audio**.

However that didn't work out well because **OpenAI TTS API** only supports maxmum of 4096 characters input.

In this PR, I have attempted to fix that issue and now ChatCraft can **Speak** and **Download** messages of infinite length (potentially) as long as you have enough network and willing to wait.

1. I have written a `getSentenceChunksFrom` function that leverages [compromise](https://github.com/spencermountain/compromise) (the nlp library we are using), to try to break piece of text into chunks of sentences where each chunk is at max the number of characters specified in function call.
2. This function is not perfect, since I do not further break the sentences given to me by the nlp tokenize function. This means my function will fail to keep a chunk under specified character limit if any of the sentences itself exceeds that limit. However, this case is next to impossible since no sentence should be longer than 4096 characters.
3. The download option makes use of this function to split huge messages into sentence chunks, sends those chunks to tts api for processing, combines the returned blobs into a single blob using the default Blob constructor (possible because all blobs have same encoding), and downloads that single combined blob on user's file system.

With this fix, I was able to download an hour long interview and the generated audio size was **33 minutes**.
Generated Audio

This fixes #525

**Note:** I did not have to use [ffmpeg](https://github.com/ffmpegwasm/ffmpeg.wasm?tab=readme-ov-file) as originally discussed during triage.

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -262,18 +262,29 @@ function MessageBase({
         });
 
         try {
-          const textChunks = getSentenceChunksFrom(text, 4096);
-          const audioClips: Blob[] = [];
+          const textChunks = getSentenceChunksFrom(text, 500);
+          const audioClips: Blob[] = new Array<Blob>(textChunks.length);
 
-          for (const textChunk of textChunks) {
-            const audioClipUrl = await textToSpeech(
-              textChunk,
-              settings.textToSpeech.voice,
-              "tts-1-hd"
-            );
+          // Limit the number of concurrent tasks
+          const pLimit = (await import("p-limit")).default;
 
-            audioClips.push(await fetch(audioClipUrl).then((r) => r.blob()));
-          }
+          const limit = pLimit(8); // Adjust the concurrency limit as needed
+
+          const tasks = textChunks.map((textChunk, index) => {
+            return limit(async () => {
+              const audioClipUrl = await textToSpeech(
+                textChunk,
+                settings.textToSpeech.voice,
+                "tts-1-hd"
+              );
+
+              const audioClip = await fetch(audioClipUrl).then((r) => r.blob());
+              audioClips[index] = audioClip;
+            });
+          });
+
+          // Wait for all the tasks to complete
+          await Promise.all(tasks);
 
           const audioClip = new Blob(audioClips, { type: audioClips[0].type });
 
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -53,7 +53,14 @@ import {
   ChatCraftSystemMessage,
 } from "../../lib/ChatCraftMessage";
 import { ChatCraftModel } from "../../lib/ChatCraftModel";
-import { download, formatDate, formatNumber, getMetaKey, screenshotElement } from "../../lib/utils";
+import {
+  download,
+  formatDate,
+  formatNumber,
+  getMetaKey,
+  screenshotElement,
+  utilizeAlert,
+} from "../../lib/utils";
 import ImageModal from "../ImageModal";
 import Markdown from "../Markdown";
 
@@ -65,8 +72,8 @@ import { useUser } from "../../hooks/use-user";
 import { ChatCraftChat } from "../../lib/ChatCraftChat";
 import { textToSpeech } from "../../lib/ai";
 import { usingOfficialOpenAI } from "../../lib/providers";
-import "./Message.css";
 import { getSentenceChunksFrom } from "../../lib/summarize";
+import "./Message.css";
 
 export interface MessageBaseProps {
   message: ChatCraftMessage;
@@ -247,14 +254,28 @@ function MessageBase({
     if (messageContent.current) {
       const text = messageContent.current.textContent;
       if (text) {
-        try {
-          info({
-            title: "Downloading...",
-            message: "Please wait while we prepare your audio download.",
-          });
+        const { loading, closeLoading } = await utilizeAlert();
 
-          const audioClipUrl = await textToSpeech(text, settings.textToSpeech.voice, "tts-1-hd");
-          const audioClip = await fetch(audioClipUrl).then((r) => r.blob());
+        const alertId = loading({
+          title: "Downloading...",
+          message: "Please wait while we prepare your audio download.",
+        });
+
+        try {
+          const textChunks = getSentenceChunksFrom(text, 4096);
+          const audioClips: Blob[] = [];
+
+          for (const textChunk of textChunks) {
+            const audioClipUrl = await textToSpeech(
+              textChunk,
+              settings.textToSpeech.voice,
+              "tts-1-hd"
+            );
+
+            audioClips.push(await fetch(audioClipUrl).then((r) => r.blob()));
+          }
+
+          const audioClip = new Blob(audioClips, { type: audioClips[0].type });
 
           download(
             audioClip,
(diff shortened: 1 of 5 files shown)


src/components/Message/MessageBase.tsx
```diff
@@ -246,27 +254,44 @@ function MessageBase({
...
+
         try {
-          info({
-            title: "Downloading...",
-            message: "Please wait while we prepare your audio download.",
-          });
+          const textChunks = getSentenceChunksFrom(text, 4096);
+          const audioClips: Blob[] = [];
+
+          for (const textChunk of textChunks) {
```
humphd, line 268: We could run these in parallel I bet.

https://chatcraft.org/api/share/humphd/dgaqPMXbuhzMvzyNJNELQ

src/lib/summarize.ts
```diff
@@ -56,6 +56,44 @@ export function tokenize(text: string) {
 
   return { sentences, terms };
 }
+/**
+ *
+ * Tries to split the provided text into
+ * an array of text chunks where
+ * each chunk is composed of one or more sentences.
+ * The function attempts to limit each chunk to maximim
```
humphd, line 64: maximum

src/lib/summarize.ts
```diff
@@ -56,6 +56,44 @@ export function tokenize(text: string) {
...
+ * preferred characters preferred, but the chunk limit may still exceed
```
humphd, line 65: preferred repeated

src/lib/summarize.ts
```diff
@@ -56,6 +56,44 @@ export function tokenize(text: string) {
...
+ */
+export function getSentenceChunksFrom(text: string, maxCharsPerSentence: number = 4096): string[] {
+  const { sentences } = tokenize(text);
+  const chunks: string[] = [];
+
+  let currentText = "";
+
+  for (const sentence of sentences) {
+    if (currentText.length + sentence.length < maxCharsPerSentence) {
+      currentText += ` ${sentence}`;
```
humphd, line 80: `sentence.trim()` so you don't have a leading space.

humphd (commented):
Very cool.  Bonus marks for not needing ffmpeg :)

I left some feedback.

kliu57 (commented):
Looks good to me.

But already discussed I did find it weird that in production when the text limit is exceeded we are not triggering the error toast when we click "Speak". If it's too much work you don't have to fix it in this PR. But yeah now that you got rid of the limit it will be hard for me to test triggered that toast since we no longer have a limit.

And are we sure we don't want a limit for the download? Because although I managed to download the 30,000 character as audio file it took over 5 minutes and close to 10, I doubt anyone would want to wait that long to download it without seeing a progress bar. Should we still have some sort of upper cap?

src/lib/summarize.ts
Amnish04: Fixed all the typos in latest commit

src/components/Message/MessageBase.tsx
Amnish04: @humphd I've implemented this suggestion in the latest commit and the download speed by has improved **10 times**.
This is a great thing to learn on my last day in CPA 😄

Amnish04:
> Looks good to me.
> 

@kliu57 Thanks for reviewing!

> But already discussed I did find it weird that in production when the text limit is exceeded we are not triggering the error toast when we click "Speak". If it's too much work you don't have to fix it in this PR. But yeah now that you got rid of the limit it will be hard for me to test triggered that toast since we no longer have a limit.

I have fixed that issue and the toast message should technically show up for Speak as well. Although the 4096 error can no longer be reproduced since we are sending smaller chunks now.

> And are we sure we don't want a character limit? Because although I managed to download the 30,000 character as audio file it took over 5 minutes and close to 10, I doubt anyone would want to wait that long to download it without seeing a progress bar. Should we still have some sort of upper cap?

After my latest commit, the download speed went from 5 minutes to just 50 seconds for the same message.

I agree that there should be a progress bar, so the user knows something is happening and the app has not glitched. We can probably have a follow up issue for it.

tarasglek:
Coolness, so we can generate audiobooks now?

humphd:
Glad the `p-limit` stuff worked, nice work.  Parallel execution is often hard to do well, but this problem is perfect for it.

Progress bar would be fairly easy to do in a follow-up, since the data is already chunked and we can report progress updates when each promise resolves.

Love this @Amnish04, let's get it in for 2.0!

kliu57 (approved):

mingming-ma (commented):
@Amnish04 It's amazing to download hours audio within 1 minute!

I believe it should work in most situations, I just tested on this branch and encountered 4096 error (`/import https://www.youtube.com/watch?v=QfriT4LirxA`), I'll try to find out which reason, meanwhile can you also check this and let me know where I was failed. 
image


Update:

I think I got it. That video is some kind of long sentences that can't be splited by [compromise](https://github.com/spencermountain/compromise) in this case, I think maybe we can just use the str length to split the chunk?

Amnish04:
> Coolness, so we can generate audiobooks now?

@tarasglek Yes, as long as we can import the transcriptions in ChatCraft :)

Amnish04:
> Glad the `p-limit` stuff worked, nice work. Parallel execution is often hard to do well, but this problem is perfect for it.
> 
> Progress bar would be fairly easy to do in a follow-up, since the data is already chunked and we can report progress updates when each promise resolves.
> 
> Love this @Amnish04, let's get it in for 2.0!

Thanks for your help :)

I've opened a follow-up for progress bar #604

Amnish04:
> @Amnish04 It's amazing to download hours audio within 1 minute!
> 
> I believe it should work in most situations, I just tested on this branch and encountered 4096 error (`/import https://www.youtube.com/watch?v=QfriT4LirxA`), I'll try to find out which reason, meanwhile can you also check this and let me know where I was failed. image

@mingming-ma I am not sure why that happened. Can you provide me more context of the request that failed (payload, headers and response) and the message you used for this?

Amnish04:
> @Amnish04 Yeah, here is the [chat](https://5a8cb1a1.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/TBYMGSZ7gApZ7Id0WiH9e)

Ohh right. Since the text has no punctuation, the entire thing in treated as a single sentence. I guess I will try to further break a sentence that exceeds our limit.

mingming-ma:
@Amnish04 Yeah, it's so strange the nlp tool can't do this 😢

mingming-ma:
@Amnish04 Another thought, maybe we can call gpt-3.5 api in follow up PR like  "add punctuations" to this kind of string, and then use similar workflow.
image

Amnish04:
> @Amnish04 Another thought, maybe we can call gpt-3.5 api in follow up PR like "add punctuations" to this kind of string, and then use similar workflow. image

@mingming-ma That could be a potential solution, but I feel like it's overly complicated.

I just pushed a fix where I try to force break a sentence longer than our character limit. But I've made sure that even those force broken chunks end with a full word

humphd (approved):

kliu57 (approved):
Tested with a long message without punctuation and the error mingming was getting is not longer there. LGTM!

mingming-ma (approved):
@Amnish04 I like your regex approach!

Amnish04:
> @Amnish04 I like your regex approach!

@mingming-ma That is ChatCraft's approach 😄

Issue #607, opened 2024-04-21
import openai messages json


tarasglek:
openai completion endpoints has format 
[
{role: "system", content:"systemprompt"}
{role:"user"....
]

Would be great if we could import such json. This would make it easy to debug issues in ai apps with chatcraft.

Similarly, it would be nice to export in this format too

tarasglek:
So I thought about it some more...

You gonna hate this idea probably... this is getting close to my libchatcraft idea.

Why not offer to render chat view as a yaml/json textbox. Call it developer-view. Then we can get rid of copy/export UI and also not bother with import UI.

This would let you have a chat with llm in developer style, where you see the messages array and can edit it..then every time user presses ask..we reparse the yaml into our internal format and submit it into llm..we do something like this already anyway, its a "small" ui format change

humphd:
This is an interesting idea.  We should work out the details, but I think offering a dev mode that is raw data vs. rendered is a neat way to solve the ingestion problem.

PR #608, opened 2024-04-21
Allow exporting/copying to OpenAI format JSON, YAML, fix menus, remove chat header menu

humphd:
This is the export part of #607.  I'm not sure what the right UX is for import...

The PR adds support for exporting (download or copy) a chat as OpenAI formatted JSON and YAML:

Screenshot 2024-04-21 at 1 17 59 PM

I've removed the Chat Header menu, since we do everything and more via the Options button/menu now.

I've fixed our custom Menu components to work better with Chakra colors, zIndex, naming for `isDisabled` prop, passing your own menu button, etc. and reworked the other users of it.

### DIFF
--- a/src/Chat/ChatHeader.tsx
+++ b/src/Chat/ChatHeader.tsx
@@ -89,57 +89,55 @@ function ChatHeader({ chat }: ChatHeaderProps) {
         mt={2}
       >
         <CardBody pb={0}>
-          <Flex justify="space-between" align="center">
-            <Box w="100%">
-              {isEditing ? (
-                <form onSubmit={handleSaveSummary}>
-                  <Flex align="center" gap={2}>
-                    <Input
-                      flex={1}
-                      defaultValue={chat.summary}
-                      type="text"
-                      name="summary"
-                      bg="white"
-                      _dark={{ bg: "gray.700" }}
-                      size="sm"
-                      w="100%"
-                      autoFocus={true}
-                      placeholder="Chat Summary"
-                    />
-                    <ButtonGroup>
-                      <Button variant="outline" size="xs" onClick={() => setIsEditing(false)}>
-                        Cancel
-                      </Button>
-                      <Button size="xs" type="submit">
-                        Save
-                      </Button>
-                    </ButtonGroup>
-                  </Flex>
-                </form>
-              ) : (
-                <Flex align="center" gap={2} maxW="100%">
-                  <Box>
-                    <MdOutlineChatBubbleOutline />
-                  </Box>
-                  <Text fontSize="md" fontWeight="bold" noOfLines={1} title={title}>
-                    <Link as={ReactRouterLink} to={`/c/${chat.id}`}>
-                      {title}
-                    </Link>
-                  </Text>
-                  {!chat.readonly && settings.currentProvider.apiKey && (
-                    <IconButton
-                      variant="ghost"
-                      size="sm"
-                      icon={<AiOutlineEdit />}
-                      aria-label="Edit summary"
-                      title="Edit summary"
-                      onClick={() => setIsEditing(true)}
-                    />
-                  )}
+          <Box w="100%">
+            {isEditing ? (
+              <form onSubmit={handleSaveSummary}>
+                <Flex align="center" gap={2}>
+                  <Input
+                    flex={1}
+                    defaultValue={chat.summary}
+                    type="text"
+                    name="summary"
+                    bg="white"
+                    _dark={{ bg: "gray.700" }}
+                    size="sm"
+                    w="100%"
+                    autoFocus={true}
+                    placeholder="Chat Summary"
+                  />
+                  <ButtonGroup>
+                    <Button variant="outline" size="xs" onClick={() => setIsEditing(false)}>
+                      Cancel
+                    </Button>
+                    <Button size="xs" type="submit">
+                      Save
+                    </Button>
+                  </ButtonGroup>
                 </Flex>
-              )}
-            </Box>
-          </Flex>
+              </form>
+            ) : (
+              <Flex align="center" gap={2} maxW="100%">
+                <Box>
+                  <MdOutlineChatBubbleOutline />
+                </Box>
+                <Text fontSize="md" fontWeight="bold" noOfLines={1} title={title}>
+                  <Link as={ReactRouterLink} to={`/c/${chat.id}`}>
+                    {title}
+                  </Link>
+                </Text>
+                {!chat.readonly && settings.currentProvider.apiKey && (
+                  <IconButton
+                    variant="ghost"
+                    size="sm"
+                    icon={<AiOutlineEdit />}
+                    aria-label="Edit summary"
+                    title="Edit summary"
+                    onClick={() => setIsEditing(true)}
+                  />
+                )}
+              </Flex>
+            )}
+          </Box>
         </CardBody>
         <CardFooter pt={0}>
           <Flex w="100%" gap={4} color="gray.500" _dark={{ color: "gray.400" }}>
--- a/src/Chat/ChatHeader.tsx
+++ b/src/Chat/ChatHeader.tsx
@@ -10,43 +10,35 @@ import {
   IconButton,
   Input,
   Link,
-  Menu,
-  MenuButton,
-  MenuDivider,
-  MenuItem,
-  MenuList,
   Tag,
   Text,
   useDisclosure,
 } from "@chakra-ui/react";
-import { Link as ReactRouterLink, useFetcher } from "react-router-dom";
+import { Link as ReactRouterLink } from "react-router-dom";
 import { MdOutlineChatBubbleOutline } from "react-icons/md";
-import { TbDots } from "react-icons/tb";
 import { AiOutlineEdit } from "react-icons/ai";
-import { useCopyToClipboard, useKey } from "react-use";
+import { useKey } from "react-use";
 
 import { ChatCraftChat } from "../lib/ChatCraftChat";
-import { download, formatCurrency, formatDate, formatNumber } from "../lib/utils";
+import { formatCurrency, formatDate, formatNumber } from "../lib/utils";
 import ShareModal from "../components/ShareModal";
 import { useSettings } from "../hooks/use-settings";
 import useTitle from "../hooks/use-title";
-import { useAlert } from "../hooks/use-alert";
 import { useCost } from "../hooks/use-cost";
+import { useAlert } from "../hooks/use-alert";
 
 type ChatHeaderProps = {
   chat: ChatCraftChat;
 };
 
 function ChatHeader({ chat }: ChatHeaderProps) {
-  const [, copyToClipboard] = useCopyToClipboard();
-  const { info, error } = useAlert();
-  const fetcher = useFetcher();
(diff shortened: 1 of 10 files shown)


src/Chat/ChatHeader.tsx
```diff
@@ -167,31 +139,6 @@ function ChatHeader({ chat }: ChatHeaderProps) {
...
-
-                    <MenuDivider />
-                    <MenuItem color="red.400" onClick={() => handleDeleteClick()}>
-                      Delete
-                    </MenuItem>
-                  </>
-                )}
-              </MenuList>
-            </Menu>
           </Flex>
```
Amnish04, line 142: We can probably remove this `Flex` now that we removed the menu.

src/components/Menu/Menu.tsx
```diff
@@ -13,23 +14,21 @@ import "./Menu.css";
 
 export type MenuProps = Omit<ReactMenuProps, "menuButton" | "theming" | "transition"> & {
   isDisabled?: boolean;
+  menuButton?: ReactMenuProps["menuButton"];
 };
```
Amnish04, line 18:
```suggestion
export type MenuProps = Omit<ReactMenuProps, "theming" | "transition"> & {
  isDisabled?: boolean;
};
```

Maybe I am missing something, but can we just remove `menuButton` from omitted props instead of adding it again?

src/components/Menu/Menu.tsx
```diff
@@ -13,23 +14,21 @@ import "./Menu.css";
...
   const { colorMode } = useColorMode();
+  const menuButton = props.menuButton ?? (
+    <IconButton aria-label="Menu" icon={<TbDots />} variant="ghost" isDisabled={props.isDisabled} />
+  );
+
   return (
-    <Box>
+    <Box zIndex={theme.zIndices.dropdown}>
       <ReactMenu
         align={"end"}
```
Amnish04, line 29:
```suggestion
        {...props}
        align={"end"}
```

Should we also pass down the rest of the props?

Also for some reason, `align="end"` has stopped working
image


In Prod:
image

src/components/Message/MessageBase.tsx
```diff
@@ -540,31 +540,27 @@ function MessageBase({
...
-                        className="delete-button"
-                        icon={<TbTrash color="red.400" />}
-                      >
+                      <MenuItem onClick={onDeleteClick} color="red.400" icon={<TbTrash />}>
                         Delete Message
                       </MenuItem>
                     ) : (
-                      <SubMenu label="Delete" className="delete-button">
+                      <SubMenu label="Delete">
                         {onDeleteBeforeClick && (
```
Amnish04, line 548: Is it intentional to not have delete submenu item as red?

src/components/OptionsButton.tsx
```diff
@@ -142,108 +133,181 @@ function OptionsButton({
...
 
-  const handleDownloadClick = useCallback(() => {
+  const handleCopyAsJson = useCallback(() => {
     if (!chat) {
       return;
     }
 
-    const text = chat.toMarkdown();
-    download(text, "chat.md", "text/markdown");
+    const model = settings.model.name;
```
Amnish04, line 161: I am not sure what's the actual use case for this export, but do we not care about the model that was used to generate these exported messages?

Amnish04 (commented):
@humphd The JSON and YAML exports work well!

I have left a few comments after reading the code.

src/components/OptionsButton.tsx
tarasglek: would be cool to generate with model

tarasglek:
Having both download and copy further multiplied by all the download formats ia bit weird, but I can't decide which I like better :)

src/components/OptionsButton.tsx
tarasglek: from testing it exports model correctly

tarasglek:
lgtm

src/components/Menu/Menu.tsx
humphd: What is `align="end"` supposed to be doing here?  Is it what is breaking the width?  I'm not sure what the right fix is for that, could use some help.

humphd:
> I noticed Options/Clear does not work anymore


really?  It's working for me. So nothing happens?  Any console log?

humphd:
Pushed some fixes.

I'm not sure that I like "Download" since it implies that we've hosted this "on the web" vs. "in your browser."  I wonder if "Export" would be better, and also pave the way for "Import"?

Another thing I think is wrong is using the copy/download icons each time for all menu items.

@Amnish04 I'm not sure how to fix the width, position of the menu.  Any ideas?

src/components/OptionsButton.tsx
Amnish04: @humphd Makes sense, thanks for clarifying.

src/components/Message/MessageBase.tsx
Amnish04: If we do not want to use `className`, then this approach is also working.

```tsx
// SubMenu.tsx

const SubMenu: React.FC<SubMenuProps> = (props) => {
  const theme = useTheme();

  const toCssColor = (colorValue?: string) => {
    if (!colorValue) {
      return undefined;
    }

    // Check for ChakraUI colors vs normal CSS (e.g., "red.400")
    if (!/\w+\.\d+/.test(colorValue)) {
      return colorValue;
    }

    const [color, shade] = colorValue.split(".");
    return theme.colors[color]?.[shade];
  };
  return (
    <ReactMenuSubMenu
      itemProps={{ style: { color: toCssColor(props.color) } }}
      align="center"
      {...props}
    >

// MessageBase.tsx

<SubMenu label="Delete" color="red.400">
    {onDeleteBeforeClick && (
```

src/components/Menu/Menu.tsx
Amnish04: `align` is used to position the menu/submenu contents with respect to the **anchor** element.
image

In this case, it should be "start" so the contents do not touch the edge. (I was mistaken that it should be "end")
Also, instead of overriding the default `align` value here (which is "start"), we can change it wherever we need a different value.
For example, in `MessageBase` we need it to be `end` so we do not have a horizontal scrollbar.
```tsx
<Menu align="end" isDisabled={isLoading}>
  <MenuItem onClick={handleCopy} icon={<MdContentCopy />}>
    Copy
  </MenuItem>
```



For the **width**, I did not find any official property in the docs. But we can use `menuStyle` to set a `minWidth`.
```tsx
// Props that we want to force-override can go after `{...props}`
return (
  <Box zIndex={theme.zIndices.dropdown}>
    <ReactMenu
      transition={true}
      theming={colorMode === "dark" ? "dark" : undefined}
      {...props}
      menuButton={menuButton}
      menuStyle={{
        minWidth: "225px",
        ...props.menuStyle,
      }}
    >
      {props.children}
    </ReactMenu>
  </Box>
);
```

**Final Result:**
image

However, even though all of this works, I don't like how we have to micro-manage the positioning of the menu in different scenarios. Not sure, if its the component's design or we are using it wrong.

Amnish04:
> I'm not sure that I like "Download" since it implies that we've hosted this "on the web" vs. "in your browser." I wonder if "Export" would be better, and also pave the way for "Import"?

`Export` sounds good to me.

> Another thing I think is wrong is using the copy/download icons each time for all menu items.

I tried to only have the icon only on the Submenu item.

**SubMenu.tsx**
```tsx
import { Flex, Icon, useTheme } from "@chakra-ui/react";
import { SubMenu as ReactMenuSubMenu, type SubMenuProps } from "@szhsin/react-menu";
import React from "react";
import { IconType } from "react-icons";

type ReactSubMenuProps = Omit<SubMenuProps, "label"> & {
  label: string;
  icon?: IconType;
  iconSpacing?: number;
};

const SubMenu: React.FC<ReactSubMenuProps> = ({ color, label, icon, iconSpacing, ...props }) => {
  const theme = useTheme();

  const toCssColor = (colorValue?: string) => {
    if (!colorValue) {
      return undefined;
    }

    // Check for ChakraUI colors vs normal CSS (e.g., "red.400")
    if (!/\w+\.\d+/.test(colorValue)) {
      return colorValue;
    }

    const [color, shade] = colorValue.split(".");
    return theme.colors[color]?.[shade];
  };
  return (
    <ReactMenuSubMenu
      itemProps={{ style: { color: toCssColor(color) } }}
      align="center"
      {...props}
      label={
        <Flex gap={iconSpacing ?? 2} alignItems={"center"}>
          {icon && <Icon as={icon}></Icon>}
          {label}
        </Flex>
      }
    >
      {props.children}
    </ReactMenuSubMenu>
  );
};

export default SubMenu;
```

**OptionsButton.tsx**
```tsx
<SubMenu label="Copy" icon={TbCopy}>
  <MenuItem isDisabled={!chat} onClick={() => handleCopyAsMarkdown()}>
    Copy as Markdown
  </MenuItem>
  <MenuItem isDisabled={!chat} onClick={() => handleCopyAsJson()}>
    Copy as JSON
  </MenuItem>
  <MenuItem isDisabled={!chat} onClick={() => handleCopyAsYaml()}>
    Copy as YAML
  </MenuItem>
</SubMenu>
<SubMenu label="Download" icon={TbDownload}>
  <MenuItem isDisabled={!chat} onClick={handleDownloadMarkdown}>
    Download as Markdown
  </MenuItem>
  <MenuItem isDisabled={!chat} onClick={handleDownloadJson}>
    Download as JSON
  </MenuItem>
  <MenuItem isDisabled={!chat} onClick={handleDownloadYaml}>
    Download as YAML
  </MenuItem>
</SubMenu>
```

And this is how it looks
image

> @Amnish04 I'm not sure how to fix the width, position of the menu. Any ideas?

I have posted my research regarding this in our original conversation.

humphd:
@Amnish04 can you take another look?  I've made changes based on your review.

I think everything is good now except the broken "Clear" menu item that @tarasglek noticed (I have confirmed that it breaks in the preview URL, but works locally).

```tsx
// before
        <MenuItem as={ReactRouterLink} to="/new" target="_blank">
          Clear
        </MenuItem>

// after
      <MenuItem>
        <Link as={ReactRouterLink} to="/new">
          Clear
        </Link>
      </MenuItem>
```

Not sure what I'm doing wrong here, or if the preview URLs are handling things weird?  Seems like prod will fail with this, though.

Amnish04:
> @Amnish04 can you take another look? I've made changes based on your review.

@humphd Looks good for the most part! Just wondering should we also rename the toast messages to use "Export" instead of "Download".
image

> Not sure what I'm doing wrong here, or if the preview URLs are handling things weird? Seems like prod will fail with this, though.

For the `Clear` option, I guess the problem is that `Link` element does not take full width and height in a menu item.

We could probably wrap the entire `MenuItem` in a `Link`

```tsx
<Link as={ReactRouterLink} to="/new" _hover={{ textDecoration: "none" }}>
  <MenuItem>Clear</MenuItem>
</Link>
```

humphd:
@Amnish04 take a look at these changes.  I created a new `MenuItemLink` component to deal with the special cases for handling links.

I also fixed the download vs. exported titles/labels.

Amnish04 (approved):
@humphd Looks good to me. Thanks for refactoring the Menu!

PR #616, opened 2024-04-25
Clear regular prompt message if using Re-Ask Option

### LINKED ISSUE #587: Re-asking Human message duplicates the message
I noticed when editing and re-asking a message, ChatCraft duplicates my updated message:
messageDuplication

rjwignar:
This fixes #587 

Background
---

7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d allowed web handlers and slash commands to be executed with the Re-ask option (issue described in #541) by passing the prompt text to the **onPrompt** call in **MessagesView::onResubmitClick()**.

Before this change, retrying with a regular prompt message would trigger the regular re-try routine (since the prompt text **wasn't passed** to the **onPrompt** call in **MessagesView::onResubmitClick()**:
https://github.com/tarasglek/chatcraft.org/blob/9f73deac651f7caa6144e905e326d72a543db002/src/Chat/ChatBase.tsx#L283-L297

**After this change however,** if re-asking with a regular prompt message, the updated message will be passed to **onPrompt** (as **prompt**) and add a **duplicate message** to the chat (line 278):
https://github.com/tarasglek/chatcraft.org/blob/9f73deac651f7caa6144e905e326d72a543db002/src/Chat/ChatBase.tsx#L275-L278

Changes
---

I added a **boolean** parameter to **ChatBase::onPrompt** called `retry` (**false** by default):
```tsx
    async (prompt?: string, imageUrls?: string[], retry: boolean = false) => {
      // ...
      try {
        // If retrying, set the prompt to null to avoid duplicate message
        if (retry) {
          prompt = "";
        }
        // If the prompt text exist, package it up as a human message and add to the chat
        // ...
```
If `retry` is True, the prompt will be cleared to prevent message duplication.

`retry` is False **by default** because I only want to explicitly set it to True when calling **onPrompt()** from **onResubmitClick()**:
```tsx
type MessagesViewProps = {
  // ...
  onPrompt: (prompt?: string, imageUrls?: string[], retry?: boolean) => void;
};
// ... 

function MessagesView({
          onResubmitClick={async (promptText?: string) => {
            await deleteMessages(message.id, "after");
            onPrompt(promptText, undefined, true); // pass prompt text and true for retry, don't include any imageURLs
          }}
          // ...
        })
```

I had to update the onPrompt signature in MessagesViewProps for syntax reasons. If there's a better way to implement this I'm open to suggestions.

Comparison to Production
---

Below are the observations I've made on the Re-Ask options **before** and **after** 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d, as well as for this Pull Request:

| Task | Observations pre- 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d (tested using adcfd2d44c63447e6c9c5cfbd42c91e15116f990) | Observations post- 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d (tested using Production as of e4967a99b4cb7a6478ad5d10c738c3763ffdb126) (Prompt duplication bug observed) | PR Observations |
| --- | --- | --- | --- |
| Ask with "help" (to invoke /help) | Executes /help command  | Executes /help command | Executes /help command | 
| Re-ask with "help" (to invoke /help)     | Response from LLM **(undesired)** | Executes /help command  | Executes /help command | 
| Ask with web handler | Executes web handler | Executes web handler  | Executes web handler  | 
| Re-ask with web handler | Response from LLM **(undesired)** | Executes web handler  | Executes web handler  | 
| Ask with slash command (/import) | Invokes /import | Invokes /import | Invokes /import |
| Re-ask with slash command (/import) | Response from LLM **(undesired)** | Invokes /import | Invokes /import |
| Ask with regular prompt | Response from LLM | Response from LLM | Response from LLM |
| Re-ask with regular prompt | Response from LLM  | Response from LLM, **duplicates new message (undesired)** | Response from LLM, **no message duplication**|

My PR preserves the changes from 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d but prevent the message duplication when re-trying with a regular prompt message.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -202,7 +202,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   // Handle prompt form submission
   const onPrompt = useCallback(
-    async (prompt?: string, imageUrls?: string[]) => {
+    async (prompt?: string, imageUrls?: string[], retry: boolean = false) => {
       setLoading(true);
 
       // Special-case for "help", to invoke /help command
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -271,6 +271,10 @@ function ChatBase({ chat }: ChatBaseProps) {
       // Not a slash command, so pass this prompt to LLM
       let promptMessage: ChatCraftHumanMessage | undefined;
       try {
+        // If retrying, set the prompt to null to avoid duplicate message
+        if (retry) {
+          prompt = "";
+        }
         // If the prompt text exist, package it up as a human message and add to the chat
         if (prompt) {
           // Add this prompt message to the chat
--- a/src/components/MessagesView.tsx
+++ b/src/components/MessagesView.tsx
@@ -22,7 +22,7 @@ type MessagesViewProps = {
   isPaused: boolean;
   onTogglePause: () => void;
   onCancel: () => void;
-  onPrompt: (prompt?: string) => void;
+  onPrompt: (prompt?: string, imageUrls?: string[], retry?: boolean) => void;
 };
 
 function MessagesView({
@@ -99,7 +99,7 @@ function MessagesView({
           isLoading={isLoading}
           onResubmitClick={async (promptText?: string) => {
             await deleteMessages(message.id, "after");
-            onPrompt(promptText);
+            onPrompt(promptText, undefined, true); // pass prompt text and true for retry, don't include any imageURLs
           }}
           onDeleteBeforeClick={
             hasMessagesBefore ? () => deleteMessages(message.id, "before") : undefined


rjwignar:
Moving to draft because I haven't yet checked how re-asking with an Image has been affected (if at all).
Will mark this PR as ready in the morning once I've checked.

tarasglek:
seems to work well!

rjwignar:
I've added my Observations for Re-asking with **Image as Input <sup>**1**</sup>** before and after the message duplication bug.
| Task | Observations pre- 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d (tested using adcfd2d44c63447e6c9c5cfbd42c91e15116f990) | Observations post- 7cc1e2f3ec1aaf77014a56703d7d8e77c706f72d (tested using Production as of e4967a99b4cb7a6478ad5d10c738c3763ffdb126) (Prompt duplication bug observed) | PR Observations |
| --- | --- | --- | --- |
| Ask with Image as Input (no prompt) | Appropriate LLM response | Appropriate LLM response | Appropriate LLM response |
| Re-ask with Image as Input (no prompt, same image <sup>**2**</sup>)| Appropriate LLM Response | Appropriate LLM Response (**no message duplication as no prompt present, expected**) | Appropriate LLM (**no message duplication as no prompt present, expected**) |
| Ask with prompt and Image as Input | Appropriate LLM Response | Appropriate LLM Response | Appropriate LLM Response |
| Re-ask with prompt and Image as Input (same image <sup>**2**</sup>) | Appropriate LLM Response (**no message duplication, desired result**) | Appropriate LLM Response (**message duplication with text only, undesired <sup>**3**</sup>**)| Appropriate LLM Response (**no message duplication, desired result**) |

`1` Image as Input tested using `gpt-4-1106-vision-preview`
`2` Message editing allows text editing only, not image
`3` Duplication example provided below:
image

Future Considerations
---

As of e4967a99b4cb7a6478ad5d10c738c3763ffdb126, we currently don't pass any **imageUrls** (from **Image as Input**) in the **onPrompt()** call in **onResubmitClick()** during the Re-Ask Option.

However, if this changes in the future, we might see this bug again, in the form of **Input Image(s)** duplication.
**If** this bug appears in the future, we _might_ be able to fix it by extending the changes in this PR to clear **imageUrls** on retry, with something like:
```tsx
        // If retrying, empty the prompt and imageUrls to avoid duplicate message
        if (retry) {
          prompt = "";
          imageUrls = [];
        }
```

tarasglek:
@humphd this looks good to me, can you take a look?

humphd (commented):
I think this is fine, but I'd change the `onPrompt` signature.

src/components/MessagesView.tsx
```diff
@@ -99,7 +99,7 @@ function MessagesView({
           isLoading={isLoading}
           onResubmitClick={async (promptText?: string) => {
             await deleteMessages(message.id, "after");
-            onPrompt(promptText);
+            onPrompt(promptText, undefined, true); // pass prompt text and true for retry, don't include any imageURLs
```
humphd, line 102: A better pattern to use for this is to convert the args for `onPrompt` to be an `Object`:

```ts
async ({ prompt, imageUrls, retry = false }: { prompt?: string; imageUrls?; string[], retry?; boolean } = {}) => {
```

Now you can call it like this:

```ts
onPrompt({ prompt: promptText, retry: true });
```

This deals with undefined/optional args much better, and makes it self-documenting.

rjwignar: @humphd 
I updated the **onPrompt** signature in ChatBase and MessagesView, but to get this working I also had to make the same updates to the **onPrompt** and **onSendClick** (and any calls to these) signatures in various other components.

These changes still fix the duplication issues observed and preserve the other behaviours I've tested.
The **onPrompt**/**onSendClick** signature updates don't seem like they'll break anything (as long as any associated calls have also been updated accordingly), but I can't confirm anything at this time.

Should we put my recent commits (everything from bbdc1ab3d9950e5f38f3f43448b416d779579b0c onward) on the backburner for a follow-up PR?

src/Chat/ChatBase.tsx
```diff
@@ -202,7 +202,15 @@ function ChatBase({ chat }: ChatBaseProps) {
...
-    async (prompt?: string, imageUrls?: string[]) => {
+    async ({
+      prompt,
+      imageUrls,
+      retry = false,
+    }: {
+      prompt?: string;
+      imageUrls?: string[];
+      retry?: boolean;
+    }) => {
```
humphd, line 213: A trick you can use when all properties on the object are optional is to set the options object to have a default value:

```ts
    async ({
      prompt,
      imageUrls,
      retry = false,
    }: {
      prompt?: string;
      imageUrls?: string[];
      retry?: boolean;
    } = {}) => {
```

The value of this is that it lets callers do `onPrompt()` and it will work.

src/Chat/ChatBase.tsx
```diff
@@ -346,7 +358,7 @@ function ChatBase({ chat }: ChatBaseProps) {
 
           // If the user has opted to always send function results back to LLM, do it now
           if (settings.alwaysSendFunctionResult) {
-            await onPrompt();
+            await onPrompt({});
```
humphd, line 361: If you make my fix above, you can avoid this.

src/components/Message/FunctionResultMessage.tsx
```diff
@@ -39,7 +39,7 @@ function FunctionResultMessage(props: FunctionMessageProps) {
           Always send
         </Checkbox>
         {!settings.alwaysSendFunctionResult && (
-          <Button size="xs" onClick={() => onPrompt()}>
+          <Button size="xs" onClick={() => onPrompt({})}>
```
humphd, line 42: and here...

humphd (changes requested):
You could also define a reusable type for these options, since you have to use them in so many places.

src/Chat/ChatBase.tsx
rjwignar: Something like this?
```
async ({ 
    prompt = "", 
    imageUrls = [], 
    retry = false,
//...
```

humphd: No, see my code above.  I'm setting the value to `{}` by default.

rjwignar: Oh! I see it now. My bad

rjwignar: I followed your advice and added a reusable type called OnPromptFunction.
I tried the original trick you suggested but got this error when I applied it to the signature on other files on some files:
image

So I ended up using a different syntax [suggested by ChatCraft](https://chatcraft.org/api/share/rjwignar/VRYPZGK--yQ47nPgFu-ST):
```tsx
export type OnPromptFunction = (options?: {
  prompt?: string;
  imageUrls?: string[];
  retry?: boolean;
}) => void;
```

I wasn't sure where to place the reusable type, so I placed it in a new file, `src/lib/OnPromptFunction.ts`.

src/Chat/ChatBase.tsx
```diff
@@ -201,8 +202,10 @@ function ChatBase({ chat }: ChatBaseProps) {
   }, [messageListRef, handleScroll, shouldAutoScroll]);
 
   // Handle prompt form submission
-  const onPrompt = useCallback(
-    async (prompt?: string, imageUrls?: string[]) => {
+  const onPrompt: OnPromptFunction = useCallback(
+    async (options?: { prompt?: string; imageUrls?: string[]; retry?: boolean }) => {
```
humphd, line 206: Export a `PromptFunctionOptions` type too, so you don't have to write this out.

src/Chat/ChatBase.tsx
```diff
@@ -201,8 +202,10 @@ function ChatBase({ chat }: ChatBaseProps) {
...
+      let prompt = options?.prompt;
+      const { imageUrls, retry } = options || {};
```
humphd, line 208: I'm still confused why you don't use the pattern I mentioned and assign `= {}` as the default instead of doing this.

src/components/PromptForm/index.tsx
```diff
@@ -5,11 +5,12 @@ import DesktopPromptForm from "./DesktopPromptForm";
 import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
 import { useSettings } from "../../hooks/use-settings";
 import { ChatCraftChat } from "../../lib/ChatCraftChat";
+import { OnPromptFunction } from "../../lib/OnPromptFunction";
```
humphd, line 8: Move your exported types into this file.

humphd (changes requested):

src/Chat/ChatBase.tsx
rjwignar: Done.

src/components/PromptForm/index.tsx
rjwignar: I hope I did this correctly.
I put `PromptFunctionOptions` inside `OnPromptFunction.ts` with `OnPromptFunction`.

src/Chat/ChatBase.tsx
rjwignar: The pattern was fine for the `onPrompt` implementation in `ChatBase`, but when I tried using the same pattern throughout the `onPrompt/onSendClick` props TypeScript reports this error (props can't have default values):
image

To create a common type for both the onPrompt implementation and the onPrompt component props I had to use an `options` parameter and make that optional.

PR #618, opened 2024-04-27
Extend ImportCommand to fetch error messages from CloudFlare Worker error pages

### LINKED ISSUE #605: Errors when using /import on YouTube videos with no captions
While investing another issue I was using the /import feature and I noticed that [youtube-captions-scraper](https://www.npmjs.com/package/youtube-captions-scraper) throws errors for the following:
- no captions found
image
Video in question (note that subtitles are not enabled for this video):
https://www.youtube.com/watch?v=TKfS5zVfGBc

- no English captions found
image
Video in question (note that only Spanish subtitles are available): https://www.youtube.com/watch?v=y6120QOlsfU

In either case, ChatCraft will throw this toast error:
image

and this error on the console:
image

I think we can update the toast error to be a bit more descriptive by doing any of the following:
- show the specific URL or video ID for which /import failed
- specify why the import failed

rjwignar:
This fixes #605

Background
---
Currently, if you use /import with a Youtube video that either has no captions or English subtitles, ChatCraft will throw this Error Toast:

image

This Error Toast isn't very descriptive.

When /import successfully parses YouTube video captions, `res.text()` returns the actual content that populates the Message:
image.

However, **if /import** (when using [youtube-captions-scraper](https://github.com/algolia/youtube-captions-scraper/tree/master)) fails to do so (either due to [no captions in specified language](https://github.com/algolia/youtube-captions-scraper/blob/394ce9d61d2d2166f92ac394fa678c705e94fac8/src/index.js#L47-L49) or [no captions at all](https://github.com/algolia/youtube-captions-scraper/blob/394ce9d61d2d2166f92ac394fa678c705e94fac8/src/index.js#L30-L32)), **res.text() returns the HTML of a CloudFlare Worker error page**:

image

We only want the actual error message, nested inside the html content:
image

Solution
---

I tried modifying the error handling in [youtube-rewriter.ts](https://github.com/tarasglek/chatcraft.org/blob/main/functions/lib/rewriters/youtube-rewriter.ts) and in [proxy.ts](https://github.com/tarasglek/chatcraft.org/blob/main/functions/api/proxy.ts) (API handler for /import command) so **res.text()** would contain just the error message, but it seems no matter what I did, a bad response would always return HTML in **res.text()**.

Instead, I decided to use [Cheerio](https://github.com/cheeriojs/cheerio) (already installed) to extract the error message from any bad responses that have an HTML type. I kept the old error message in case other error responses don't contain HTML.

The result is that if /import fails on a YouTube video, the user will get a useful Error Toast telling them why:
**No English captions (https://www.youtube.com/watch?v=y6120QOlsfU)**:
image

**No captions at all (https://www.youtube.com/watch?v=TKfS5zVfGBc)**:
image

### DIFF
--- a/src/lib/commands/ImportCommand.ts
+++ b/src/lib/commands/ImportCommand.ts
@@ -60,7 +60,7 @@ export class ImportCommand extends ChatCraftCommand {
         // Parse the HTML string into a Document object
         const doc = parser.parseFromString(content, "text/html");
         // Extract error message
-        const errorMessage = doc.getElementsByClassName("error-message")[0].innerHTML;
+        const errorMessage = doc.getElementsByClassName("error-message")[0].innerHTML.trim();
         throw new Error(`${errorMessage}`);
       } else {
         throw new Error(`Unable to proxy request for URL: ${res.statusText}`);
--- a/src/lib/commands/ImportCommand.ts
+++ b/src/lib/commands/ImportCommand.ts
@@ -59,9 +59,9 @@ export class ImportCommand extends ChatCraftCommand {
         const parser = new DOMParser();
         // Parse the HTML string into a Document object
         const doc = parser.parseFromString(content, "text/html");
-        // Extract error message
-        const errorMessage = doc.getElementsByClassName("error-message")[0].innerHTML.trim();
-        throw new Error(`${errorMessage}`);
+        // Try to extract CloudFlare error message from HTML
+        const errorMessage = doc.querySelector(".error-message")?.innerHTML?.trim();
+        throw new Error(errorMessage ?? `Unable to proxy request for URL: ${res.statusText}`);
       } else {
         throw new Error(`Unable to proxy request for URL: ${res.statusText}`);
       }
--- a/src/lib/commands/ImportCommand.ts
+++ b/src/lib/commands/ImportCommand.ts
@@ -1,7 +1,7 @@
 import { ChatCraftCommand } from "../ChatCraftCommand";
 import { ChatCraftChat } from "../ChatCraftChat";
 import { ChatCraftHumanMessage } from "../ChatCraftMessage";
-import { load } from "cheerio";
+
 // To keep this small, just deal with some common cases vs doing proper parser/list
 const guessType = (contentType: string | null) => {
   if (!contentType) {
@@ -55,10 +55,12 @@ export class ImportCommand extends ChatCraftCommand {
     if (!res.ok) {
       // If res.text() is a CloudFlare Worker error page (HTML), extract the error message
       if (type === "html") {
-        // Load HTML content using Cheerio
-        const htmlContent = load(content);
-        // Error message located in 'error-message'
-        const errorMessage = htmlContent(".error-message").html();
+        // Create a new DOMParser instance
+        const parser = new DOMParser();
+        // Parse the HTML string into a Document object
+        const doc = parser.parseFromString(content, "text/html");
+        // Extract error message
+        const errorMessage = doc.getElementsByClassName("error-message")[0].innerHTML;
         throw new Error(`${errorMessage}`);
       } else {
         throw new Error(`Unable to proxy request for URL: ${res.statusText}`);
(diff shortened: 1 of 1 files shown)


src/lib/commands/ImportCommand.ts
```diff
@@ -1,7 +1,7 @@
 import { ChatCraftCommand } from "../ChatCraftCommand";
 import { ChatCraftChat } from "../ChatCraftChat";
 import { ChatCraftHumanMessage } from "../ChatCraftMessage";
-
+import { load } from "cheerio";
```
humphd, line 4: You don't need cheerio in a browser, since you have a full DOM already (these kinds of libraries are for situations where you don't have a full browser).

```TS
// Create a new DOMParser instance
const parser = new DOMParser();

// Parse the HTML string into a Document object
const doc = parser.parseFromString(htmlString, 'text/html');
```

Now you can query that `doc` for whatever you need to find, like a regular HTML page in the DOM.

humphd (changes requested):

src/lib/commands/ImportCommand.ts
rjwignar: Thanks for the suggestion!

I recently updated my PR to use DOMParser instead of cheerio.
I'm also ensuring that extracted error messages are trimmed:
image

src/lib/commands/ImportCommand.ts
```diff
@@ -49,13 +49,24 @@ export class ImportCommand extends ChatCraftCommand {
...
 
+    if (!res.ok) {
+      // If res.text() is a CloudFlare Worker error page (HTML), extract the error message
+      if (type === "html") {
+        // Create a new DOMParser instance
+        const parser = new DOMParser();
+        // Parse the HTML string into a Document object
+        const doc = parser.parseFromString(content, "text/html");
+        // Extract error message
+        const errorMessage = doc.getElementsByClassName("error-message")[0].innerHTML.trim();
```
humphd, line 63: Let's put some more safety around this string access and a fallback error message if it fails:

```ts
// Try to extract CloudFlare error message from HTML
const errorMessage = doc.querySelector(".error-message")?.innerHTML?.trim();
throw new Error(errorMessage ?? `Unable to proxy request for URL: ${res.statusText}`);
```

humphd (changes requested):

src/lib/commands/ImportCommand.ts
rjwignar: Thanks for this.
I've incorporated this into my code.

I'm still a little concerned about this PR.

I'm only fiddling with the CloudFlare HTML because it's the only way I currently know how to get the error message.
I tried adding error handling for these caption errors to the **getSubtitles** call in [YouTubeRewriter](https://github.com/tarasglek/chatcraft.org/blob/65597c70b3d0e484a4400a0836e39678aff871a3/functions/lib/rewriters/youtube-rewriter.ts#L23-L38) but the API call still returned a CloudFlare error page.

Is there a better way to grab these error messages?

humphd: https://github.com/algolia/youtube-captions-scraper/issues/15

rjwignar: It's funny. I found this post during my investigation and when I tried it out the errors were still returned in a CloudFlare Worker page, so I dismissed it.
However, that was because I was originally just throwing the caught error

I've given this another look, and I can get the **error message to show up as a ChatCraft message instead of an Error Toast** (changes not yet on this branch):
image
using this catch block:
```tsx
    } catch (error) {
      return new Response(error?.message, {
        headers: { "Content-Type": "text/markdown; charset=UTF-8" },
      });
    }
```
We now have a couple options on how to solve #605:
- modify the message in the error toast (current branch)
- put the error messages in a ChatCraft message (handle caught errors in YouTubeRewriter)
- however else we can handle caught errors in YouTubeRewriter

I'm kind of partial to putting the error message in a ChatCraft message, but I'd like another opinion on this.

rjwignar: @humphd I submitted a new PR that uses these changes instead of the changes I made to `src/lib/commands/ImportCommand.ts`. Please let me know which solution you prefer.

humphd: Let's do the other one.

rjwignar:
Closing as we're exploring #625 instead

PR #625, opened 2024-05-01
Update youtube-rewriter error handling to pass captioning errors as ChatCraft message

### LINKED ISSUE #605: Errors when using /import on YouTube videos with no captions

rjwignar:
This fixes #605 

Background
---

If you use /import with a Youtube video that either has no English captions or no captions at all, ChatCraft will throw a non-descriptive error toast:
image

Caption extraction for YouTube videos is handled by [youtube-captions-scraper](https://github.com/algolia/youtube-captions-scraper/tree/master) within [functions/lib/rewriters/youtube-rewriter.ts](https://github.com/tarasglek/chatcraft.org/blob/ad93ffccbee020f1df97036bb9a7c05bba533d39/functions/lib/rewriters/youtube-rewriter.ts).

A previous PR for this (#618) would extract unhandled error messages from the HTML of a generated CloudFlare Pages page and put it inside an Error Toast.

Fix
---

In this PR, any errors thrown in youtube-rewriter will show up in the **/import** Response (as a ChatCraft message):
image

### DIFF
--- a/functions/lib/rewriters/youtube-rewriter.ts
+++ b/functions/lib/rewriters/youtube-rewriter.ts
@@ -28,20 +28,26 @@ export class YouTubeRewriter extends DefaultRewriter {
   }
 
   async fetchData(url: URL) {
-    const videoInfo = getVideoId(url.href);
-    if (!videoInfo.id) {
-      throw new Error(`Unable to parse YouTube video ID from URL ${url.href}`);
-    }
+    try {
+      const videoInfo = getVideoId(url.href);
+      if (!videoInfo.id) {
+        throw new Error(`Unable to parse YouTube video ID from URL ${url.href}`);
+      }
 
-    const videoId = videoInfo.id;
-    // TODO: could allow passing lang so we can pass to getSubtitles
-    const captions: WebVTTCaption[] = await getSubtitles({ videoID: videoInfo.id, lang: "en" });
-    const text = `[![http://www.youtube.com/watch?v=${videoId}](http://img.youtube.com/vi/${videoId}/0.jpg)](http://www.youtube.com/watch?v=${videoId})
+      const videoId = videoInfo.id;
+      // TODO: could allow passing lang so we can pass to getSubtitles
+      const captions: WebVTTCaption[] = await getSubtitles({ videoID: videoInfo.id, lang: "en" });
+      const text = `[![http://www.youtube.com/watch?v=${videoId}](http://img.youtube.com/vi/${videoId}/0.jpg)](http://www.youtube.com/watch?v=${videoId})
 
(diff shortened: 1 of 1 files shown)


functions/lib/rewriters/youtube-rewriter.ts
```diff
@@ -28,20 +28,26 @@ export class YouTubeRewriter extends DefaultRewriter {
...
 
-    return new Response(text, {
-      headers: { "Content-Type": "text/markdown; charset=UTF-8" },
-    });
+      return new Response(text, {
+        headers: { "Content-Type": "text/markdown; charset=UTF-8" },
+      });
+    } catch (error) {
+      return new Response(`**Error**: ${error?.message}`, {
+        headers: { "Content-Type": "text/markdown; charset=UTF-8" },
```
humphd, line 49: This is probably OK, but I've never seen Markdown used as an Error body before.

humphd (approved):
Can we remove the other fix you made before, or does it still have value?

functions/lib/rewriters/youtube-rewriter.ts
rjwignar: I wanted to bold it like the "Command" text, but I hadn't considered whether we should include Markdown in an Error body.
I can remove the Markdown for now

humphd: Usually an error is informational, not stylistic.  You don't want to assume too much about how the client will use it.

rjwignar:
> Can we remove the other fix you made before, or does it still have value?

I had both fixes in one branch and found that only this fix would be triggered if an error occured (there'd be a ChatCraft message but no Error Toast).
We can remove the other fix.

functions/lib/rewriters/youtube-rewriter.ts
rjwignar: I recently removed Markdown styling from the error message:
```tsx
return new Response(`Error: ${error?.message}`, {
```

humphd (approved):

PR #627, opened 2024-05-04
Add custom progress alert and integrate with export as audio

### LINKED ISSUE #604: Show audio download status with a progress bar
#602 added the ability to download any length message as audio, with limited exceptions (a single sentence cannot be longer than 4096 characters).

However, when the message to be downloaded is too long, it sometimes takes a while for the download to complete and users can be confused if something failed or the download is still happening.

To fix this, we can show a **progress bar** in the toast and update its value as the audio chunks are downloaded.

Amnish04:
As discussed in https://github.com/tarasglek/chatcraft.org/pull/602#pullrequestreview-2010490014,

I have added a custom progress bar toast to our existing `useAlert` hook.

1. Now we see this progress bar toast when trying to download a message as audio, and the value updates as the promises processing text to speech chunks get resolved.
2. I am now also clearing the `pLimit` queue if we ever run into an exception as I noticed the queued promises continued executing even though one of them failed.
3. I am not sure what color would be the best for our progress bar and have selected the best combination I could come up with. I could use some help with color selection :)
4. I also found that `tts-1-hd` model has a rate limit of `3` on non-enterprise openai keys, and had to switch to standard `tts-1` model for local testing. We should probably have another issue to add a preference control so users can choose which model they prefer for audio downloads.

**Demo:** (Ignore distorted color, the gif is corrupt)
CustomProgressBar

This fixes #604

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -255,17 +255,16 @@ function MessageBase({
           progressPercentage: 0,
         });
 
+        // Limit the number of concurrent tasks
+        const pLimit = (await import("p-limit")).default;
+        const limit = pLimit(8); // Adjust the concurrency limit as needed
+
         try {
           const textChunks = getSentenceChunksFrom(text, 500);
           const chunksToBeProcessed = textChunks.length;
           const audioClips: Blob[] = new Array<Blob>(chunksToBeProcessed);
 
           let chunksProcessed = 0;
-
-          // Limit the number of concurrent tasks
-          const pLimit = (await import("p-limit")).default;
-
-          const limit = pLimit(8); // Adjust the concurrency limit as needed
 
           const tasks = textChunks.map((textChunk, index) => {
             return limit(async () => {
@@ -279,12 +278,13 @@ function MessageBase({
               audioClips[index] = audioClip;
 
               ++chunksProcessed;
-              const processedPercentage = Math.floor(chunksProcessed / chunksToBeProcessed);
+              const processedPercentage = Math.floor((chunksProcessed * 100) / chunksToBeProcessed);
               progress({
                 id: alertId,
                 title: "Downloading...",
                 message: "Please wait while we prepare your audio download.",
                 progressPercentage: processedPercentage,
+                updateOnly: true,
               });
             });
           });
@@ -310,6 +310,7 @@ function MessageBase({
 
           closeToast(alertId);
           error({ title: "Error while downloading audio", message: err.message });
+          limit.clearQueue();
         }
       }
     }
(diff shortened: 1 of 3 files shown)


tarasglek:
This is pretty cool. Also unsure re colors. For rate limits we should maybe try exponential backoff?

humphd (commented):
Can we add a "cancel" button to this alert as well?  Since it's a long-running task, I can see people wanting to get out of it while it's running.

I would say this is probably in pretty good shape to land and iterate on as we use it.

Amnish04:
@humphd Just added a button to minimize the toast, but not sure if you meant cancelling the download altogether.

Amnish04:
> This is pretty cool. Also unsure re colors. For rate limits we should maybe try exponential backoff?

@tarasglek So you mean we should start adding delay between our api calls until we start getting a successful result? But wouldn't that make the downloads too long, as we are just allowed to call `tts-1-hd` 3 times every **minute**?

tarasglek:
> > This is pretty cool. Also unsure re colors. For rate limits we should maybe try exponential backoff?
> 
> @tarasglek So you mean we should start adding delay between our api calls until we start getting a successful result? But wouldn't that make the downloads too long, as we are just allowed to call `tts-1-hd` 3 times every **minute**?

If we have download progress and a cancel button, no problem to wait for HQ ver :) I would personally prefer higher quality over faster generation for a lot of my usecases.

Amnish04:
> If we have download progress and a cancel button, no problem to wait for HQ ver :) I would personally prefer higher quality over faster generation for a lot of my usecases.

@tarasglek We could still use the HQ version if we have a preference setting for that :)

But the idea of allowing users to download HQ audio even with rate limiting also sounds cool to me.

I have filed a follow-up for it.

https://github.com/tarasglek/chatcraft.org/issues/632

humphd (commented):
This looks good, but there is one UI bug I notice with the close button:

Screenshot 2024-05-07 at 7 35 00 PM

Amnish04:
> This looks good, but there is one UI bug I notice with the close button:
> Screenshot 2024-05-07 at 7 35 00 PM

Does it get stuck on clicking that button? I can't reproduce

humphd:
It's the `-` vs. `x` that I mean.

Amnish04:
> It's the `-` vs. `x` that I mean.

Oh, that was intentional, as it felt more natural for **minimize** vs **cancel**.

But I can change it for sake of consistency.

humphd:
It should only be "cancel" imho, don't do minimize.

Amnish04:
@humphd Just changed minimize behaviour to cancel download.

humphd (approved):
lgtm

PR #628, opened 2024-05-04
Don't show free mode banner if an API key has been added (check settings.providers.length)

### LINKED ISSUE #544: Don't show "free mode" banner when multiple providers are set
We show a "You are using the free model..." banner when the user first comes to ChatCraft and we want to let them know about adding more providers for better features.

However, even after you add an OpenAI or OpenRouter provider, when you switch back to the free provider, we show the banner again.

Let's improve our check for the free provider banner so that it looks to see if the free provider is the *only* provider you have configured.  If you have more than one, don't show it.

rjwignar:
This fixes #544.

Sorry for sitting on this for a while. There's one concern I have that I wasn't able to figure out.
More details below.

Background
---
> We show a "You are using the free model..." banner when the user first comes to ChatCraft and we want to let them know about adding more providers for better features.
However, even after you add an OpenAI or OpenRouter provider, when you switch back to the free provider, we show the banner again.

Solution
---
If you run `console.log(settings)` you'll see that `settings` has a property called `providers`.
`settings.providers` is empty when no provider API keys are added (length 0):
image

...and is populated when you add an API key:
image

This PR does the following:
- adds a useEffect that toggles the Free Banner based on the length of `settings.provider`:
  - length=0 (no API Keys added) - Show banner
  - length>0 (1+ API Key(s) added) - Don't show banner
-  defaults showAlert to `false` so banner doesn't appear temporarily when user opens ChatCraft after adding a provider/API key

My concern
---
The useMemo no longer checks the current provider when showing the Free Mode banner.
This means if a user has a paid provider (like OpenAI) as the current provider and removes their API key, the **Free Mode banner will re-appear**, but the currentProvider will still be OpenAI:
freeBannerIgnoreCurrentProvider

We can revert back to the **original behaviour** if that's preferred (don't show Free Mode if Current Provider is a Paid provider):
freeBannerCheckCurrentProvider.

I'd like some opinions on this.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -61,7 +61,7 @@ function ChatBase({ chat }: ChatBaseProps) {
   const { error } = useAlert();
   const { user } = useUser();
   const { clearAudioQueue } = useAudioPlayer();
-  const [showAlert, setShowAlert] = useState(true);
+  const [showAlert, setShowAlert] = useState(false);
   const {
     isOpen: isPrefModalOpen,
     onOpen: onPrefModalOpen,
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -68,6 +68,11 @@ function ChatBase({ chat }: ChatBaseProps) {
     onClose: onPrefModalClose,
   } = useDisclosure();
 
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


humphd (approved):
I think this is the right behaviour.  Once you've used the app for a bit, and played around with providers, we don't need to bother you any more with whether this is the free one or not.

Issue #630, opened 2024-05-07
Can't cancel model responses


tarasglek:
https://github.com/tarasglek/chatcraft.org/assets/857083/57ebf7b6-a4d6-4fae-83f3-9fb2b2ffda02

tarasglek:
I think this is related to #629 cos I also can't dismiss those red error boxes when this happens. Seems like whole webpage eventloop is dead

Issue #629, opened 2024-05-07
Chatcraft hangs webpage


tarasglek:
This is a bugreport from twitter. I've noticed similar weird behavior where chrome seemingly freezes while streaming responses.

I found a UI bug on chatcraft. Asked: "What is apache apisix?" using the free models in auto mode. And I got a response repeating APACHE APACHE APACHE after a normal sentence. The UI bug is not the model response, but that that some JavaScript code hangs the Firefox tab, until Firefox ask to stop the JS code running there. Sorry, I didn't found any feedback option on the website.

Note, it only hangs on a saved conversation, not on a clean one, or one that repeats the same questions (two unrelated ones)

robmv:
Sometimes  it displays the repeated APACHE words faster., it may be a Firefox redisplay issue, I switched this time to another tab and back and the words where shown at the end of the video

[Screencast from 2024-05-07 08-47-23.webm](https://github.com/tarasglek/chatcraft.org/assets/1652983/e5e92d21-aaf0-4046-8ff2-967cdd7d7404)

The stack trace after Firefox ask for the script being stopped is:

```
Script terminated by timeout at:
oR@https://chatcraft.org/assets/index-Dl5d8yCF.js:603:53627
yFe@https://chatcraft.org/assets/index-Dl5d8yCF.js:603:55638
Bze@https://chatcraft.org/assets/index-Dl5d8yCF.js:609:8007
Fze@https://chatcraft.org/assets/index-Dl5d8yCF.js:609:8441
Pze/e.prototype.sweep@https://chatcraft.org/assets/index-Dl5d8yCF.js:609:5822
l5t@https://chatcraft.org/assets/index-Dl5d8yCF.js:1073:63873
compute/<@https://chatcraft.org/assets/index-Dl5d8yCF.js:603:8626
compute@https://chatcraft.org/assets/index-Dl5d8yCF.js:603:8582
Pt@https://chatcraft.org/assets/index-Dl5d8yCF.js:603:15281
cR@https://chatcraft.org/assets/index-Dl5d8yCF.js:613:1225
onData@https://chatcraft.org/assets/index-Dl5d8yCF.js:1073:127315
O@https://chatcraft.org/assets/index-Dl5d8yCF.js:586:327
lae/V<@https://chatcraft.org/assets/index-Dl5d8yCF.js:586:995
```

humphd:
@robmv I can't reproduce (yet).  Do you see this in Chrome as well?

[...]

PR #638, opened 2024-05-16
Use requestIdleCallback to avoid blocking main thread

humphd:
Fixes #629
Fixes #630

This schedules work when responses are being streamed in so we don't swamp the main thread with too many re-renders.

@robmv, @tarasglek can you two try this on the preview branch?  I've tried a few approaches and they all have some trade offs.

PR #640, opened 2024-05-17
alternative requestAnimationFrame fix

tarasglek:
Fixes https://github.com/tarasglek/chatcraft.org/issues/629
Fixes https://github.com/tarasglek/chatcraft.org/issues/630

Implementing it cos https://github.com/tarasglek/chatcraft.org/pull/638 didnt work for me

### DIFF
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -194,6 +194,7 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
   const buffer: string[] = [];
   let functionName: string = "";
   let functionArgs: string = "";
+
   const streamOpenAIResponse = async (token: string, func: string, args: string) => {
     if (func || args) {
       functionName += func;
@@ -202,9 +203,17 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
         onData({ token: func, currentText: functionArgs });
       }
     } else if (token) {
+      console.log(token);
       buffer.push(token);
       if (onData && !isPaused) {
-        onData({ token, currentText: buffer.join("") });
+        // wait for UI to be ready to render via requestingFramePromise before proceeding
+        const requestingFramePromise: Promise<void> = new Promise((resolve) => {
+          requestAnimationFrame(() => {
+            onData({ token, currentText: buffer.join("") });
+            resolve();
+          });
+        });
+        await requestingFramePromise;
       }
     }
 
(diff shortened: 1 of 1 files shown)


tarasglek:
This currently throttles llm responses to as fast as we can draw..need to make that promise into a member variable so it can buffer multiple tokens while waiting, but dont understand what the token: thing is for

humphd:
Here's a concept for tidying up the code a bit:

```ts
const createStreamOpenAIResponseHandler = () => {
  const buffer: string[] = [];
  let functionName: string = "";
  let functionArgs: string = "";
  let maxWaitTimeMS = 0;
  let waitingTokens: string[] = [];
  let streamOpenAIResponsePromise: Promise<void> | null = null;

  const handleTokenRendering = (): Promise<void> => {
    const startTime = performance.now();
    return new Promise((resolve) => {
      requestAnimationFrame(() => {
        onData({
          token: waitingTokens.join(""),
          currentText: buffer.join(""),
        });
        if (waitingTokens.length > 1) {
          console.log("waitingTokens", waitingTokens.length);
        }
        waitingTokens = [];
        const endTime = performance.now();
        const waitTime = endTime - startTime;
        if (waitTime > maxWaitTimeMS && waitTime > 16) {
          maxWaitTimeMS = waitTime;
          console.log("streamOpenAIResponse maxWaitTimeMS", maxWaitTimeMS);
        }
        streamOpenAIResponsePromise = null;
        resolve();
      });
    });
  };

  const streamOpenAIResponse = async (
    token: string,
    func: string,
    args: string
  ): Promise<void> => {
    if (func || args) {
      functionName += func;
      functionArgs += args;
      if (onData && !isPaused) {
        onData({ token: func, currentText: functionArgs });
      }
    } else if (token) {
      buffer.push(token);
      if (onData && !isPaused) {
        waitingTokens.push(token);
        if (!streamOpenAIResponsePromise) {
          streamOpenAIResponsePromise = handleTokenRendering();
        }
      }
    }
  };

  const chunkStream = async (
    streamResponse: Stream<ChatCompletionChunk>
  ): Promise<void> => {
    for await (const streamChunk of streamResponse) {
      const parsedData = parseOpenAIChunkResponse(streamChunk);
      await streamOpenAIResponse(
        parsedData.token,
        parsedData.functionName,
        parsedData.functionArgs
      );
    }
  };

  const renderResponse = async (): Promise<void> => {
    if (streamOpenAIResponsePromise) {
      await streamOpenAIResponsePromise;
    }
    const content = buffer.join("");
    return handleOpenAIResponse(content, functionName, functionArgs);
  };

  return { chunkStream, renderResponse };
};

const { chunkStream, renderResponse } = createStreamOpenAIResponseHandler();

const handleStreamingResponse = async (
  streamResponse: Stream<ChatCompletionChunk>
): Promise<void> => {
  await chunkStream(streamResponse);
  await renderResponse();
};
```

tarasglek:
I'm ok with however you want to structure this code. Please change to your liking and land?

humphd:
I think this is great.  Thanks for your help iterating on the idea.  I couldn't get an rAF-based version to work this well.

PR #642, opened 2024-05-30
Change Providers/Models order in Ask menu

### LINKED ISSUE #641: Scroll models, but not providers in Ask menu
We have 2 things in the Ask menu:

- models for the chosen provider
- providers

The models list can be really long (OpenRouter.ai), and it needs to overflow.  However, right now, we also scroll the providers out of sight:

Screenshot 2024-05-20 at 5 45 52 PM

It would be nice if we didn't scroll the providers away, and only scrolled the models.  The Search at the top should probably also stay anchored.

menghif:
Fixes #641 

This PR moves the Providers list above the Models list in the Ask menu.

Additionally, I changed the selection icon to a checkmark for better clarity. I aligned the menu to its button and changed the variant of the search bar.

### DIFF
--- a/src/components/PromptForm/PromptSendButton.tsx
+++ b/src/components/PromptForm/PromptSendButton.tsx
@@ -132,7 +132,10 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
                 variant="outline"
                 placeholder="Search models..."
                 value={searchQuery}
-                onChange={(e) => setSearchQuery(e.target.value)}
+                onChange={(e) => {
+                  e.preventDefault();
+                  setSearchQuery(e.target.value);
+                }}
               />
             </InputGroup>
             {models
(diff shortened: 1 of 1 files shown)


humphd:
I love how clean this looks @menghif!

Question: can we do something about how the `overflow` gets applied, so it stays limited to the models vs. the entire menu.  On providers with a lot of models, you lose track easily:


https://github.com/tarasglek/chatcraft.org/assets/427398/15476be0-0cc1-48c1-bbbc-693736cc2e17

If that's too complicated, we can leave that for a follow-up.

menghif:
@humphd Yesterday I tried changing the overflow to only the Models list but I wasn't sure how it would behave with a long list of providers. I only tested it with up to 3.

humphd:
This works well for me (sorry for lower video quality, had to reduce it to fit in 10M):

https://github.com/tarasglek/chatcraft.org/assets/427398/d44c79c7-60cd-48de-98ab-cc8c480ea075

I'd say "ship it."

Anyone else have feedback before we merge?

menghif:
One last thing I noticed from your video is that I messed up the Menu placement. It should be `top-end`. Fixed it now!

tarasglek:
this a nice improvement

humphd (approved):
@menghif land this when you're ready.  Code looks good, r+ from me.

tarasglek:
This is a bit weird in mobile as we now cover up search results by keyboard on ios :(

menghif:
> This is a bit weird in mobile as we now cover up search results by keyboard on ios :(

I'll look into how to fix it.

PR #644, opened 2024-06-01
Single source of truth for commands help text

Amnish04:
I always felt that the way we manage commands help text documentation could be done better.
Currently, every time we add a new command we have to manually go and add the corresponding documentation to the `Help` component, which many new developers might not even know about.

In this PR,
1. I have added a `metaData` property to `ChatCraftCommand` class, that takes a `helpTitle` and `helpDescription` as a markdown string.
2. The `ChatCraftCommandRegistry` now exposes a list of all the supported command objects, that is used to generate the relevant documentation markdown for the commands table (instead of static text)
3. Making the `metaData` a mandatory argument in `ChatCraftCommand` constructor also ensures that if anyone adds a new command in the future, they cannot get away without adding corresponding documentation to it.

I did not file an issue prior to this, as I was just experimenting and it turned out to work well.

### DIFF
--- a/src/lib/ChatCraftCommand.ts
+++ b/src/lib/ChatCraftCommand.ts
@@ -1,14 +1,16 @@
 import { ChatCraftChat } from "./ChatCraftChat";
 
-type ChatCraftCommandMetaData = {
-  helpTitle: string; // Help strings can be
-  helpDescription: string; // markdown text
-};
-
 export abstract class ChatCraftCommand {
+  /**
+   *
+   * @param command Name of the command to be used in format - `/<command-name>`
+   * @param helpTitle Title of the command to be displayed in help grid
+   * @param helpDescription A brief paragraph explaining how the command works, and any supported options
+   */
   constructor(
     public command: string,
-    public metaData: ChatCraftCommandMetaData
+    public helpTitle: string,
+    public helpDescription: string
   ) {}
 
   // This method should be overridden by subclasses to implement the command
--- a/src/lib/ChatCraftCommand.ts
+++ b/src/lib/ChatCraftCommand.ts
@@ -1,7 +1,15 @@
 import { ChatCraftChat } from "./ChatCraftChat";
 
+type ChatCraftCommandMetaData = {
+  helpTitle: string; // Help strings can be
+  helpDescription: string; // markdown text
+};
+
 export abstract class ChatCraftCommand {
-  constructor(public command: string) {}
+  constructor(
+    public command: string,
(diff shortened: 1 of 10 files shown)


src/lib/ChatCraftCommand.ts
```diff
@@ -1,7 +1,15 @@
...
+type ChatCraftCommandMetaData = {
+  helpTitle: string; // Help strings can be
+  helpDescription: string; // markdown text
+};
+
 export abstract class ChatCraftCommand {
-  constructor(public command: string) {}
+  constructor(
+    public command: string,
+    public metaData: ChatCraftCommandMetaData
```
humphd, line 11: Seeing the mixed casing in `metaData` is a bad code smell (e.g., `metadata` vs. `metaData`) and a lot of times you can solve this by simply not using a name.  Here, I'd suggest we add `title` and `description` to the `ChatCraftCommand` instead.

src/lib/ChatCraftCommandRegistry.ts
```diff
@@ -25,6 +25,14 @@ export class ChatCraftCommandRegistry {
       command.execute(chat, user, parsed.args);
   }
 
+  static getCommandsList() {
```
humphd, line 28:
```ts
return Array.from(this.commands.values())
```

I'd call this `getCommands()` (drop the `List`).

humphd (changes requested):

src/lib/ChatCraftCommandRegistry.ts
Amnish04: Renamed in the latest commit

src/lib/ChatCraftCommand.ts
Amnish04: I see. I have replaced metadata object with flattened properties based on your advise. I did not remove the `help` prefix though as I thought only `title` and `description` could be ambiguous.

Please let me know if we need further changes.

Amnish04:
@humphd Not sure why, I've started getting a lint error in a file I did not even change.

https://github.com/tarasglek/chatcraft.org/actions/runs/9333938306/job/25691497752?pr=644

tarasglek:
> @humphd Not sure why, I've started getting a lint error in a file I did not even change.
> 
> https://github.com/tarasglek/chatcraft.org/actions/runs/9333938306/job/25691497752?pr=644
> 
> **Update:** I think its related to @tarasglek's latest commit, and #646 should fix it.

sorry, was distracted and forgot to check on first landing and then forgot to land the fixup

PR #647, opened 2024-06-03
Add a stop announcing button for TTS

### LINKED ISSUE #521: Make text to speech have more obvious Play/Pause UX
When using real-time TTS streaming, we can stop announcing a message by clicking on toggle TTS button.
image

But there is no way elegant way to stop announcing a message when using Speak option.

We need to figure out a better way to handle this.

Amnish04:
I've been thinking for a long time to have a dedicated control that can stop text to speech announcements for a particular message, instead of having to disable and enable TTS setting that is persisted in local storage.

In this PR:
1. I have added a `DisableAudioPlayerButton` component that normally stays off screen but appears when ChatCraft is streaming any audio in top left part of view port. I chose fixed positioning as I couldn't find enough real estate in the page where it could fit while being easily visible.
2. It works by clearing any audio clips in the audio queue and temporarily disabling it so no new audio clips are added as the response gets generated. The queue is automatically enabled once the message has finished streaming.
3. It also has a tooltip suggesting that "ChatCraft is Speaking" and clicking it would stop the current announcement.
4. It works with both real time streaming and **Speak** option.

StopAnnouncementButton

I could probably position this better. Might need some help in figuring out.

This fixes #521


humphd:
Nice idea.  A few quick thoughts:

- `StopSpeechAudioButton` or `Interrupt...` might be better than `DisableAudioPlayerButton`, which implies that you're disabling it vs. stoping the in-progress action.
- I would move this either beside the button you use to get speech to start, or put it with the `Pause`/`Cancel` buttons.

Amnish04:
>     * `StopSpeechAudioButton` or `Interrupt...` might be better than `DisableAudioPlayerButton`, which implies that you're disabling it vs. stoping the in-progress action.

@humphd Agreed, I have renamed the button component.

Another thing I have added is that the button is `solid` variants only when an audio is playing.

>     * I would move this either beside the button you use to get speech to start, or put it with the `Pause`/`Cancel` buttons.

For the positioning, we probably cannot place it with `Pause/Cancel` button, since this is also for scenarios when a message is not streaming (Speak option and Audio Previews).

Also, by "beside the button you use to get speech to start", do you mean with these buttons?
Screenshot 2024-06-03 at 6 27 47 PM

humphd:
We should look at how other apps solve the positioning problem.  This must be something that has visual design patterns we can leverage, but I don't tend to use a ton of TTS stuff myself.

Amnish04:
@humphd I reviewed how ChatGPT handles the stop audio scenario and re-designed to something following similar pattern.

Whenever any audio clip is playing, the toggle TTS button is replaced with the interrupt audio button, with a custom bouncing balls animation indicating audio playback.

 
InterruptSpeechButton

tarasglek:
May I suggest animation similar to [this](https://www.istockphoto.com/uk/%D0%B2%D1%96%D0%B4%D0%B5%D0%BE/%D0%B7%D0%B2%D1%83%D0%BA%D0%BE%D0%B2%D0%B0-%D1%85%D0%B2%D0%B8%D0%BB%D1%8F-%D0%B0%D0%B1%D0%BE-%D1%87%D0%B0%D1%81%D1%82%D0%BE%D1%82%D0%B0-%D1%86%D0%B8%D1%84%D1%80%D0%BE%D0%B2%D0%BE%D0%B3%D0%BE-%D1%80%D1%83%D1%85%D1%83-gm1348366967-425584056)?

Amnish04:
> May I suggest animation similar to [this](https://www.istockphoto.com/uk/%D0%B2%D1%96%D0%B4%D0%B5%D0%BE/%D0%B7%D0%B2%D1%83%D0%BA%D0%BE%D0%B2%D0%B0-%D1%85%D0%B2%D0%B8%D0%BB%D1%8F-%D0%B0%D0%B1%D0%BE-%D1%87%D0%B0%D1%81%D1%82%D0%BE%D1%82%D0%B0-%D1%86%D0%B8%D1%84%D1%80%D0%BE%D0%B2%D0%BE%D0%B3%D0%BE-%D1%80%D1%83%D1%85%D1%83-gm1348366967-425584056)?

@tarasglek Sure, I can try to create that

Amnish04:
@tarasglek Just added another variant to the component, imitating the type of animation you mentioned

tarasglek:
perfect

src/components/InterruptSpeechButton.tsx
```diff
@@ -0,0 +1,184 @@
...
+  IconButton,
+  IconButtonProps,
+  Tooltip,
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { motion, AnimationProps, MotionProps } from "framer-motion";
+import { useCallback, useMemo, useState } from "react";
+import useAudioPlayer from "../hooks/use-audio-player";
+
+//#region AudioPlayingIcon
```
humphd, line 13: What is this?

src/components/InterruptSpeechButton.tsx
```diff
@@ -0,0 +1,184 @@
...
+          duration: 1.5,
+          ease: "easeInOut",
+          repeat: Infinity,
+        },
+      };
+    }, []);
+
+  return (
+    <Box
+      as={Flex}
```
humphd, line 100: You could use `<Flex>` for this instead.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -43,7 +44,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
     return !!models.filter((model) => model.id.includes("tts"))?.length;
   }, [models]);
 
-  const { clearAudioQueue } = useAudioPlayer();
+  const { clearAudioQueue, isQueueEmpty } = useAudioPlayer();
```
humphd, line 47: `isAudioQueueEmpty` would be more clear

humphd (changes requested):
This is great.  Some small nits, but let's go with this.

src/components/InterruptSpeechButton.tsx
Amnish04: Right, updated in latest commit

src/components/PromptForm/PromptSendButton.tsx
Amnish04: Makes sense, I've changed the name in latest commit

src/components/InterruptSpeechButton.tsx
Amnish04: I usually use this to mark different regions and group items in a file where I have multiple things.

I can remove it if it causes confusion.

Amnish04:
@humphd Just rebased.

humphd (approved):

PR #648, opened 2024-06-06
Fix for mobile Ask menu and z-index fixes

menghif:
After the change in https://github.com/tarasglek/chatcraft.org/pull/642, on iOS the keyboard covers the search results in the Ask Menu.

I fixed this by moving the search bar to the bottom. Bonus: it's now easier to reach!

I left the Desktop version as is, but let me know if I should change it to match the mobile version.

---

I also fixed these two visual bugs:

Show More Button overlapping the Favourite Providers List

Sidebar Menu Button overlapping the Settings Menu

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -144,7 +144,7 @@ function Header({ chatId, inputPromptRef, searchText, onToggleSidebar }: HeaderP
           onClick={toggleColorMode}
         />
 
-        <Box>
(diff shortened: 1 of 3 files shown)


tarasglek:
This fixes the issue for me. thanks for the quick turnaround!

menghif:
Is this good to be merged?

humphd (approved):
Land it

PR #649, opened 2024-06-06
Connect RAG to chatcraft

PreciousNiphemi:


tarasglek:
Hi @PreciousNiphemi, what's your thinking of rag integration into chatcraft?

One interesting way to do that would be to integrate cohere command r+ support which has rag support on completion endpoint.

PR #650, opened 2024-06-06
Use Exponential Backoff for downloading large messages

### LINKED ISSUE #632: Explore implementing exponential backoff for HQ TTS downloads
On non-enterprise accounts, the limit for `tts-1-hd` model is just 3 RPM, which makes it impossible for people to download longer messages as audio with lower RPM keys.

As discussed in https://github.com/tarasglek/chatcraft.org/pull/627#issuecomment-2095352160,

we could implement [exponential backoff](https://platform.openai.com/docs/guides/rate-limits/retrying-with-exponential-backoff) strategy to limit the rate at which TTS download requests are sent if we start getting rate limit errors.

image

This library could be of help, but need to research more.
https://www.npmjs.com/package/exponential-backoff

We could also have a setting, so user can choose between `tts-1` (faster downloads) or `tts-1-hd` (high quality) as some downloads could be **extremely** slow with 3 RPM.

Amnish04:
In https://github.com/tarasglek/chatcraft.org/pull/627#issuecomment-2095352160, I was advised to use exponential backoff to counter rate limiting of **3 RPM** on HQ TTS downloads.

In this PR, I am using the [exponential-backoff](https://www.npmjs.com/package/exponential-backoff?activeTab=readme) package to retry any failed operations with increasing delays between failed attempts.

The package size was around 36 kb, so I am using dynamic import to only load it when needed.

Here's how the network tab looks when downloading,

ExponentialBackoff

I think I can optimize it more by increasing the `startingDelay` value, so we don't bombard so many requests. Not sure what else could be better.

This fixes #632

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "dexie-export-import": "^4.1.2",
     "dexie-react-hooks": "^1.1.7",
     "esbuild-wasm": "^0.21.3",
+    "exponential-backoff": "^3.1.1",
     "feed": "^4.2.2",
     "framer-motion": "^11.2.3",
     "get-video-id": "^3.6.5",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -50,6 +50,9 @@ dependencies:
   esbuild-wasm:
     specifier: ^0.21.3
     version: 0.21.3
+  exponential-backoff:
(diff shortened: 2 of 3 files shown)


humphd (approved):
Let's try it and tweak.

PR #652, opened 2024-06-08
Include message card in Image Export so dark mode text is visible

### LINKED ISSUE #651: Export as Image doesn't work in dark mode
"Export as Image" doesn't include text when exported from dark mode:

| image in light mode | image in dark mode |
|----|-----|

rjwignar:
This fixes #651

The white text in the dark mode card body is camouflaged by the white background of [html2canvas](https://html2canvas.hertzen.com/).

One solution to make this text visible is to include the entire card (which wraps the text) in the HTML sent to html2canvas:
**Before (card body only, white text invisible):**
message
**After (entire card, white text appears as rendered inside card:**
message

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -127,6 +127,7 @@ function MessageBase({
   const isNarrowScreen = useMobileBreakpoint();
   const messageForm = useRef<HTMLFormElement>(null);
   const messageContent = useRef<HTMLDivElement>(null);
+  const messageCard = useRef<HTMLDivElement>(null);
   const meta = useMemo(getMetaKey, []);
   const [imageModalOpen, setImageModalOpen] = useState<boolean>(false);
   const [selectedImage, setSelectedImage] = useState<string>("");
@@ -210,7 +211,7 @@ function MessageBase({
   }, [info, text]);
 
   const handleDownloadImage = useCallback(() => {
-    const elem = messageContent.current;
+    const elem = messageCard.current;
     if (!elem) {
       return;
     }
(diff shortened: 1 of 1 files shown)


rjwignar:
I must add that this change has a visual bug in which the background colours invert:
**On Edge**:
message 
**On  Chrome:**
message

On the Chrome screenshot you can see the message options menu didn't completely close.
I wonder if we should add a delay between clicking "Export as Image" and html2canvas taking a screenshot

src/lib/utils.ts
```diff
@@ -65,10 +65,7 @@ export const screenshotElement = (element: HTMLElement): Promise<Blob> => {
   return import("html2canvas")
     .then((module) => {
       const html2canvas = module.default;
-      return html2canvas(element, {
-        windowWidth: element.scrollWidth,
-        windowHeight: element.scrollHeight,
-      });
+      return html2canvas(element, {});
```
rjwignar, line 68: Since the options object is empty (`{}`) we can remove it from our code.

I'm wondering if we should replace it with a Partial<Options> object so maintainers know we can configure options for it:
```ts
import { Options } from "html2canvas";
// ...
export const screenshotElement = (element: HTMLElement): Promise<Blob> => {
  return import("html2canvas")
    .then((module) => {
      const options: Partial<Options> = {};
      const html2canvas = module.default;
      return html2canvas(element, options);
    })
    .then(
    // ...
    );
};
```

menghif:
> On the Chrome screenshot you can see the message options menu didn't completely close.
I wonder if we should add a delay between clicking "Export as Image" and html2canvas taking a screenshot

I tested this and I found the same problem you described on Firefox and Safari (surprisingly not on Chrome after many attempts). I think adding a delay is a good idea.


screenshot showing message option still open

rjwignar:
I think I fixed the color inversion by using a windowWidth of `2.5 * element.scrollWidth` (I played around with this setting and found the inverted portion shrunk as I scaled the `element.scrollWidth` value until it eventually disappeared.

Although, a lot of the text is still pushed-down slightly (most noticeable at the avatar icon):
image

The screenshots don't look the best in light mode either:
message (3)

PR #653, opened 2024-06-10
Visual Layout fixes

menghif:
This PR makes the following changes (they're all small so I thought I would group them together):
- Add missing icons to functions title menu.
- Hide star arrow button when there are no starred prompts.
- Fix divider position in starred prompts list.
- Move header search icon inside search box.
- Show Saved Chats by default when opening the Side Menu.
- Show "Today" or "Yesterday" as date in Side Menu.
- Padding and Button layout improvements.

Chatcraft.org screenshot


humphd (approved):
❤️

PR #654, opened 2024-06-16
Improve mobile prompt

### LINKED ISSUE #336: Change mobile interface to be easier to use.

menghif:
Fixes #336

This PR improves the mobile prompt interface by making space for the Text area (or the "Recording... 00:01" area) in the following ways:
- change padding and button sizes.
- move text-to-speech button inside menu.
- hide microphone button when typing into prompt.
- hide promptSendButton when recording.

| Before | After |
| --- | --- |
| IMG_5170 | IMG_5171 |

https://github.com/tarasglek/chatcraft.org/assets/53121061/65b0ef03-1ece-404c-bc1b-50a57b044bb1

### DIFF
--- a/src/Chat/ChatHeader.tsx
+++ b/src/Chat/ChatHeader.tsx
@@ -101,6 +101,7 @@ function ChatHeader({ chat }: ChatHeaderProps) {
                     bg="white"
                     _dark={{ bg: "gray.700" }}
                     size="sm"
+                    borderRadius={4}
                     fontSize="1rem"
                     w="100%"
                     autoFocus={true}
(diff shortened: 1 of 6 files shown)


tarasglek:
I love all these changes :)

humphd (approved):
So much to love here, well done!

Issue #660, opened 2024-07-04
Model selection is not working


iramykytyn:
I used OpenAI models, and then I switched to OpenRouter models and selected Cloude. I was able to use Cloude for one request, but after that it switched to germini and I was not able to select Cloude anymore. 
https://www.loom.com/share/bb1c14802bb346f1a8a5055c33c2bafd

humphd:
I think the issue here is likely that you have attached images, and we are trying to be "helpful" by switching to a model that we think has vision support.  See https://github.com/tarasglek/chatcraft.org/blob/fa04130e9c37c8945fda40a05594b2ff79bd0add/src/components/PromptForm/MobilePromptForm.tsx#L78 and https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ChatCraftModel.ts#L73.  Our current logic is way to simple, and based on OpenAI vs. OpenRouter.

We should make this better and not force a particular model.

cc @mingming-ma

PR #664, opened 2024-07-21
Fix chatcraft selecting wrong model per issue #660

tarasglek:
Fix chatcraft selecting wrong model per issue #660


tarasglek:
This was driving me insane on mobile

humphd:
I wonder if we should add some UI hints about what a model can do, so people are forced to be LLM experts to use various features...

PR #674, opened 2024-09-02
Fix zIndex bug with message header menus

### LINKED ISSUE #671: z-index bug with message 3-dots and menus
Screenshot 2024-09-01 at 4 38 31 PM

Amnish04:
This PR fixes the `zIndex` bug with messages header menus, where the `menuButton` of other menu places itself over the menu list of the first expanded menu.

**Preview:**
image

**Explanation:**

Since the header options had a `zIndex` of `1` and they were higher up in [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) than the **options menus**, the options menus could never be placed higher with whatever zIndex we gave them were placed within the stacking context of the **header options**.

image

One thing I was wondering was that why would a zIndex value on header options flexbox work as it wasn't a positioned element.
Later I found that elements that are children to a flex container [also form a stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context).
image

image

This fixes #671

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -1,9 +1,7 @@
-import { Box, IconButton, useColorMode } from "@chakra-ui/react";
+import { IconButton, useColorMode } from "@chakra-ui/react";
 import { Menu as ReactMenu, type MenuProps as ReactMenuProps } from "@szhsin/react-menu";
 import React from "react";
 import { TbDots } from "react-icons/tb";
-
-import theme from "../../theme";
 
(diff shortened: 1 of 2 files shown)


humphd (approved):
Love it, thank you Dr. CSS!

tarasglek:
Dr. CSS :)

Amnish04:
Thank you :D

Issue #677, opened 2024-09-07
Don't throw away partial messages on network error


tarasglek:
with non-openai models its somewhat common to get a network error midway through response, in which case chatcraft throws away partial response 😦

PR #678, opened 2024-09-07
debugging issue #677

tarasglek:
I inject a failure after finishing streaming completions, so we can figure out how to handle that.

### DIFF
--- a/src/components/MessagesView.tsx
+++ b/src/components/MessagesView.tsx
@@ -147,7 +147,7 @@ function MessagesView({
 
         {instructions}
 
-        {newMessage && (
+        {isLoading && newMessage && (
           <NewMessage
             message={newMessage}
             chatId={chatId}
--- a/src/hooks/use-chat-openai.ts
+++ b/src/hooks/use-chat-openai.ts
@@ -164,7 +164,7 @@ function useChatOpenAI() {
           return response;
         })
         .finally(() => {
-          setStreamingMessage(undefined);
+          // setStreamingMessage(undefined);
           setPaused(false);
           resetScrollProgress();
           setShouldAutoScroll(false);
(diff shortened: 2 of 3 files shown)


tarasglek:
@humphd this is as far as I can get without your help. I could finish this pull req, but next step requires which hack to pick. I'd probably just add a .catch() handler in front of that finally where i left a comment from you..and add partialMessage to the error...or we could do something else that you might want?

humphd:
I tried today as well, with a different approach, see https://github.com/tarasglek/chatcraft.org/pull/679.  I can't figure out why it won't clear the streaming response; otherwise, it's working.

tarasglek:
@humphd here is another attempt. I kinda like this one

src/hooks/use-chat-openai.ts
```diff
@@ -164,7 +164,7 @@ function useChatOpenAI() {
           return response;
         })
         .finally(() => {
-          setStreamingMessage(undefined);
+          // setStreamingMessage(undefined);
```
humphd, line 167: I don't love this.  Why not capture it in a `.catch()` before it goes away?

src/components/MessagesView.tsx
```diff
@@ -147,7 +147,7 @@ function MessagesView({
 
         {instructions}
 
-        {newMessage && (
+        {isLoading && newMessage && (
```
humphd, line 150: You've already handled the loading/non-loading case in the parent before passing props

src/hooks/use-chat-openai.ts
tarasglek: because there are a lot of catches.

if we modify the code a bit a catch near this finally would become too soon(eg something could throw after it)

keeping streaming message all the way through feels more robust

src/components/MessagesView.tsx
tarasglek: code in this component gets confused if i dont do this cos it uses presence of newMessage as indicator of loading sometimes 😢

PR #679, opened 2024-09-08
Don't throw away incomplete responses from ai

humphd:
Trying to figure out #677 (cc @tarasglek)

I spent some time on this, and I have a solution, but for some reason it doesn't clear the "streaming" state when the error occurs, so the old `newMessage` (i.e., `streamingMessage`) is still being rendered.

There's some state bug here I don't understand.  @Amnish04 do you see what I'm doing wrong?

To test this, ask an LLM anything, and after it streams a few bits of the response it will throw.

### DIFF
--- a/src/Chat/ChatBase.tsx
+++ b/src/Chat/ChatBase.tsx
@@ -39,6 +39,7 @@ import { WebHandler } from "../lib/WebHandler";
 import { ChatCraftCommandRegistry } from "../lib/commands";
 import ChatHeader from "./ChatHeader";
 import PreferencesModal from "../components/Preferences/PreferencesModal";
+import { ChatCompletionError } from "../lib/ai";
 
 type ChatBaseProps = {
   chat: ChatCraftChat;
@@ -356,6 +357,11 @@ function ChatBase({ chat }: ChatBaseProps) {
           forceScroll();
         }
       } catch (err: any) {
+        if (err instanceof ChatCompletionError && err.incompleteResponse) {
+          // Add this partial response to the chat
+          await chat.addMessage(err.incompleteResponse);
+        }
+
         error({
           title: `Response Error`,
           message: err.message,
@@ -367,7 +373,7 @@ function ChatBase({ chat }: ChatBaseProps) {
       }
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [user, chat, setLoading, setShouldAutoScroll, callChatApi, error]
+    [user, chat, streamingMessage, setLoading, setShouldAutoScroll, callChatApi, error]
   );
 
   // Restart auto-scrolling and resume a paused response when Follow Chat is clicked
--- a/src/lib/ai.ts
+++ b/src/lib/ai.ts
@@ -31,10 +31,6 @@ export class ChatCompletionError extends Error {
   constructor(originalError: Error, incompleteResponse?: ChatCraftAiMessage) {
     super(`API Error: ${originalError.message}`);
     this.incompleteResponse = incompleteResponse?.clone();
-    // Indicate that this message is incomplete with ellipses
-    if (this.incompleteResponse) {
-      this.incompleteResponse.text += "...";
-    }
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/lib/ai.ts
```diff
@@ -24,6 +24,20 @@ export type ChatOptions = {
...
+export class ChatCompletionError extends Error {
+  // The partial response that was returned before the error occurred
+  incompleteResponse?: ChatCraftAiMessage;
+
+  constructor(originalError: Error, incompleteResponse?: ChatCraftAiMessage) {
+    super(`API Error: ${originalError.message}`);
+    this.incompleteResponse = incompleteResponse?.clone();
+    // Indicate that this message is incomplete with ellipses
+    if (this.incompleteResponse) {
+      this.incompleteResponse.text += "...";
```
tarasglek, line 36: we should not do this ...

Cos if we ask the model to continue response, it will not be awesome. We can make a <resume> button for after chat fails in style of <pause> functionality we have

src/Chat/ChatBase.tsx
```diff
@@ -356,6 +357,11 @@ function ChatBase({ chat }: ChatBaseProps) {
           forceScroll();
         }
       } catch (err: any) {
+        if (err instanceof ChatCompletionError && err.incompleteResponse) {
+          // Add this partial response to the chat
+          await chat.addMessage(err.incompleteResponse);
```
tarasglek, line 362: ok so you agree with me on this approach

src/lib/ai.ts
```diff
@@ -230,11 +248,16 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
   };
 
   const handleError = async (error: Error) => {
+    const chatCompletionError = new ChatCompletionError(
+      error,
+      buffer.length ? new ChatCraftAiMessage({ model, text: buffer.join("") }) : undefined
```
tarasglek, line 253: pretty sure this is just the incremental buffer for inbetween requestAnimationFrame stuff

Amnish04:
> for some reason it doesn't clear the "streaming" state when the error occurs, so the old `newMessage` (i.e., `streamingMessage`) is still being rendered.

@humphd I tested this rn and for some reason, it works fine for me. The message you add in the catch block is added to the chat, and the streaming message does not show up.
Here's what it looks like after page refresh.

image

humphd:
When I do it, if I scroll down, the message is shown twice.  Might be hooks rendering twice in dev...

Amnish04:
> When I do it, if I scroll down, the message is shown twice. Might be hooks rendering twice in dev...

@humphd I was able to reproduce this error on my windows laptop, probably cuz that is slower than my mac.

image

As per my analysis, this is because of a race condition between the [onData](https://github.com/tarasglek/chatcraft.org/blob/ebd165ed22cd28411cc017646c5a7ec003efb6cd/src/hooks/use-chat-openai.ts#L88) and [requestAnimationFrame](https://github.com/tarasglek/chatcraft.org/blob/ebd165ed22cd28411cc017646c5a7ec003efb6cd/src/lib/ai.ts#L236) callback that we added a few months ago.

Since the `requestAnimationFrame` callback waits for the next repaint, on some devices that might happen after `setStreamingMessage(undefined)` has already been called throwing us back into `streaming` state.

To avoid this, we can check if the streaming message is `undefined` before setting updating it with the latest `currentText`.
Doing that fixed the issue for me. Is it okay if I push that commit in this PR?

humphd:
@Amnish04 sounds good, yes, you can add that here.

Amnish04:
@humphd Just pushed. The only thing that's left is to remove the forced error we are throwing for testing purposes.

tarasglek:
I looked over these changes a few times, asked a few ai models to review them, lgtm. Landing this, we can address anything else in followups. This was a complex issue, impressed with @Amnish04 to finding the root cause

tarasglek:
i'm pretty sure this completely broke rendering of incremental responses..just noticed this with slower llms, we dont get any updates until the end.oops

humphd:
> i'm pretty sure this completely broke rendering of incremental responses..just noticed this with slower llms, we dont get any updates until the end.oops

Yup, can you please review https://github.com/tarasglek/chatcraft.org/issues/681?

PR #682, opened 2024-09-10
Fix regression in streaming messages, cancel rAF on error

### LINKED ISSUE #681: [Regression] Streaming messages no longer show
Since #679 landed, I've noticed that I don't get incremental, streaming responses--only the final message shows.

cc @Amnish04, @tarasglek

humphd:
Fixes #681

I think the fix we did for that rAF race was wrong.  This does it higher-up the stack, keeping track of the current animation frame and canceling it on errors.

This seems to work well for me locally.

### DIFF
--- a/src/hooks/use-chat-openai.ts
+++ b/src/hooks/use-chat-openai.ts
@@ -123,22 +123,15 @@ function useChatOpenAI() {
 
             //#endregion
 
-            // Only set a streaming message, if we are still in the process of streaming it
-            // If the message is `undefined`, that means we have already exited the streaming state
-            // See https://github.com/tarasglek/chatcraft.org/blob/ebd165ed22cd28411cc017646c5a7ec003efb6cd/src/hooks/use-chat-openai.ts#L173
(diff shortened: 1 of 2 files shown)


Amnish04 (commented):
@humphd I found the issue with the previous method.

For some reason, the `streamingMessage` state is always `undefined` in the `onData` callback which is why our streaming Message was never set to something.

Changing it to this fixes the issue for me locally.
```ts
setStreamingMessage((prevValue) => {
  if (prevValue) {
    incrementScrollProgress();
    return new ChatCraftAiMessage({
      id: message.id,
      date: message.date,
      model: message.model,
      text: currentText,
    });
  }
});
```

But I just don't understand why react state is always set to its initial value when accessing inside callbacks using global scope.

This gave me a lot of trouble when working on evals dashboard as well, which is why I had to maintain corresponding refs for lots of states so I could use their value inside grid callbacks.

Do you know why this happens?

tarasglek:
tested, seems to work fine with errors and without

PR #695, opened 2024-10-04
add logic to track credit left

AnhChienVu:
Hi, I implemented the logic to track the credit left. However, I couldn't integrate to the click send button. 
I only made changes in use-cost.tsx file, it's a hook so you can use it at any time.
You need to set the API to be able fectch the data from OpenRouter:
```
const response = await fetch("https://openrouter.ai/api/v1/auth/key", {
        headers: {
          Authorization: `Bearer APIKey`,
        },
      });
```

It's not a fully integrated feature so just let me know any hints that I can enhance it.

### DIFF
--- a/src/hooks/use-cost.tsx
+++ b/src/hooks/use-cost.tsx
@@ -1,14 +1,28 @@
-import { useCallback, createContext, useContext, type ReactNode, type FC } from "react";
+import {
+  useCallback,
+  createContext,
+  useContext,
+  useState,
+  useEffect,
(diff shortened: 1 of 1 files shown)


humphd:
This can only work with the OpenRouter provider, so you'll need to deal with the user not having it selected as their current provider.  You'll also need to get the API key from the settings, there's lots of existing code that does this in the tree for you to use.

Finally, we'll have to think about how this would get hooked into the UI.

PR #700, opened 2024-10-27
fix various model hardcoding. Should move to model features

tarasglek:
Eg we should add functionality to identify model capabilities:
- will it accept image input
- is it speech to text
- is it text to speech

Then we can filter models better and help users decide what to use

### DIFF
--- a/src/components/Preferences/ModelsSettings.tsx
+++ b/src/components/Preferences/ModelsSettings.tsx
@@ -39,7 +39,7 @@ import { useModels } from "../../hooks/use-models";
 import { useSettings } from "../../hooks/use-settings";
 import { ChatCraftModel } from "../../lib/ChatCraftModel";
 import { ChatCraftProvider, ProviderData } from "../../lib/ChatCraftProvider";
-import { textToSpeech } from "../../lib/ai";
+import { isTextToSpeechModel, textToSpeech } from "../../lib/ai";
 import db from "../../lib/db";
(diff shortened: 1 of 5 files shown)


tarasglek:
@humphd i ran prettier, but vs code still do a bunch of header "cleanups" in this diff. wdyt re this change?

tarasglek:
We can do this for vlms too

humphd (approved):
Looks good, works great

humphd:
@Amnish04 this touches some of your code, in case you want to look too. I think the changes are fine.

src/components/Preferences/ModelsSettings.tsx
```diff
@@ -500,8 +500,9 @@ function ModelsSettings(isOpen: ModelsSettingsProps) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isOpen]);
 
+  // we have multiple isTtsSupported
   const isTtsSupported = useMemo(() => {
-    return !!models.filter((model) => model.id.includes("tts"))?.length;
+    return !!models.find((model) => isTextToSpeechModel(model.id));
   }, [models]);
```
Amnish04, line 506: Yes, this needs to be moved inside a hook. Filed #703 for this.

Amnish04 (approved):
This looks pretty good! Thanks for cleaning up parts of my code.

PR #702, opened 2024-10-28
Only show one error toast at a time

Amnish04:
@tarasglek wanted to only show the last error on screen to avoid confusion with lots of stacked error toasts.

This quick PR does exactly that by keeping track of any open error toast ids, and closing them whenever a new error toast is triggered.

Here's a quick **demo** of how it looks.
OnlyLastErroraToast

It wasn't clear to me if this was exactly what we wanted, or hide any open toast on submitting a completion, so I went with whatever I understood.

One problem I see with this approach is that some important error messages might be lost if multiple relevant error toast need are relevant at a time.

### DIFF
--- a/src/hooks/use-alert.tsx
+++ b/src/hooks/use-alert.tsx
@@ -21,6 +21,10 @@ function truncateMessage(message?: string): string {
 
   return message;
 }
+
+// Keep track of open error toasts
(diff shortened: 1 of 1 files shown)


humphd (approved):

tarasglek:
love it. thank you.

Issue #655, opened 2024-06-17
Add safe-bottom padding on mobile iOS


humphd:
The new changes just landed in #654 are amazing.  I was trying them today on my iPhone, and I notice that we could use a little more safe-padding on the bottom, see https://webkit.org/blog/7929/designing-websites-for-iphone-x/:

![Screenshot 2024-06-17 at 5 30 32 PM](https://github.com/tarasglek/chatcraft.org/assets/427398/95504bfb-b663-4da0-a941-5a6c181ee739)

gitdevjin:
Hello @humphd.
I am interested in this issue and was trying to fix it.
On the inspect page of Chrome, it seems to need extra padding on the bottom.

On inpsect:
<img width="1201" alt="Screenshot 2024-10-25 at 4 46 22 PM" src="https://github.com/user-attachments/assets/90cb8c4f-bf11-4c79-9eb4-3db11372bda6">

However, on my phone(iPhone pro 15) using different browsers, Safari, Chrome, [Naver](https://apps.apple.com/us/app/%EB%84%A4%EC%9D%B4%EB%B2%84-naver/id393499958)
I looked fine due to the navbar on the bottm.

### Chrome:
![IMG_0722](https://github.com/user-attachments/assets/3ddfb4cd-39bb-4705-afc3-a9a8b8282eae)

### Safari
![IMG_0723](https://github.com/user-attachments/assets/22a858fa-15b2-4212-89f7-22da134eacc0)

### Naver:
![IMG_0724](https://github.com/user-attachments/assets/7ad4c810-8698-4738-802d-0416d5891941)

Could you let me know what mobile browser you used or how you entered the chatcraft page?

tarasglek:
You need to add it to home screen via share function first ![IMG_EBAF7359206F-1](https://github.com/user-attachments/assets/439351ad-72cf-4ae6-b242-c75e7b30573d)

humphd:
Fixed in #705

PR #705, opened 2024-10-29
`safe-area-inset-bottom` padding added to MobilePromptForm

gitdevjin:
# Description
This PR handles [Issue-655](https://github.com/tarasglek/chatcraft.org/issues/655)
It added saft-bottom padding for Mobile Devices.

# How I tested
I tried to test it with xcode simulator; however, it didn't hide the bottom navbar even though I added it to homscreen as @tarasglek told me. I also tried to access it with my phone to the locally-run chatcraft, but failed.
As a result, I tested it with using [Mobile simulator - responsive testing tool](https://www.webmobilefirst.com/en/).

# The screenshots of test for different devices.
## iPhone 13 Pro
iPhon13Pro

## Galaxy Z Flip 3
GalaxyZ-Flip3

## Galaxy Ultra 21
GalaxyUltra21

## Google Pixel 6 Pro
GooglePixel6Pro

### DIFF
--- a/src/components/PromptForm/MobilePromptForm.tsx
+++ b/src/components/PromptForm/MobilePromptForm.tsx
@@ -150,7 +150,15 @@ function MobilePromptForm({
   };
 
   return (
-    <Box flex={1} w="100%" h="100%" my={1} px={2} py={1}>
+    <Box
+      flex={1}
+      w="100%"
+      h="100%"
+      my={1}
+      px={2}
+      py={1}
+      pb="calc(env(safe-area-inset-bottom) + 15px)"
+    >
       <chakra.form onSubmit={handlePromptSubmit} h="100%">
         <Flex alignItems="end" gap={2}>
           <OptionsButton


tarasglek:
Please try it on local device.

For future reference:
You can do this via
Install [cloudflared](https://github.com/cloudflare/cloudflared) then do
cloudflared --url http://localhost:5173/ (orwhatever url pnpm dev is listening), that gives you a temp url you can use on any device.

I'll also approve it for ci, you can try the pull req preview

gitdevjin:
@tarasglek, Thank you for the tip for how to test it with my local device! It's great!

I found something interesting that happened while I was trying to test it, and did some experiments

*****
#### The following experiments are done on MacBook using xcode simulator. I did the same experiment using iPhone(via cloudflared), and what happened are exactly the same.
*****

The locally-run `Chatcraft`(I will call it `Local Chatcraft`) acts differently than the deployed `Chatcraft`(I will call it `Real Chatcraft`). 
When I add `Local Chatcraft`to `Home Screen`, it looks like the following screenshot.

## Adding `Local Chatcraft` to Home Screen 
Screenshot 2024-10-30 at 9 09 12 PM

The address includes `chat id`. It looks like it's added just as a bookmark.


However, when I run `Real Chatcraft`, it doens't include `chat id`
## Adding `Real Chatcraft` to Home Screen
Screenshot 2024-10-30 at 9 09 42 PM

And also, the default name is different. It's `New Chat - ChatCraft` for `Local Chatcraft`; On the other hand,`Chatcraft` for `Real Chatcraft`. 

## When I open the icon added from `Local Chatcraft`, it runs on `Safari`(default browser).
Screenshot 2024-10-30 at 9 09 58 PM

## When I open the icon added frin `Real Chatcraft`, it looks like it's running independently.

Screenshot 2024-10-30 at 9 10 33 PM

## Comparison Screenshot
Screenshot 2024-10-30 at 9 10 55 PM

As a result, I couldn't test it the way you wanted me to. I am sorry.

tarasglek:
we can try landing it and see what happens, what do you think?

menghif:
This is the right solution but you also need to change the viewport to use `viewport-fit=cover` (like it says [here](https://webkit.org/blog/7929/designing-websites-for-iphone-x/)):

```
<meta name='viewport' content='initial-scale=1, viewport-fit=cover'>
```

To test this, you need to do `pnpm build` and then `pnpm preview -host`.

Go to the network URL on an iPhone, then press the "Share" icon and "Add to Home Screen". 

| Safari | Web App (with no URL bar) |
|--------|--------|
| IMG_6386 | IMG_6384 | 

I think you could also decrease the bottom padding from 15px to something like 10px.

menghif (changes requested):
This is the right solution but you also need to change the viewport to use `viewport-fit=cover` (like it says [here](https://webkit.org/blog/7929/designing-websites-for-iphone-x/)):

```
<meta name='viewport' content='initial-scale=1, viewport-fit=cover'>
```

To test this, you need to do `pnpm build` and then `pnpm preview -host`.

Go to the network URL on an iPhone, then press the "Share" icon and "Add to Home Screen". 

| Safari | Web App (with no URL bar) |
|--------|--------|
| IMG_6386 | IMG_6384 | 

I think you could also decrease the bottom padding from 15px to something like 10px.

gitdevjin:
@menghif  Thank you so much! The testing method you showed me worked perfectly. I tested it on an iPhone 15 Pro, and it looks good without any additional padding—I was able to use env(safe-area-inset-bottom) directly. Thanks again for your help!
IMG_0728

menghif (approved):

menghif:
I'm glad you got it working!

humphd:
I'v been using this since last night on my iPhone, and it's amazing.  Thank you for the fix @gitdevjin and review @menghif!

PR #707, opened 2024-11-02
Try to improve perf

tarasglek:
Step 1 in profiler:
image

With that I put breakpoint on hook 1 made that update stop happening..(disable scrolling commit).

However hook 1:
image
Keeps triggering an update of the whole screen
image

This makes no sense to me...first of all why is it saying that dark mode and gridtemplatecolumns changed..wtf

### DIFF
--- a/src/hooks/use-chat-openai.ts
+++ b/src/hooks/use-chat-openai.ts
@@ -131,7 +131,7 @@ function useChatOpenAI() {
                 text: currentText,
               })
             );
-            incrementScrollProgress();
+            // incrementScrollProgress();
           }
         },
       });


tarasglek:
This makes me wonder if we should have separate state management for incremental message rendering..eg so only the message listens for completion updates....but that seems like we are then side-stepping and not understanding how react works

tarasglek:
> This makes no sense to me...first of all why is it saying that dark mode and gridtemplatecolumns changed..wtf

So basically in react one has to subscribe to state as close as possible to component being rendered... it will recompute everything without useMemo

PR #708, opened 2024-11-02
input lag fix

tarasglek:
Stopped keeping prompt in react. @humphd how's this change?

### DIFF
--- a/src/components/PromptForm/DesktopPromptForm.tsx
+++ b/src/components/PromptForm/DesktopPromptForm.tsx
@@ -77,7 +77,6 @@ function DesktopPromptForm({
 }: DesktopPromptFormProps) {
   const [prompt, setPrompt] = useState("");
   // Has the user started typing?
-  const [isDirty, setIsDirty] = useState(false);
   const { error } = useAlert();
   const { settings } = useSettings();
   const [isRecording, setIsRecording] = useState(false);
@@ -91,12 +90,6 @@ function DesktopPromptForm({
   const [selectedImageUrl, setSelectedImageUrl] = useState<string>("");
   const location = useLocation();
 
-  // If the user clears the prompt, allow up-arrow again
-  useEffect(() => {
-    if (!prompt) {
-      setIsDirty(false);
-    }
-  }, [prompt, setIsDirty]);
 
   // Focus the prompt form when the user navigates
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


tarasglek:
Not super excited to duplicate this fix for mobile..would be nice to unify the logic and just keep rendering separate

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -164,27 +161,25 @@ function DesktopPromptForm({
     switch (e.key) {
       // Allow the user to cursor-up to repeat last prompt
       case "ArrowUp":
-        if (!isDirty && previousMessage) {
+        if (isPromptEmpty && previousMessage && inputPromptRef.current) {
```
humphd, line 164: Could you get rid of the `isPromptEmpty` state completely and do:

```ts
if (inputPromptRef.current?.value === "" && previousMessage) {
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -422,7 +418,7 @@ function DesktopPromptForm({
                 />
 
                 <Flex alignItems="center" gap={2}>
-                  <KeyboardHint isVisible={!!prompt.length && !isLoading} />
+                  <KeyboardHint isVisible={!isPromptEmpty && !isLoading} />
```
humphd, line 421: I'm not sure if doing `inputPromptRef.current.value !== ""` works here, but it might

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -164,27 +161,25 @@ function DesktopPromptForm({
...
         break;
 
       // Prevent blank submissions and allow for multiline input.
       case "Enter":
-        // Deal with Enter key based on user preference and state of prompt form
         if (settings.enterBehaviour === "newline") {
           handleMetaEnter(e);
         } else if (settings.enterBehaviour === "send") {
-          if (!e.shiftKey && prompt.length) {
+          if (!e.shiftKey && !isPromptEmpty) {
```
humphd, line 176: Same here

humphd (commented):
Seems to work, not sure what it loses, if anything.

I wonder if we can further drop state here and compare to values whenever necessary.

src/components/PromptForm/DesktopPromptForm.tsx
tarasglek: doesnt cos you need to trigger a react rerender

src/components/PromptForm/DesktopPromptForm.tsx
tarasglek: here either would work, this is just more consistent cos it does trim in state

PR #709, opened 2024-11-03
fix input overhead on mobile

tarasglek:
Basically ripped out all kb logic from mobile

### DIFF
--- a/src/components/PromptForm/MobilePromptForm.tsx
+++ b/src/components/PromptForm/MobilePromptForm.tsx
@@ -1,4 +1,4 @@
-import { FormEvent, KeyboardEvent, type RefObject, useEffect, useState } from "react";
+import { FormEvent, type RefObject, useEffect, useState } from "react";
 import { Box, chakra, CloseButton, Flex, Image, Spinner } from "@chakra-ui/react";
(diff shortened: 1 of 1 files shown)


src/Chat/ChatBase.tsx
```diff
@@ -384,7 +392,8 @@ function ChatBase({ chat }: ChatBaseProps) {
 
   const defaultProviderAlert = useMemo(() => {
     // If we are using default provider, show alert banner to notify user
-    if (showAlert) {
+    // add impossible condition so lint doesn't complete re `React Hook useMemo has an unnecessary dependency: 'settings.currentProvider'.`
```
humphd, line 395: Let's not do this and add a lint ignore comment

humphd (changes requested):
Seems to work well for me in my limited testing.

@Amnish04, any issues you see here?

src/Chat/ChatBase.tsx
tarasglek: ok dropped it for now

Amnish04 (commented):
@humphd @tarasglek This looks good to me! The only catch is that people who connect keyboard with mobile would no longer be able to use the shortcuts. But that's not too common, so ig better to remove this in favour of performance.

PR #710, opened 2024-11-03
Don't block sending chat when clicking Ask button

humphd:
I've removed the extra logic that was blocking the chat from being submitted to the LLM when you click the Ask button.  It's annoying how it just doesn't do anything when you click "Ask" without telling you what's wrong.


tarasglek:
lovely when simplification of ux and code come together like this

PR #714, opened 2024-11-06
Fixing an overflow in Web handler body

### LINKED ISSUE #613: Overflow in Web Handlers settings body
In the new User Settings Modal, the Web Handlers body has an overflow.

Screenshot of User Settings with overflow

gitdevjin:
# Description
This Fixes #613 

`<Flex/>` seems to need `width` specified to prevent from overflowing its parent. I added a width to it.

# How I tested
I Tested it with different screen size on a desktop(google chrome) using inspect page and also on a mobie device (iPhone pro 15)

In the video, when I try to move the mouse to check for overflow, it may seem stuck or unresponsive. However, this is just because I'm trying to move the mouse beyond the visible area to check if there is any overflow.

## Test Video
[Test Video](https://youtu.be/CBqwkEaSLTI)


menghif (approved):
Nice fix!

PR #715, opened 2024-11-08
Allow Attaching/Drag-and-Dropping more file types

humphd:
This updates our "Attach Files..." menu option and Drag-and-Drop to support the following file types:

- images: works same as before
- PDF files: converts to markdown with jina.ai, then creates a new message with the contents
- text files: tries to syntax highlight when type is known and creates a new message with the contents
- Word DOCX: converts to HTML then to Markdown (the last bit isn't working yet)

You can attach or drag-and-drop a bunch of files and it will process them all.

A few things could be improved, which I ran out of time to work on (can happen in follow ups, or others can push fixes):

1. I want proper CSS to happen when the user drags over the prompt form so you know it's a drop target.  I tried to do something with a green border, but it's not working.
2. ~There is a bunch of code duplication here that could be tamed~
3. I want to be able to add a jina.ai provider key in the settings modal so you can have LLM Providers and "Document Providers" (e.g., services that allow you to process file types into text for LLMs).  This isn't critical, since it works now.

@menghif, @Amnish04: can you see what I'm doing wrong with my CSS on the drop target?

### DIFF
--- a/src/components/PromptForm/DesktopPromptForm.tsx
+++ b/src/components/PromptForm/DesktopPromptForm.tsx
@@ -1,12 +1,4 @@
-import {
-  FormEvent,
-  KeyboardEvent,
-  type RefObject,
-  useCallback,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import { FormEvent, KeyboardEvent, type RefObject, useEffect, useMemo, useState } from "react";
 import {
   Box,
   Card,
@@ -37,9 +29,7 @@ import { useKeyDownHandler } from "../../hooks/use-key-down-handler";
 import { useAlert } from "../../hooks/use-alert";
 import ImageModal from "../ImageModal";
 import { ChatCraftChat } from "../../lib/ChatCraftChat";
-import { ChatCraftHumanMessage } from "../../lib/ChatCraftMessage";
-import { JinjaReaderResponse } from "../../lib/ai";
-import { importFiles } from "../../lib/file";
+import { useFileImport } from "../../hooks/use-file-import";
 
 type KeyboardHintProps = {
   isVisible: boolean;
@@ -89,7 +79,7 @@ function DesktopPromptForm({
   previousMessage,
 }: DesktopPromptFormProps) {
   const [isPromptEmpty, setIsPromptEmpty] = useState(true);
-  const { error, progress, closeToast } = useAlert();
+  const { error } = useAlert();
   const { settings } = useSettings();
   const [isRecording, setIsRecording] = useState(false);
   const [isTranscribing, setIsTranscribing] = useState(false);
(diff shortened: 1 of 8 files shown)


humphd:
It would be cool to add DOCX support next by passing the file through https://github.com/mwilliamson/mammoth.js/ and then sending the HTML to jina.ai to get Markdown.

humphd:
Refactored to remove code duplication, simplified things, added support for Word DOCX (I can't get Jina.ai to do the right thing when I POST an HTML file to it, so for now I'm leaving the Word DOCX in HTML format, which still works).

humphd:
I'd say this is good to land and we can do follow-ups.  Anyone want to test before I do?

src/lib/ai.ts
```diff
@@ -510,3 +510,109 @@ export function isChatModel(model: string): boolean {
...
+  return convertToMarkdownWithJina(file, "pdf");
+}
+
+/**
+ * Parse an HTML to Markdown using https://jina.ai/reader/
+ * @param file an HTML file to parse
+ */
+export async function htmlToMarkdown(html: File | string): Promise<JinjaReaderResponse> {
+  // TODO: jina.ai is not returning the expected result for HTML, not sure why...
+  return convertToMarkdownWithJina(html, "html");
```
Amnish04, line 617: Is this because you return the input content as is when `file` is a `string` and `type` is `html`?

https://github.com/tarasglek/chatcraft.org/blob/c454e077b4e5e68cf2229c3e1a8c81f2aa231802/src/lib/ai.ts#L566-L568

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -291,19 +304,23 @@ function DesktopPromptForm({
...
       <Card flex={1} my={3} mx={1}>
         <chakra.form onSubmit={handlePromptSubmit} h="100%">
-          <CardBody h="100%" px={6} py={4}>
+          <CardBody
+            h="100%"
+            px={6}
+            py={4}
+            transition="background-color 0.2 ease"
+            _hover={{ borderColor: activeBorder }}
+            borderColor={isDragActive ? activeBorder : "inherit"}
```
Amnish04, line 319:
```suggestion
            border={"2px solid"}
            borderColor={isDragActive ? activeBorder : "inherit"}
```

The reason is there is no border here to start with :)

image

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -291,19 +304,23 @@ function DesktopPromptForm({
     // Otherwise, let the default paste handling happen
   };
 
+  const activeBorder = useColorModeValue("green.100", "green.600");
```
Amnish04, line 307:
```suggestion
  const activeBorder = useColorModeValue("green.200", "green.600");
```

I would change the lighter version to `green.200`. Current shade is too light.

**Before:**
image

**After:**
image

src/hooks/use-file-import.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const settings = getSettings();
+
+  const importFile = useCallback(
+    (file: File, contents: string | JinjaReaderResponse) => {
+      if (file.type.startsWith("image/")) {
+        const base64 = contents as string;
+        onImageImport(base64);
+      } else if (file.type === "application/pdf") {
+        const document = (contents as JinjaReaderResponse).data;
+        chat.addMessage(new ChatCraftHumanMessage({ text: `${document.content}\n` }));
```
Amnish04, line 182: 1. While testing, I ran into a case where my pdf content was not text and jina returned the following content:
```
{
  "text": "\n"
}
```

leading to an empty message like this:
image

Should we have an additional check here so we only add a user message when we have some `content.trim()`?

2. Also, should we be paying attention the chat's size by rejecting large files and showing error alerts, so we don't exceed model context lengths? Could probably be done in a follow up if we consider this an issue.

Amnish04 (commented):
@humphd Looks and works great!

Left some comments on why CSS and html to markdown might not be working.

src/lib/ai.ts
```diff
@@ -510,3 +510,109 @@ export function isChatModel(model: string): boolean {
     !(isTextToSpeechModel(model) || isSpeechToTextModel(model) || isTextToImageModel(model))
   );
 }
+
+export type JinjaReaderResponse = {
```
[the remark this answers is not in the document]
humphd: You can't believe how many times I've made this mistake.  For whatever reason, my brain can't do jina vs. jinja...

src/lib/ai.ts
humphd: No, I want that behaviour: I only want to read the file into a base64 string if it's a PDF, but if it's already parsed HTML content, I want that bare string.

However, the bug seems to be with Jina.ai requiring a URL when you post your local HTML file, which makes no sense to me.  Even when I add it, it seems to only return the first line of text.

I think I'll give up on this bit and rip it out.

src/components/PromptForm/DesktopPromptForm.tsx
humphd: Thanks, I've done a version of this.  I still think it could be made better than what I can do in a follow-up, but now it works.

src/hooks/use-file-import.tsx
humphd: Added an info warning when a file is empty instead of importing.

For now, I think we should just try to import all files and let the browser/LLM complain if things are too large.

PR #726, opened 2024-11-11
Support TTS and STT beyond selected provider

### LINKED ISSUE #721: Allow TTS/STT in more situations
If I have an OpenAI key, but I'm using some other provider/model that doesn't TTS, I'd still like to be able to use Whisper with my OpenAI key separate to the LLM I'm using.

cc @Amnish04

Amnish04:
This PR makes **Text to Speech** and **Speech to Text** functionalities independent of the current LLM provider being used for chat completions.

Here's a brief overview of changes:
1. We now cache all the providers and their models in `useModels` hook so we can query any provider's model info without an API request. This was necessary since we need to determine if we support TTS very frequently, at a rate that is greater than 1 models query per TTS chunk request.
2. Created a new hook called `useTextToSpeech` and moved the logic to determine TTS support, and generating audio from text into it, all based on the `providersWithModels` info cached in `useModels`.
3. The `textToSpeech` function now uses an OpenAI client [configured ](https://github.com/tarasglek/chatcraft.org/compare/amnish04/extend-tts-stt-availability?expand=1#diff-1c4a0de27d11b095e2cdada2a3f887c85668ebc3b88e98abce347ca9c69a181fR24)with one of the available providers that support TTS. The idea is to use the `currentProvider` if it supports TTS, else find any other providers if they do.
4. Speech to Text logic also follows a similar logic to create an OpenAI client which is now passed into the `SpeechRecognition` instance vs hard coding the client information in its `trancscribe` method.
5. I feel like `isSpeechToTextSupported` and `getSpeechToTextClient` belong in a separate hook outside `useModels`, but not sure if its necessary.

I don't think this is the best approach, but it worked out of the many I tried.

Here's a **working demo**:
https://www.loom.com/share/67024f7807254bd6a8ba89fb3e985133?sid=238e6ee8-fcb4-4b34-9b32-14464be3c7af

This closes #721.

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -128,6 +128,8 @@ function MessageBase({
   const [imageModalOpen, setImageModalOpen] = useState<boolean>(false);
   const [selectedImage, setSelectedImage] = useState<string>("");
   const { isOpen, onToggle: originalOnToggle } = useDisclosure();
+  const { clearAudioQueue, addToAudioQueue } = useAudioPlayer();
+  const { isTextToSpeechSupported, textToSpeech } = useTextToSpeech();
   const isSystemMessage = message instanceof ChatCraftSystemMessage;
   const isLongMessage =
     text.length > 5000 || (isSystemMessage && summaryText && text.length > summaryText.length);
@@ -344,6 +346,7 @@ function MessageBase({
     progress,
     settings.currentProvider.name,
     settings.textToSpeech.voice,
+    textToSpeech,
   ]);
 
   const handleClick = useCallback((e: MouseEvent<HTMLButtonElement>) => {
@@ -429,9 +432,6 @@ function MessageBase({
   };
   const closeModal = () => setImageModalOpen(false);
 
-  const { clearAudioQueue, addToAudioQueue } = useAudioPlayer();
-  const { isTextToSpeechSupported, textToSpeech } = useTextToSpeech();
-
   const handleSpeakMessage = useCallback(
     async (messageContent: string) => {
       try {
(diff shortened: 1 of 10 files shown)


tarasglek:
So if I'm reading the code right, the logic is "see if current provider has an STT/TTS model, otherwise fallback". I like this. Will test.

humphd (approved):
Code looks good.  I'll let @tarasglek test it functionally.

Amnish04:
> So if I'm reading the code right, the logic is "see if current provider has an STT/TTS model, otherwise fallback". I like this.

Exactly!

tarasglek:
Cross-provider algo does not seem to work.

I tried this with free provider, ...which has a stt model and it keeps using openai according to network requests in devtools

Amnish04:
> Cross-provider algo does not seem to work.
> 
> I tried voice to text with free provider, ...which has a stt model and it keeps using openai according to network requests in devtools

Hmm,  I guess that's because we only cache the provider info that is in localstorage and free providers are not part of that?

Amnish04:
@tarasglek Thanks for finding the issue, just pushed a fix.

Here's a test with a custom provider selected:
image

tarasglek:
My expectation is that when I have a provider selected that offers whisper, it would prefer whisper from that provider. After most recent push it always defaults to free STT, eg the behavior flipped.

Amnish04:
> My expectation is that when I have a provider selected that offers whisper, it would prefer whisper from that provider. After most recent push it always defaults to free STT, eg the behavior flipped.

@tarasglek This should be fixed by recent changes, where we now include default supported providers in models cache and select the STT model based on STT client's provider.

**With Free Provider:**
image

**With OpenAI:**
image

tarasglek:
works

PR #729, opened 2024-11-15
Added ModelProviderMenu component and reused it in PromptsendButton and MessageBase

hpatel292-seneca:
**This fixes #643.**

Added a reusable component, `ModelProviderMenu`, and replaced the logic in `PromptSendButton` and `MessageBase` to utilize this component.

### Overview of Changes:
1. **Added a new component:**  
   `ModelProviderMenu` contains the logic for displaying models and providers, allowing the user to change models and providers directly from it.
   
2. **Updated `PromptSendButton`:**  
   Replaced the existing menu logic in `PromptSendButton` with the `ModelProviderMenu` component. This menu appears next to the send button.

3. **Updated `MessageBase`:**  
   Refactored the retry logic in `MessageBase` to use the `ModelProviderMenu` component, ensuring consistency and reusability.

### Demo

https://github.com/user-attachments/assets/b55bf561-08c3-4d74-8958-1f7b3c6ac559

PR #730, opened 2024-11-15
Added ModelProviderMenu component and reused it in PromptsendButton and MessageBase

hpatel292-seneca:
This is Requested from  #729


humphd:
Can we close #729?

src/components/Menu/ModelProviderMenu.tsx
```diff
@@ -0,0 +1,185 @@
...
+import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
+
+interface ModelProviderMenuProps {
+  label?: string;
+  onItemSelect: (modelId: string) => void;
+  openOnHover?: boolean;
+  menuButtonLabel?: ReactNode;
+}
+
+const ModelProviderMenu: React.FC<ModelProviderMenuProps> = ({
```
Amnish04, line 35: I would name this `ModelSelectionMenu`. @humphd What do you think?

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -229,48 +230,21 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
...
-
   return (
     <ButtonGroup isAttached>
-      <Button type="submit" size="sm" isLoading={isLoading} loadingText="Sending">
+      <Button
+        type="submit"
+        size="sm"
+        isLoading={isLoading}
+        loadingText="Sending"
+        borderRightRadius="0"
```
Amnish04, line 246: We shouldn't need this inside a `ButtonGroup`. The problem you're facing is because you have an invalid `div` inside ButtonGroup that breaks the following style:
image

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -308,89 +283,16 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
...
-      </Menu>
+      {
+        <Box>
+          <ModelProviderMenu
+            onItemSelect={(modelId) => {
+              const model = models.find((m) => m.id === modelId);
+              if (model) setSettings({ ...settings, model });
+            }}
+          />
+        </Box>
```
Amnish04, line 294:
```suggestion
          <ModelProviderMenu
            onItemSelect={(modelId) => {
              const model = models.find((m) => m.id === modelId);
              if (model) setSettings({ ...settings, model });
            }}
          />
```

Get rid of this `Box` to solve border issues.

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -282,6 +256,7 @@ function DesktopPromptSendButton({ isLoading }: PromptSendButtonProps) {
           }
         >
           <Button
+            borderRightRadius="0"
```
Amnish04, line 259:
```suggestion
```

src/components/PromptForm/PromptSendButton.tsx
```diff
@@ -59,7 +60,7 @@ function MobilePromptSendButton({ isLoading }: PromptSendButtonProps) {
 
   return (
     <ButtonGroup variant="outline" isAttached>
-      <Menu placement="top" strategy="fixed" closeOnSelect={false} offset={[-90, 0]}>
+      <Menu placement="top" strategy="absolute" closeOnSelect={false} offset={[-90, 0]}>
```
Amnish04, line 63: Why change this?

src/components/Menu/ModelProviderMenu.tsx
Amnish04, line 35: I would name this `ModelSelectionMenu`. @humphd What do you think?

Amnish04 (changes requested):
@hpatel292-seneca I've left some feedback. We need to replace this Menu with our custom implementation that supports SubMenus so you don't have to use an ad-hoc approach.

tarasglek:
Thanks for moving the pull req over. I agree with Amnish re reusing code more. I suspect that might fix the only bug I found..there on first load if I click retry-with and search for "3b", the menu closes itself....on subsequent uses it seems ok

src/components/PromptForm/PromptSendButton.tsx
hpatel292-seneca: Initially, I changed this because previously I was using `60vh` for `MenuList` and it was overflowing in mobile devices to I tried using `absolute` if that can fix my issue but I didn't so I change the height to `50vh` and it was working fine but I forget to change it back to `fixed`.

src/components/PromptForm/PromptSendButton.tsx
hpatel292-seneca: Understood.

src/components/Menu/ModelProviderMenu.tsx
```diff
@@ -0,0 +1,185 @@
...
+          cursor="pointer"
+          color="inherit"
+          padding="0"
+          background="none"
+          border="none"
+          aria-label={label || "Choose Model"}
+          title={label || "Choose Model"}
+          onClick={adjustedOpenOnHover ? undefined : onOpen} // Open on click if not openOnHover
+          onMouseEnter={adjustedOpenOnHover ? onOpen : undefined} // Open on hover if openOnHover
+          onMouseLeave={adjustedOpenOnHover ? onClose : undefined} // Close on hover out if openOnHover
```
Amnish04, line 88: We shouldn't need to do this ourselves once we migrate to using React-Menu.

src/components/Menu/ModelProviderMenu.tsx
```diff
@@ -0,0 +1,185 @@
...
+          fontSize="1.25rem"
+          aria-label={label || "Choose Model"}
+          title={label || "Choose Model"}
+          icon={<TbChevronUp />}
+          onClick={openOnHover ? undefined : onOpen} // Open on click if not openOnHover
+          onMouseEnter={openOnHover ? onOpen : undefined} // Open on hover if openOnHover
+          onMouseLeave={openOnHover ? onClose : undefined} // Close on hover out if openOnHover
+          borderLeftRadius="0"
+        />
+      )}
```
Amnish04, line 105: @hpatel292-seneca Right now, this implementation is too specific to be reused anywhere as you'll have to add another `MenuButton` **inside** the component implementation wherever this needs to be used.

There are 2 ways we can solve this:
1. Either accept a `variant: "menu" | "submenu"` and a `MenuButton` in the props so you can dynamically provide the host element for menu. But this will be too dirty, and not much better than the current approach.
2. I would say a **better way** for this would be to just extract the `MenuList` part as a separate component so you could reuse it anywhere - wrap it in a `Menu` with whatever `MenuButton` or throw it inside a `SubMenu`. The component could be renamed to `ModelSelectionMenuList`. Just make sure you also import the `MenuGroup` from the library we are wrapping, i.e. [@szhsin/react-menu](https://szhsin.github.io/react-menu/docs#menu-group)

@humphd What are your thoughts on this?

Amnish04 (changes requested):
@hpatel292-seneca Just wrote some implementation ideas.

src/components/Menu/ModelProviderMenu.tsx
humphd: Yeah, your "better way" sounds good

tarasglek:
> Thanks for moving the pull req over. I agree with Amnish re reusing code more. I suspect that might fix the only bug I found..there on first load if I click retry-with and search for "3b", the menu closes itself....on subsequent uses it seems ok

Sweet, functionality-wise this is perfect!

hpatel292-seneca:
Thanks @tarasglek, I found few bugs on mobile view while working it. I will open issues for those and if you agree to make those changes, I would to work on those.

Amnish04 (approved):
@hpatel292-seneca This works well and code looks good to me. Thanks for following up!

humphd:
@hpatel292-seneca can you "Resolve" the issues above that are now dealt with?  It's hard to follow the current state of this PR otherwise.

Let me know when this is ready to land.

hpatel292-seneca:
Hi @humphd,

I can send a PR for the above issue if needed. However, if we merge the PR for the above issue first, this PR will likely encounter merge conflicts, as it also modifies the same file. Let me know how you'd like to proceed.

humphd:
I want to get this merged first then you can improve things in follow-ups.  Please resolve whatever is done above, and file a new issue with the things to be fixed next, then resolve, so we can land this.

src/components/Menu/ModelProviderMenu.tsx
```diff
@@ -0,0 +1,185 @@
...
+  InputLeftElement,
+  Menu,
+  MenuButton,
+  MenuDivider,
+  MenuGroup,
+  MenuItem,
+  MenuList,
+  IconButton,
+  useDisclosure,
+} from "@chakra-ui/react";
```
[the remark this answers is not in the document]
hpatel292-seneca: This is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/8e706c19f003c517e2bcc61b0293f65b449b416d

src/components/Menu/ModelProviderMenu.tsx
hpatel292-seneca: I followed your suggested in below comments and named the new component `[ModelSelectionMenuList]` in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/8e706c19f003c517e2bcc61b0293f65b449b416d

src/components/PromptForm/PromptSendButton.tsx
hpatel292-seneca: This is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/8e706c19f003c517e2bcc61b0293f65b449b416d

src/components/PromptForm/PromptSendButton.tsx
hpatel292-seneca: This is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/5a286eb40b8abf796c9e9558a1a20320668ffc0e

src/components/PromptForm/PromptSendButton.tsx
hpatel292-seneca: It is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/5a286eb40b8abf796c9e9558a1a20320668ffc0e

src/components/Menu/ModelProviderMenu.tsx
hpatel292-seneca: This is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/8e706c19f003c517e2bcc61b0293f65b449b416d

src/components/Menu/ModelProviderMenu.tsx
hpatel292-seneca: This is fixed in commit https://github.com/tarasglek/chatcraft.org/pull/730/commits/8e706c19f003c517e2bcc61b0293f65b449b416d

hpatel292-seneca:
I resolved all 'requested changes.' Sorry, I misinterpreted your message at first.

tarasglek:
@hpatel292-seneca this is a freaking huge quality of life improvement. Awesome work!

hpatel292-seneca:
Thanks @tarasglek

PR #732, opened 2024-11-16
it allows zoom-gesture when PreferencesModal is open

### LINKED ISSUE #731: PreferencesModal not allwoing zoom-gesture causes a problem on iPhone
### Description
On desktop, preventing the zoom gesture when the `PreferencesModal` is open doesn't cause any issues. However, on iPhone, when you tap an `input` element to enter an API key or modify the `Web Handlers` settings, the screen automatically zooms in as the keyboard appears. Once this happens, you are unable to return to the normal screen scale, which creates significant inconvenience.

### Possible Solutions:

1. We can prevent zoom-in by setting `maximum-scale=1`, which will not affect the desktop environment but will restrict zoom-in and zoom-out across any screen on mobile devices.

OR

2. Alternatively, we can allow zoom-in and zoom-out functionality even when the `PreferencesModal` is open, providing a more flexible mobile experience.

gitdevjin:
# Description

This fixes #731 

I implemented the second solution, enabling zoom-in functionality when the `PreferencesModal` is open on both mobile and desktop

# How I tested
I tested it on an iPhone 15 Pro and Google Chrome on desktop, and it now allows zoom-in, zoom-out, as well as scrolling when zoomed in

### Zoomed in
IMG_0734

### Zoomed out
IMG_0735

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <meta name="description" content="Web-based AI Assistant for Software Developers" />
     <meta name="og:site_name" content="chatcraft.org" />
     <meta property="og:image" content="https://chatcraft.org/favicon-32x32.png" />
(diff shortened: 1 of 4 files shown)


menghif (commented):
This fixes the bug!

You should also increase the textbox font size to at least 16px to prevent the automatic zoom on safari iOS.

https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/

gitdevjin:
@menghif Thank you always for your kind feedback, 
and apologies for the late reply—I've been swamped with assignments as the semester comes to a close.
I’ve adjusted the font size of input elements in the PreferencesModal to prevent Safari’s auto-zoom-in behavior.

menghif (approved):
LGTM!

PR #738, opened 2024-11-16
Update deps Nov 2024

humphd:
Updated remaining deps, nothing seems to be an issue, but FYI in case you spot things in the coming days.


humphd:
Downgrading ESlint to 9.14.0 to deal with an upstream bug https://github.com/eslint/eslint/issues/19134

Issue #718, opened 2024-11-10
Add support for non-LLM AI provider API keys


humphd:
We currently have a way to add API keys for OpenAI compatible LLM providers.  In #715 I added support for using https://jina.ai without an API key to convert PDFs to Markdown, see https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/ai.ts#L552

They also allow you to include an API key, if you send longer documents or use it a lot.  It would be cool if we could add a personal API key and then have this code use it.

There are lots of providers that do this stuff, e.g., https://mathpix.com/. We could do a better job of allowing users to specify a "document provider" in the settings with an API key and using that when we convert certain file types.

Kannav02:
Hey,

I was looking into this issue and I found 2 possible ways to get this working,
In the following image, it can be seen that the option to add providers can be facilitated by the user

<img width="867" alt="Screenshot 2024-11-11 at 12 26 46 PM" src="https://github.com/user-attachments/assets/018f9f35-e0fb-4e16-9b89-a3862d0d6fc9">


First Method,

Maybe we can inform the user about the fact that they can add the API key for specific providers to enable premium features without getting rate limited,

As for the developer side, since the URL would be a constant here,we can uniquely identify different providers based on their URL, preferably using a Map Data structure here and storing the providers


Second Method,

I found for only specific use cases certain providers are being used, what we could do is if that certain provider is being called as a function of the prompt, before proceeding with the request we could provide a UI pop-up to the user asking if he wants to provide an API key, similar to how ChatGPT does with custom GPTs before facilitating the request and the API key could be persisted just for the duration of the chat maybe?

```
      } else if (file.type === "application/pdf") {
        const markdown = await pdfToMarkdown(file);
        onFile(file, markdown);
```

humphd:
In the Settings dialog, let's separate "LLM Providers" from "Other AI Providers" so it's clear they aren't the same thing.

Kannav02:
sure, that sounds great, just to follow up on this, do you have a specific design for the same that you want to follow, if not I can work on getting the design and this issue resolved

thank you!

humphd:
Since we only have a single non-LLM provider for now, I'd keep this simple:

1. Change the text "Providers" to "LLM Providers"
2. Add a new section below "LLM Providers" for "Other AI Providers" which will look similar to the one now, but no need for API URL I don't think (e.g. table of name, api key)
3. Hard-code `jina.ai` as the first and only one we support

[...]

PR #740, opened 2024-11-17
Add API key management for non-language model AI providers

Kannav02:
This PR Fixes issue #718 

The Objective of this PR is to achieve the following tasks

- [x] Change the Provider label text from "Providers" to "LLM Providers"
- [x] Add a new section below "LLM Providers" for "Other AI Providers"
- [x] Hard-code `jina.ai` as the first and only one we support


Kannav02:
Hey

I have tested the changes made over here and I observed the following

- [x] API key was being validated and used for subsequent requests to Jina AI Reader
- [x] Personal Document was being converted to markdown with longer lengths than the free tier of Jina AI

Image of the working solution
Screenshot 2024-11-17 at 10 40 01 PM

humphd (changes requested):
Can you please fix CI failures before we look at this?

src/lib/ChatCraftProvider.ts
```diff
@@ -84,3 +87,20 @@ export abstract class ChatCraftProvider {
...
+  apiUrl: string;
+  apiKey?: string;
+  constructor(name: string, url: string, key?: string) {
+    this.id = nanoid();
+    this.name = name;
+    this.apiUrl = url;
+    this.apiKey = key;
+  }
+  abstract validateApiKey(key: string): Promise<boolean>;
+  get clientHeaders() {
```
humphd, line 103: This is a bit odd, I'd remove it.

src/lib/file.ts
```diff
@@ -1,6 +1,7 @@
 import { compressImageToBase64, formatAsCodeBlock } from "./utils";
 import { getSettings } from "./settings";
-import { JinaAiReaderResponse, pdfToMarkdown } from "./ai";
+import { JinaAiReaderResponse } from "./ai";
```
humphd, line 3: This type should live with the `JinaAIProvider`

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+import { JinaAiReaderResponse } from "../ai";
+
+const JINA_AI = "Jina AI";
+const JINA_API_URL = "https://r.jina.ai/";
+
+export class JinaAIProvider extends NonLLMProviders {
+  constructor(apiKey?: string) {
+    super(JINA_AI, JINA_API_URL, apiKey);
+  }
+  get clientHeaders() {
```
humphd, line 11: Because you only call this in on place below, I'm not sure what the value of abstracting it out like this is.  I'd move it inline with it's only use.

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+    const headers: HeadersInit = {
```
humphd, line 12: Also, if you do this inline below, you avoid having to explicitly do types like this.

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+      Accept: "application/json",
+    };
+
+    if (this.apiKey) {
+      headers.Authorization = `Bearer ${this.apiKey}`;
+    }
+
+    return headers;
+  }
+  async validateApiKey(key: string) {
```
humphd, line 23: Remove this

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+        const error = await res.json();
+        throw new Error(`Error converting PDF file with Jina.ai Reader: ${error}`);
+      }
+      const result: JinaAiReaderResponse = await res.json();
+      if (result.code !== 200) {
+        throw new Error(`Error converting PDF file with Jina.ai Reader: got code ${result.code}`);
+      }
+      return result;
+    } catch (err) {
+      console.error(`Error converting PDF to markdown`, err);
```
humphd, line 78: Add a case for when the error means that we have exceeded the free tier so we can provide a better message.

src/hooks/use-file-import.tsx
```diff
@@ -143,7 +144,9 @@ async function processFile(
   }
 
   if (file.type === "application/pdf") {
-    const contents = await pdfToMarkdown(file);
+    const jinaAIProvider =
+      (settings.nonLLMProviders["Jina AI"] as JinaAIProvider) || new JinaAIProvider();
```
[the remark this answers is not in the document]
Kannav02: I wanted to use the API key if it was available, for that the logic that I thought was, if the user puts in the API key, a new object for the same gets stored in the localStorage for us to operate on, if no API key we just use a throwaway object to do our pdfToMarkdown operation, if I go with the static method approach, I was just a bit skeptical about how it would manage the API key.

src/lib/ChatCraftProvider.ts
```diff
@@ -84,3 +87,20 @@ export abstract class ChatCraftProvider {
...
+export abstract class NonLLMProviders {
+  id: string;
+  name: string;
+  apiUrl: string;
+  apiKey?: string;
+  constructor(name: string, url: string, key?: string) {
+    this.id = nanoid();
+    this.name = name;
+    this.apiUrl = url;
+    this.apiKey = key;
```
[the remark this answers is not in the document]
Kannav02: Yes , when i was making the object even I thought so the same, for now I just wanted to separate the parent classes for LLM providers and Non LLM providers , maybe I could resolve all of this in another issue if you want, or maybe even in this one, thank you!

src/lib/ChatCraftProvider.ts
```diff
@@ -84,3 +87,20 @@ export abstract class ChatCraftProvider {
...
+  }
+  abstract validateApiKey(key: string): Promise<boolean>;
```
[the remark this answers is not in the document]
Kannav02: sure, I will look into this

Kannav02:
> Question: can we deal with errors from Jina.ai when we try to use the free tier and send a file that's too big, indicating that you can add an API key in Settings to handle this case?

I can definitely look into this, although do you want it to be a part of this PR or should I create another PR and a separate issue for the same

Thank you

humphd:
The jina.ai error logic could happen in a follow-up, sure.

Kannav02:
I believe the code should be good for review now, I have tried to incorporate all of the changes, but if  I have missed out on something, please do let me know and I will work on the same,
thank you!

On another note, I never knew JinaAI was really this good, I put in an 85 pages PowerPoint PDF into the files and it gave me all of it back in markdown format, and all of the content was pretty good, I mean I could abstract all of this logic, create a markdown to pdf pipeline and feed it to any LLM for training, its powerful 

here is the image for the reference 

Screenshot 2024-11-22 at 8 22 14 PM

src/components/Preferences/ModelsSettings.tsx
```diff
@@ -238,6 +241,19 @@ function ModelsSettings(isOpen: ModelsSettingsProps) {
 
     setApiKeySaved(true);
   };
+  const handleNonLLMApiKeyChange = (value: string) => {
+    setNonLLMProviders({
```
humphd, line 245: Why are we store the same NonLLM Provider data twice in state?  Can we not use what's in settings only?

src/components/Preferences/ModelsSettings.tsx
```diff
@@ -509,7 +525,7 @@ function ModelsSettings(isOpen: ModelsSettingsProps) {
         <VStack gap={4}>
           <FormControl>
             <FormLabel>
-              Providers
+              LLM Providers
```
humphd, line 528: Let's add some helper text below this one too, similar to what you did below:

```tsx
              <FormHelperText fontSize="xs">
                OpenAI-Compatible Chat Completion Providers
              </FormHelperText>
```

src/components/Preferences/ModelsSettings.tsx
```diff
@@ -802,6 +818,70 @@ function ModelsSettings(isOpen: ModelsSettingsProps) {
...
+              <Thead>
+                <Tr>
+                  <Th>Name</Th>
+                  <Th>Description</Th>
+                  <Th>API Key</Th>
+                </Tr>
+              </Thead>
+              <Tbody>
+                <Tr>
+                  <Td fontSize="xs">Jina AI</Td>
```
humphd, line 856: Jina Reader API

src/components/Preferences/ModelsSettings.tsx
```diff
@@ -802,6 +818,70 @@ function ModelsSettings(isOpen: ModelsSettingsProps) {
...
+                  <Td fontSize="xs">
+                    Optional service for processing larger files.{" "}
+                    <Link
+                      href="https://jina.ai/"
```
humphd, line 860: https://jina.ai/reader/

src/hooks/use-file-import.tsx
```diff
@@ -143,7 +143,8 @@ async function processFile(
   }
 
   if (file.type === "application/pdf") {
-    const contents = await pdfToMarkdown(file);
+    const jinaAIProvider = JinaAIProvider.fromSettings();
+    const contents = await jinaAIProvider.pdfToMarkdown(file);
```
humphd, line 147: You can simplify this further:

```ts
const contents = await JinaAIProvider.pdfToMarkdown(file);
```

The `pdfToMarkdown` function can be static and call `JinaAIProvider.fromSettings()` internally.

src/hooks/use-file-import.tsx
```diff
@@ -218,13 +219,21 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
...
           } catch (err: any) {
-            if (err.cause.code === "EmptyFile") {
+            if (err.cause?.code === "EmptyFile") {
               info({
                 title: "Warning: empty file",
                 message: `The file ${file.name} was empty after processing, skipping import.`,
               });
+            } else if (err.cause?.code === "FreeTierExceeded") {
+              error({
+                title: "Free Tier Limit Exceeded",
```
humphd, line 229: This doesn't tell you which free tier you exceeded, can you say "Jina Reader API"?

src/lib/file.ts
```diff
@@ -119,7 +119,8 @@ export async function importFiles(
         });
         onFile(file, base64);
       } else if (file.type === "application/pdf") {
-        const markdown = await pdfToMarkdown(file);
+        const jinaAIProvider = JinaAIProvider.fromSettings();
```
humphd, line 122: Same comment here

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+    // case 2 when the provider doesn't exist but the api key is set
+    if (provider?.apiKey) {
+      return new JinaAIProvider(provider.apiKey);
+    }
+
+    // case 3 when the provider doesn't exist and the api key is not set
+    return new JinaAIProvider();
+  }
+
+  async pdfToMarkdown(file: File): Promise<JinaAiReaderResponse> {
```
humphd, line 43: To further explain my comments above, I'd leave this code as is, and add:

```ts
static pdfToMarkdown(file: File): Promise<JinaAiReaderResponse> {
  const jinaProvider = JinaAIProvider.fromSettings();
  return jinaProvider.pdfToMarkdown(file);
}
```

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+            resolve(base64);
+          } else {
+            reject(new Error("Unable to read PDF file"));
+          }
+        };
+        reader.onerror = (error) => reject(error);
+        reader.readAsDataURL(file);
+      });
+
+      const headers = {
```
humphd, line 64: Another pattern you can use is to make a new Headers object, then do:

```ts
if(this.apiKey) {
  headers.append("Authorization", `Bearer ${this.apiKey}`);
}
```

Which is a bit easier to read.

src/lib/providers/JinaAIProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+        }),
+        headers,
+      });
+
+      if (!res.ok) {
+        const error = await res.json();
+        // we're checking for 429 , rate limit exceeded, and also checking if the error message contains "quota"
+        if (res.status === 429 || (error?.detail && error.detail.includes("quota"))) {
+          throw new Error(
+            "Free tier limit exceeded. Please add an API key in Settings to process larger files.",
```
humphd, line 83: Jina Reader API free tier limit exceeded...

humphd (changes requested):
Looking better, some more feedback.

src/lib/providers/JinaAIProvider.ts
Kannav02: I did try this, and the only issue with this is that typescript gives a compile-time error, stating that `Property 'append' does not exist on type '{ "Content-Type": string; Accept: string; }'.ts(2339)
`, so I thought to avoid this we could use the spread operator, the other option that I have is either declaring the headers as a `Headers` object and then try it out with this pattern

Kannav02:
I've implemented the changes based on your feedback. Could you please review and let me know if any further adjustments are needed?

On another note, I have still included the helper text, but as a potential user of the platform, it was looking cluttered, should we do something about it? I still have included a margin at the bottom, but again it just doesn't feel right


Screenshot 2024-11-25 at 7 01 43 PM

humphd (approved):
This is great.  Tested locally, and it's working well.  Thanks for the code simplifications.

Excited to see this get added.

Issue #736, opened 2024-11-16
Migrate iframe-resizer-react to @iframe-resizer/react v5


humphd:
We use iframe-resizer-react to help us automatically resize HTML previews in messages, and we're currently at v1.0.8.  In the meantime, they've done a ton of work to improve it, releasing v5, see the upgrade guide in https://iframe-resizer.com/upgrade/

KrinsKumar:
Hey @humphd, this looks like a cool issue, can I work on this? If yes please assign this to me.

humphd:
Sure, go for it

KrinsKumar:
Raised a PR https://github.com/tarasglek/chatcraft.org/pull/741. Please have a look whenever!!.

PR #741, opened 2024-11-18
Issue 736 upgrade iframe version

KrinsKumar:
This PR addresses the issue #736, where the objective was to upgrade the `iframe-resizer-react` package to the current latest version. 

After a bit of research I found out that the new version was now called `@iframe-resizer/react`, and the child script that goes into the HTML content of the iframe was updated as well. Therefore the changes are as follows:
- Updated package.json to reflect correct package
- Updated the js script in `/public/js/iframeResizer.contentWindow.min.js` to the current latest child script. (this script goes into the HTML content in the iframe),
- Updated the `HtmlPreview.tsx` file to use the correct package, and use the open source license.

I was able to test this iframe on google chrome and safari on mac, and it worked perfectly.

Please let me know if any more changes are needed!!l.


public/js/iframeResizer.contentWindow.min.js
```diff
@@ -1,9 +1,26 @@
-/*! iFrame Resizer (iframeSizer.contentWindow.min.js) - v4.3.7 - 2023-09-14
```
humphd, line 1: What's the source of this file?

Based on https://iframe-resizer.com/setup/child/ it looks like we should do this via NPM and include it from `node_modules/` in vite

humphd (changes requested):

public/js/iframeResizer.contentWindow.min.js
KrinsKumar: In the docs, in the CDN version, there is a CDN link, I first used that link, But noticed that the link we were using before was cached in the public/js/ folder, So instead of plugging the link, I cached it. I can use the CDN version of the link if you want.
image

KrinsKumar: fixed!!. We are using the npm child package
Here is an example of me generating an HTML for a simple counter.
image

humphd:
Your stuff looks good, but I'm not happy with all this console spam:

Screenshot 2024-11-18 at 8 46 29 PM

Is there a way to silence all this?  If not, I think we should look for an alternative, or stick with what we have.

KrinsKumar:
Hey can you talk about how to reproduce these logs? because I dont see these logs on my console. Regardless, even I am getting the license log, I dont think we can get rid of that.
image

Issue #610, opened 2024-04-22
Add cohere free model to free endpoint, set it to default


tarasglek:
They offer free models for non-prod usage. this is a 70B(i think?) way better than other free models

```bash
curl --request POST \
  --url https://api.cohere.ai/v1/chat \
  --header 'accept: application/json' \
  --header 'content-type: application/json' \
  --header "Authorization: bearer $CO_API_KEY" \
  --data '{
    "chat_history": [
    ],
    "message": "Say this is a test!",
    "stream": true,
    "logprobs": true
  }'
{"is_finished":false,"event_type":"stream-start","generation_id":"388f3ae9-22f6-416d-b441-afd56794849d"}
{"is_finished":false,"event_type":"text-generation","text":"This"}
{"is_finished":false,"event_type":"text-generation","text":" is"}
{"is_finished":false,"event_type":"text-generation","text":" a"}
{"is_finished":false,"event_type":"text-generation","text":" test"}
{"is_finished":false,"event_type":"text-generation","text":"!"}
{"is_finished":true,"event_type":"stream-end","response":{"response_id":"d36e41d6-1911-4f97-92e5-037eca9aa2f4","text":"This is a test!","generation_id":"388f3ae9-22f6-416d-b441-afd56794849d","chat_history":[{"role":"USER","message":"Say this is a test!"},{"role":"CHATBOT","message":"This is a test!"}],"finish_reason":"COMPLETE","meta":{"api_version":{"version":"1"},"billed_units":{"input_tokens":6,"output_tokens":5},"tokens":{"input_tokens":72,"output_tokens":5}}},"finish_reason":"COMPLETE"}
```

Here is how fancy their RAG over web search is:
```bash
curl --request POST \
  --url https://api.cohere.ai/v1/chat \
  --header 'accept: application/json' \
  --header 'content-type: application/json' \
  --header "Authorization: bearer $CO_API_KEY" \
  --data '{
    "chat_history": [
      {"role": "USER", "message": "Who released command-r+"},
      {"role": "CHATBOT", "message": "Cohere"}
    ],
    "message": "What size is the model, how much does it cost compared to dbrx?",
    "connectors": [{"id": "web-search"}]
  }'
```
```yaml
response_id: 2c438b14-766c-4eb0-9c30-b857ef42a412
[...]

PR #743, opened 2024-11-19
Initial work to add support for Cohere Provider

humphd:
I tried to implement #610, and was able to get to the point of sending a chat completion.  However, Cohere's API is different enough from OpenAI that I can't use the OpenAI client.

To properly fix this, I think we need to extend Chat Providers with a way to override the client and API calls that get made with it.  This would allow us to use Anthropic directly too, and others that differ significantly.

Or we could switch to a wrapper client like https://sdk.vercel.ai/docs/introduction to help smooth out the differences.


tarasglek:
I would rather do this through a fetch proxy or similar middleware.

tarasglek:
Eg vercel sdk decided on some common format and then have translators to/from it https://github.com/vercel/ai/blob/main/packages/cohere/src/convert-to-cohere-chat-prompt.ts#L7

I think doing this to/from openai would be a more reasonable option.

src/lib/ai.ts
```diff
@@ -183,7 +183,8 @@ ${func.name}(${JSON.stringify(data, null, 2)})\n\`\`\`\n`;
     functionName: string,
     functionArgs: string
   ) => {
-    if (content.length > 0) {
+    // TODO: need to deal with Cohere returning different data for tokens...
```
tarasglek, line 186: vercel link was re this TODO

tarasglek:
excited to get cohere support, i love their web search tool

Issue #723, opened 2024-11-10
Don't show RSS Feed icon if user not logged in


humphd:
We error if you try to click the Feed icon when not logged in.  Let's not show it if it can't be clicked.

<img width="1334" alt="Screenshot 2024-11-10 at 5 12 11 PM" src="https://github.com/user-attachments/assets/2e92621a-d3a6-49cd-bb6f-d882497a0393">

tarasglek:
I think rss feature was a mistake, we should just remove it.

KrinsKumar:
Hey, can I work on this? If yes please assign this to me!!. Also, are we just remove the button from the header or removing the feed page altogether?

humphd:
Sure

uday-rana:
I think we can close this.

PR #744, opened 2024-11-19
Remove RSS icon from the header.

KrinsKumar:
This PR address issue - #723.

I have removed the icon and its logic from the header component for now.

I have not yet removed `functions/api/share/[[user_id]].ts` logic which is used to generate the RSS feed. Please let me know if you want this to be removed. 

image

Thank you.


humphd (changes requested):
You need to rip out the backend code in functions/ too.  Go look at the PRs that added all this.

humphd (changes requested):
No, you've removed all the sharing backend logic.  Please only remove the RSS feed code.

KrinsKumar:
Apologies, I have added the files back and removed the function used to generate the feed. Please let me know if this works. Thank you.

humphd (changes requested):
Need to fix the lint issues https://github.com/tarasglek/chatcraft.org/actions/runs/11923809412/job/33296868143?pr=744

functions/api/share/[[user_id]].ts
```diff
@@ -1,5 +1,3 @@
-import { Feed } from "feed";
-import { DOMParser } from "linkedom";
```
Comment (humphd) [inline functions/api/share/[[user_id]].ts:2]: If you're removing these, can you confirm that we still need the dependencies, and if we don't `import` them anywhere else, please remove from `package.json` as well?

Comment (KrinsKumar) [reply functions/api/share/[[user_id]].ts:2]: I dont see us using DOMparsing from linkedom or feed anywhere else, I have removed them from `package.json`.

humphd (approved):

humphd:
Thanks for all the work on this, @KrinsKumar!

PR #754, opened 2024-11-24
Allow sending empty prompts on mobile, but not when clicking menu

humphd:


tarasglek:
works. I would not have figured this out.

PR #755, opened 2024-12-01
add /stats command for perf stats, trace dexie reads

tarasglek:
@humphd ok to land this? Can see some pretty obvious things we shouldn't be doing.

I was gonna do some opentelemetry stuff and setup backends to analyze perf..but figured I'd start by instrumenting code...turns out all the libs that we use also use `PerformanceObserver` api, pretty cool


tarasglek:
Here is me loading an empty chatcraft chat 
image

tarasglek:
Eg we shouldn't be loading stuff from db that's not visible

humphd (approved):

tarasglek:
Added a stats command so it can be used on mobile 
image

tarasglek:
> Love how you did it as a slash command

only sane way to get this on mobile

Issue #765, opened 2025-01-01
Chat history classification


tarasglek:
I really like the idea behind the last pie chart in https://borretti.me/article/how-i-use-claude

Would be really nice to create an llm-categorized history view for chatcraft.

PR #768, opened 2025-01-03
Add analytics

humphd:
Part of #765.  I've added code to pull some analytics from the db, and created two ways to view it:

- use the `/analytics` slash command, which dumps your stats as JSON into messages in the chat
- in the Settings modal, I added another temporary pane, to try adding some charts

The UI for this is all terrible at the moment, but the stuff you get out of it is cool.  It would be easy to add more.

### DIFF
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -1,0 +1,318 @@
+import { MessageType } from "./ChatCraftMessage";
+import db from "./db";
+
+type Period = "day" | "week" | "month";
+
+// Get a standardized key for a time period
+function getPeriodKey(date: Date, periodType: Period = "day"): string {
+  switch (periodType) {
+    case "day":
+      return date.toISOString().slice(0, 10); // YYYY-MM-DD
+    case "week": {
+      const startOfWeek = new Date(date);
+      startOfWeek.setDate(date.getDate() - date.getDay());
+      return startOfWeek.toISOString().slice(0, 10);
+    }
+    case "month":
+      return date.toISOString().slice(0, 7); // YYYY-MM
+    default:
+      throw new Error(`Invalid period type: ${periodType}`);
+  }
+}
+
+// Base metrics without calculation fields
+interface BaseMetrics {
+  messageCount: number;
+  characterCount: number;
+  conversationCount: number;
+  avgResponseTime: number;
+  uniqueModels: string[];
+}
+
+// Working metrics with calculation fields
+interface WorkingPeriodMetrics extends BaseMetrics {
+  totalResponseTime: number;
+  responseCount: number;
+  conversationIds: Set<string>;
+}
+
+// Working model metrics
+interface WorkingModelMetrics {
+  messageCount: number;
+  characterCount: number;
+  avgResponseLength: number;
+  avgResponseTime: number;
+  totalResponseTime: number;
+  responseCount: number;
+}
+
+// Final model metrics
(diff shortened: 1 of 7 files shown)


humphd:
OK, made this better:

- `/analytics` works, but I've added an "Analytics" menu item
- both add an app message that shows analytics
- you can switch between periods (hour, day, week, ...)

Here's how it looks:

Screenshot 2025-01-04 at 4 31 31 PM

Could use a few improvements that I'm too lazy to figure out:

- How to support "Delete" in these `ComponentMessages`
- Better colours for dark vs. light mode (Rechart might support it, not sure)
- More analytics...

Kannav02:
Hey @humphd ,

For the analytics, I think adding user metrics to supplement the given analytics would be good, like kind of having an average token usage per user,  error rate trend

I was also wondering, what the "Delete" functionality stands for  here, is it related to how the analytics are generated when some responses are deleted? I am a bit confused

Also, If you want I can work on adding more analytics and any additional features.

Thank you for your time

src/lib/analytics.ts
```diff
@@ -0,0 +1,420 @@
+import { MessageType } from "./ChatCraftMessage";
```
tarasglek, line 1: Not a fan of this. Analytics are an SQL thing, we should do this in sql. This would make it much easier to extend and open up an option of interactive analytics. 

I have some preliminary code to export idb into duckdb, can try that. It will be less code than this.

humphd:
Switched to use the new DuckDB stuff, and it mostly works, but the SQL is not right, since my graphs don't look right.  My SQL isn't correct.

Screenshot 2025-01-13 at 9 14 02 PM

tarasglek:
It would be helpful to get a better description of what these queries are aiming. Eg on my computer I noticed that the model query was wrong
```sql
SELECT model, COUNT(*) as count
FROM messages
GROUP BY model
ORDER BY count DESC
``` 
got generated with 'select top models in messages'. Is this right?
I further refined it into 
```sql
SELECT model, COUNT(*) as count,
       ROUND(COUNT(*) * 100.0 / (SELECT COUNT(*) FROM messages), 2) as percentage
FROM messages
WHERE model != ''
GROUP BY model
ORDER BY count DESC
LIMIT 10;
```

tarasglek:
Note querying SQL in a time series way used too advanced for me. Your time series query still is
- You are doing too much, you can do the time interval(eg year) + bucket(day) outside of SQL and dramatically simplify, eg pass em as params
- Rest of it looks correct, just need to look at raw data

tarasglek:
So this is the query i got 
```
WITH time_period AS (
  SELECT INTERVAL '1 month' AS period
),
hourly_stats AS (
  SELECT 
    DATE_TRUNC('hour', CAST(date AS TIMESTAMP)) AS hour,
    COUNT(DISTINCT chatId) AS num_chats,
    COUNT(*) AS num_messages,
    SUM(LENGTH(text)) AS chars
  FROM messages
  WHERE CAST(date AS TIMESTAMP) >= CURRENT_DATE - (SELECT period FROM time_period)
  GROUP BY DATE_TRUNC('hour', CAST(date AS TIMESTAMP))
)
SELECT 
  strftime('%Y-%m-%d %H:00', hour) AS hour,
  num_chats,
  num_messages,
  chars
FROM hourly_stats
ORDER BY hour;

```
gonna pause here for now

PR #773, opened 2025-01-08
deno routing to try to make existing cloudflare code run

tarasglek:
This is a bunch of hacks to make existing cloudflare code run in Deno.

Main problem is file-system routing, we'll need to convert this to static routing some day. Also need to fix imports some day. For now this can run with smallweb/deno locally, but not on deno deploy due to needing sloppy imports and "dynamic" fs routing.


tarasglek:
@humphd what do you think about enabling these ts options https://2ality.com/2025/01/nodejs-strip-type.html#tsconfig.json and switching stuff in api/ to using extensions?

tarasglek:
Here is why the current mapping looks like
```
Route: /api/share/:user_id? -> file:///home/taras/smallweb/chatcraft/functions/api/share/[[user_id]].ts
Route: /api/share/:user -> file:///home/taras/smallweb/chatcraft/functions/api/share/[user].ts
Route: /api/proxy -> file:///home/taras/smallweb/chatcraft/functions/api/proxy.ts
Route: /api/logout -> file:///home/taras/smallweb/chatcraft/functions/api/logout.ts
Route: /api/login -> file:///home/taras/smallweb/chatcraft/functions/api/login.ts
```
I would love to have these routes to be static and code inside to be maximally standard js

PR #774, opened 2025-01-08
switch to newer (standard) worker handler syntax

tarasglek:
@humphd this is bizarre, when I use the new syntax, I
1. pnpm dev-functions
2.  try to login
3.  
I get 
image

As far as I can tell login handler is functioning normally(per dev tools), not sure why dexie is hating me :(

### DIFF
--- a/functions/api/login.ts
+++ b/functions/api/login.ts
@@ -11,64 +11,66 @@ interface Env {
   GOOGLE_OUATH_CLIENT_SECRET: string;
 }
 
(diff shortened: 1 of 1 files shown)


tarasglek:
Ok, for some reason code doesn't redirect anymore. Thus the this behavior

tarasglek:
Oh lol one does not get to use this syntax unless fully opted into advanced mode https://developers.cloudflare.com/pages/functions/advanced-mode/

tarasglek:
Ok, so we can't switch without dropping filename routing syntax altogether

PR #775, opened 2025-01-09
lastlogin for dev mode

tarasglek:
This is a lot more minimal. I'm 99% sure that lastlogin lib is just generic oauth that needs to be generalized a little


tarasglek:
@humphd it's neither here nor there. Lastpass solves as many problems as it introduces, but I think it would be a great student project to make it into a generic oauth lib for us

PR #777, opened 2025-01-10
wip: duck command

tarasglek:


humphd (changes requested):
We should learn from our mistakes in the past and move the duckdb-specific code into `lib/duckdb.ts`, then use it in the command, UI, etc.  That should also allow us to lazy-load the entire module, and all related heavy-downloads, more easily.

tarasglek:
Here is a demo. 

https://github.com/user-attachments/assets/1a546676-e428-4e57-b83d-08ec81fbf436

@humphd is this enough for you to rewrite analytics?

I think we should add a global chatcraft object with .duckdb in it

I wish duckdb had functions we can register, then we could add a func called last_table() or last_bullets() to be able to dynamically do stuff on llm output.


This could be used to ingest stuff into duckdb (eg u ask llm to ocr a table and to add it to duckdb via our tools interface)

Then we just need to find some charting lib llms are good at and we can ask llm to load data in one msg, to analyze it in another :)

tarasglek:
So for storage duckdb is actually pretty cool. I think we can switch to using dexie to store ArrayBuffers which will be passed 0-copy into duckdb similar to how i register json files, but actually as tables

Here is the arrow link
https://github.com/apache/arrow/blob/main/js/test/unit/ipc/serialization-tests.ts#L64

tarasglek:
> We should learn from our mistakes in the past and move the duckdb-specific code into `lib/duckdb.ts`, then use it in the command, UI, etc. That should also allow us to lazy-load the entire module, and all related heavy-downloads, more easily.

yeah i'm all ears on this. please feel free to rewrite this to your liking, this is just me writing does stream-of-consciousness code

humphd:
```

14:21:29.795 | ✘ [ERROR] Error: Pages only supports files up to 25 MiB in size
-- | --
14:21:29.795 |  
14:21:29.796 | assets/duckdb-eh-DrTJ_0hP.wasm is 34 MiB in size

```

We're probably going to have to load these wasm files from a CDN instead of trying to bundle ourselves.

tarasglek:
> ```
> 
> 14:21:29.795 | ✘ [ERROR] Error: Pages only supports files up to 25 MiB in size
> -- | --
> 14:21:29.795 |  
> 14:21:29.796 | assets/duckdb-eh-DrTJ_0hP.wasm is 34 MiB in size
> ```
> 
> We're probably going to have to load these wasm files from a CDN instead of trying to bundle ourselves.

https://duckdb.org/docs/api/wasm/instantiation#cdnjsdelivr

src/lib/db.ts
```diff
@@ -167,6 +168,82 @@ class ChatCraftDatabase extends Dexie {
...
+      },
+    ]);
+
+    // Step 2: Create tables in DuckDB
+    const results = await Promise.all(
+      tableData.map(async ({ name, data }) => {
+        // Drop existing table if it exists
+        try {
+          // Try to drop the table if it exists
+          await query(`DROP TABLE IF EXISTS ${name}`);
```
tarasglek, line 209: https://duckdb.org/docs/sql/statements/create_table.html#create-or-replace

src/lib/db.ts
```diff
@@ -167,6 +168,82 @@ class ChatCraftDatabase extends Dexie {
...
+        }
+
+        // Convert dates to ISO strings for JSON serialization
+        const jsonData = data.map((record) => ({
+          ...record,
+          date: record.date instanceof Date ? record.date.toISOString() : record.date,
+        }));
+
+        // Create table in DuckDB from JSON
+        await insertJSON(name, JSON.stringify(jsonData), { name, schema: "main", create: true });
```
tarasglek, line 226: I would call it registerAsJSONfile or something..cos you can also unregister it later

src/lib/db.ts
```diff
@@ -167,6 +168,82 @@ class ChatCraftDatabase extends Dexie {
...
+
+  /**
+   * Exports all tables from Dexie to DuckDB
+   * @returns Object containing table names and row counts
+   */
+  async exportToDuckDB(): Promise<{
+    tables: { name: string; rowCount: number }[];
+  }> {
+    // Step 1: Get data from each Dexie table
+    const tableData = await Promise.all([
```
tarasglek, line 180: ['chats','meesages', ..].map(name => {name, this[name].toArray()}..

tarasglek: however i suspect the giant blob might be faster. Need to compare this on a full db

src/lib/db.ts
tarasglek: i see the code drops the file

humphd:
I've simplified `duckdb.ts` to remove some of the things that I think were causing you issues.  The main idea of this file is to completely eliminate the need to think about DuckDB as a WASM instance, and to handle safely using its methods and get proper cleanup.

I've updated the `/duck` command to dump the layout of all tables if you don't specify a query, which looks like this:

## DuckDB Tables

### chats

| column_name | column_type | null | key | default | extra |
| --- | --- | --- | --- | --- | --- |
| id | VARCHAR | YES |  |  |  |
| date | VARCHAR | YES |  |  |  |
| summary | VARCHAR | YES |  |  |  |
| messageIds | VARCHAR[] | YES |  |  |  |

### functions

| column_name | column_type | null | key | default | extra |
| --- | --- | --- | --- | --- | --- |
| id | VARCHAR | YES |  |  |  |
| date | VARCHAR | YES |  |  |  |
| name | VARCHAR | YES |  |  |  |
| description | VARCHAR | YES |  |  |  |
| parameters | STRUCT("type" VARCHAR, properties STRUCT(metadata STRUCT("type" VARCHAR, description VARCHAR)), required VARCHAR[]) | YES |  |  |  |
| code | VARCHAR | YES |  |  |  |

### messages

| column_name | column_type | null | key | default | extra |
| --- | --- | --- | --- | --- | --- |
| id | VARCHAR | YES |  |  |  |
| date | VARCHAR | YES |  |  |  |
| chatId | VARCHAR | YES |  |  |  |
| type | VARCHAR | YES |  |  |  |
| text | VARCHAR | YES |  |  |  |
| user | STRUCT(username VARCHAR, "name" VARCHAR, avatarUrl VARCHAR) | YES |  |  |  |
| model | VARCHAR | YES |  |  |  |
| versions | STRUCT(id VARCHAR, date VARCHAR, model VARCHAR, "text" VARCHAR, imageUrls JSON[])[] | YES |  |  |  |
| imageUrls | JSON[] | YES |  |  |  |

### shared

| column_name | column_type | null | key | default | extra |
| --- | --- | --- | --- | --- | --- |
| id | VARCHAR | YES |  |  |  |
| url | VARCHAR | YES |  |  |  |
| date | VARCHAR | YES |  |  |  |
| summary | VARCHAR | YES |  |  |  |
| chat | STRUCT(id VARCHAR, date VARCHAR, summary VARCHAR, messages STRUCT(id VARCHAR, date VARCHAR, "type" VARCHAR, "text" VARCHAR, model VARCHAR, versions STRUCT(id VARCHAR, date VARCHAR, model VARCHAR, "text" VARCHAR)[], "user" STRUCT(username VARCHAR, "name" VARCHAR, avatarUrl VARCHAR))[]) | YES |  |  |  |

### starred

| column_name | column_type | null | key | default | extra |
| --- | --- | --- | --- | --- | --- |
| text | VARCHAR | YES |  |  |  |
| date | VARCHAR | YES |  |  |  |
| usage | BIGINT | YES |  |  |  |

tarasglek:
> I've simplified `duckdb.ts` to remove some of the things that I think were causing you issues. The main idea of this file is to completely eliminate the need to think about DuckDB as a WASM instance, and to handle safely using its methods and get proper cleanup.
> 
> I've updated the `/duck` command to dump the layout of all tables if you don't specify a query, which looks like this:
> 

1. On the cloudflare link it only shows one table when i do /duck first thing
image

2. On the plus side, the wasm is loaded ondemand from jsdelivr cdn and fairly quickly on my connection!

3. The ChatCraftAppMessage is weird, it's invisible from llm and not deletable...We should be able to ask llm how to do stuff with these


Other than issue 1, I think this is good enough to land, we can discuss any improvements to code after landing. Ok?

humphd:
I've switched the way that the import works, so it shouldn't race.  I can't reproduce the "only 1 table" bug locally anymore.

I've switched to a human message, which is probably what we want here.  The App Messages are, by definition, hidden from the LLM for info reporting.

This is probably good to land if those work for you.

tarasglek:
This is starting to be amazing https://taras-duckdb.console-overthinker-dev.pages.dev/api/share/chatcraft_dev/lqJriKJzmxCA5WSyNAAkE

humphd:
There's some weird timing bug in the duckdb code that I don't understand fully.  Probably worth filing as a separate thing we track...

tarasglek:
I guess it's a run of the mill ipc problem where we shouldn't have more than one duckdb command outstanding.

humphd:
The fact that their promises don't mean "this is done" is concerning.  We should look in their issues, I bet it's filed, or there's some other call we can make to know it's done.

tarasglek:
#780 proved my theory wrong

PR #783, opened 2025-01-15
Bridge between ChatCraft DB and DuckDB

humphd:
This builds a bridge between the ChatCraft db and DuckDB by creating a "Just In Time" `chatcraft.*` schema.  When a SQL query is run, we look for references to tables like `chatcraft.chats` and pre-import them into DuckDB from Dexie before running the query.

A new `query()` function is exposed on the `duckdb-chatcraft.ts` file, which works just like `query()` from `duckdb.ts`, but it adds this extra automatic-table creation for the `chatcraft` schema.

Originally, I wanted to do this as a UDF, but the async binding for duckdb doesn't allow it.  So I've done this instead.  See what you think.

I think for the most part, we'd always want to use the enhanced `query()` from `duckdb-chatcraft.ts`, but I've kept both exposed in case there's some reason to get the raw one from DuckDB.

### DIFF
--- a/src/lib/commands/DuckCommand.ts
+++ b/src/lib/commands/DuckCommand.ts
@@ -2,9 +2,7 @@ import { ChatCraftCommand } from "../ChatCraftCommand";
 import { ChatCraftChat } from "../ChatCraftChat";
 import { ChatCraftHumanMessage } from "../ChatCraftMessage";
 import { CHATCRAFT_TABLES } from "../../lib/db";
-import { queryResultToJson } from "../duckdb";
-import { query } from "../duckdb-chatcraft";
-import { jsonToMarkdownTable } from "../utils";
+import { queryToMarkdown } from "../duckdb-chatcraft";
 
 export class DuckCommand extends ChatCraftCommand {
   constructor() {
@@ -23,9 +21,7 @@ export class DuckCommand extends ChatCraftCommand {
     }
 
     const sql = args.join(" ");
-    const data = await query(sql);
-    const json = queryResultToJson(data);
-    const markdown = jsonToMarkdownTable(json);
+    const markdown = await queryToMarkdown(sql);
     const message = [
       // show query
       "```sql",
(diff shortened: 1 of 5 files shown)


src/lib/commands/DuckCommand.ts
```diff
@@ -1,44 +1,38 @@
...
-      await Promise.all(
-        tableNames.map(async (name: string) => {
-          const rowCount = exportResult.tables.find((table) => table.name === name)?.rowCount || 0;
-          const tableDescription = await queryToMarkdown(`DESCRIBE ${name}`);
-          message.push(`### ${name} (${rowCount} rows)`, tableDescription);
+      return chat.addMessage(
+        new ChatCraftHumanMessage({
+          text:
+            `Available DuckDB Tables - ` +
+            `${CHATCRAFT_TABLES.map((t) => `chatcraft.${t}`).join(", ")}\n\nTry \`/duck describe chatcraft.chats\``,
```
tarasglek, line 20: blank /duck command should result in [fake] `show tables` output here like 
image

src/lib/commands/DuckCommand.ts
```diff
@@ -1,44 +1,38 @@
...
+    const json = queryResultToJson(data);
+    const markdown = jsonToMarkdownTable(json);
     const message = [
       // show query
       "```sql",
       sql,
       "```",
       // show results
-      results,
+      markdown,
```
tarasglek, line 35: We should have a common path here...lets leave this unchanged..so running via run and this ends up in same command

tarasglek (approved):
Added code to lazily create these tables once we absolutely know they are needed and some TODOs in the code.

I would really like these tables be accessible as part of run SQL functionality

humphd:
Nice improvements.  I did some refactoring and it works with both `/duck` and `run`.

I just need to fix the default `/duck` message and shared code paths and this is done.

src/lib/commands/DuckCommand.ts
humphd: Done.  The only thing I don't love is that we're showing tables from a different schema now, but I suspect that's ok.

Another option is that I could drop the `chatcraft.*` bit completely do everything in `main`.

humphd:
I think this is mergable and we can fix things in follow-up

PR #785, opened 2025-01-16
Use React DataTable for Markdown table rendering

humphd:
Quick experiment to see what it would be like to render Markdown tables as rich UI, with paging, sorting, responsive display, etc.  I'm using https://github.com/jbetancur/react-data-table-component which is highly customizable, see https://react-data-table-component.netlify.app/?path=/docs/getting-started-intro--docs

I've wanted to be able to do paging especially when querying massive amounts of data into the DOM.

This doesn't perform amazing for streaming, but works well for the db query case.

Not sure what I think.  Maybe this could be how we render long Markdown tables?  Or make this an option in prefs?

Screenshot 2025-01-15 at 7 33 37 PM

### DIFF
--- a/src/components/Markdown.tsx
+++ b/src/components/Markdown.tsx
@@ -7,6 +7,7 @@ import rehypeExternalLinks from "rehype-external-links";
 import rehypeKatex from "rehype-katex";
 import rehypeRaw from "rehype-raw";
 import rehypeSanitize from "rehype-sanitize";
+import MarkdownTable from "./MarkdownTable";
 import "katex/dist/katex.min.css";
 
 // Use highlight.js (via lowlight) vs. prism.js (via refractor) due to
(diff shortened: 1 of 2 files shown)


tarasglek:
I think this will also help with formatting issues (eg when content has markdown-table like elements).

I think the only change that I would make is to make row instantiation lazy..eg keep arrow table around and lazily page through it. This would be very nice for analyzing large tables like our messages table...this would also solve the problem of only sending limited context back to llm (eg only visible rows) if done right

tarasglek:
Eg, atm i can't really do `SELECT * from chatcraft.messages`

humphd:
I don't fully understand how to make this a "live" component vs. something that we render based on the contents of the Markdown.  The current code is based on whatever is in the DOM--it's just changing it appears.  There's no dynamic connection between the db, query, and table component.

Somehow, we'd need to tell the Markdown to render some React component and pass it a URL it can use to pull this data via the service worker or something.  I need to think about this...

Issue #788, opened 2025-01-18
RFC: Files Table and DuckDB


humphd:
# Add Files Table and DuckDB File Integration

We've been adding DuckDB support to ChatCraft and it has some powerful features we'd like to use with files.  For example, being able to use SQL to query CSV, JSON, or Excel files, do full-text search of large text files, etc.  DuckDB-wasm also has its own concept of a file system, which we'd like to connect to our current files feature.

## Current Behaviour

Our current file feature mostly lives in `src/hooks/use-file-import.tsx`.  Users can copy/paste, drag-and-drop, or click `Options > Attach Files...` to include file content in messages.  In other words, file contents (text or extracted text, images as base64 encoded URLs) are included in messages to be sent to the LLM. This works well for small files.

Now that we have DuckDB and the concept of Tables, we want to add Files as well.  Files would be separate from messages, and would be stored in binary form in Dexie.

### 1. New Dexie Table

Add a new Table to Dexie, `files`, with the following layout:

```ts
interface ChatCraftFile {
  id: string; // unique hash of the file contents, for deduping
  name: string; // original file name
  type: string; // mime-type of the file (e.g., "text/csv")
  size: number; // size of file in bytes
  content: Blob; // binary content of file
  text?: string; // extracted text of file, base64 encoded version, etc  
  created: Date; // when the file was created
  expires: Date; // when the file can be deleted
  metadata?: Record<string, unknown> // extra metadata
}
```

### 2. File Processing Logic

Modify `src/hooks/use-file-import.tsx` as follows:

- always store imported files in the `files` table
- add logic to decide when to include a file's contents in the message (e.g., based on file type, size).  For example, a small PDF would get added as a message but a large one would not
- modify how images work to add as Markdown to human messages when including vs. what we do now with image URLs
- larger files can become available for RAG, DuckDB queries
- Periodic file expiry (e.g., timeout that runs hourly or something and looks at expiry field to compare to current date/time?)

### 3. DuckDB Integration

We need to expose files in the current chat (and possibly all files?) to the DuckDB virtual filesystem.  This would allow us to do SQL queries against the file data, for example:

- full-text search (RAG) which gets included in chats
- sql queries to obtain info from CSV, Excel, etc
- sql queries to join across multiple files

It would be nice if DuckDB could also write to the Dexie `files` table, so you can safe the results of things back into the chat.

### 4. UI Changes

[...]

PR #789, opened 2025-01-20
Initial support for files table in Dexie and DuckDB

humphd:
This begins the journey to implement #788.  I've done the minimum to get a workable thing.  There is lots more to do, but it's already amazing!

- Added a new `files` table, which tracks files by their hashed content id and includes metadata
- Added a `ChatCraftFile` class to make it easier to work with files and convert to/from various things
- Added `fileIds` Array to the `chats` table to track files associated with a chat
- Connected `ChatCraftFile` and `files` to the `use-file-import.ts` hook so attaching, pasting, dragging files into a chat all add them to the `files` table and associate with the current chat's files.
- Extended `duckdb-chatcraft.ts` to know about `ChatCraftFile`, bridging the filesystem from Dexie -> DuckDB (we still need to go the other way).
- Similar to what we do with `chatcraft.*` tables, auto-injects files into DuckDB when they are referenced in a SQL query (by name or id). Ive you go to `Options > Attach Files...` and add a CSV file, doing something like `select * from 'some-file-name.csv'` just works.
- Enhanced `/duck` command to be able to show all tables and all files, including virtual files in the chat we could inject.
- Added `/ls` slash command to list files in a chat
- Always store files in Dexie files table when imported.  Currently I'm not changing how the `imageUrls` work at all.

For follow-ups--I don't want to bother with this stuff now, but so I don't forget:

- I haven't got the `runCode` working with file injection yet, since it requires the `chat` to be passed all the way through to the inner component.  We should probably make a React Context and Hook for this, so you can do `const chat = useChat()`
- File expiry.  Currently I only have an optional field.
- Start moving `imageUrls` out of `messages` and into `files`
- Figure out when to embed a file's content in the chat (e.g., based on type, size, etc).
- We need some preference UI for signalling how to handle things with files (e.g., expiry, sizes when to inject)
- Think about how to handle sharing.  Currently I don't try to serialize files into shared chats.
- Add support for using partial versions of file hashes (like short SHAs in GitHub) so you don't have paste the whole thing into a SQL statement
- Think about file deduping (e.g., we store by hash id) and how that affects file names.  I could add the same file more than once, and the original filename might differ from the new one.  Do we care?

### DIFF
--- a/src/hooks/use-file-import.tsx
+++ b/src/hooks/use-file-import.tsx
@@ -182,29 +182,25 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
 
   const importFile = useCallback(
     async (file: File, contents: string | JinaAiReaderResponse) => {
-      let chatCraftFile: ChatCraftFile;
+      const isImage = file.type.startsWith("image/");
+      const isPDF = file.type === "application/pdf";
+      const isWordDoc =
+        file.type === "application/vnd.openxmlformats-officedocument.wordprocessingml.document";
 
-      if (file.type.startsWith("image/")) {
-        const base64 = contents as string;
-        chatCraftFile = await ChatCraftFile.findOrCreate(file, { text: base64 });
-        onImageImport(base64);
-      } else if (file.type === "application/pdf") {
-        const document = (contents as JinaAiReaderResponse).data;
-        chatCraftFile = await ChatCraftFile.findOrCreate(file, { text: document.content });
-        await chat.addMessage(new ChatCraftHumanMessage({ text: `${document.content}\n` }));
-      } else if (
-        file.type === "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
-      ) {
-        const document = contents as string;
-        chatCraftFile = await ChatCraftFile.findOrCreate(file, { text: document });
-        await chat.addMessage(new ChatCraftHumanMessage({ text: `${document}\n` }));
-      } else {
-        const document = contents as string;
-        chatCraftFile = await ChatCraftFile.findOrCreate(file, { text: document });
-        await chat.addMessage(new ChatCraftHumanMessage({ text: `${document}\n` }));
-      }
+      // Extract text based on file type
+      const text = isPDF ? (contents as JinaAiReaderResponse).data.content : (contents as string);
 
+      // Create or find the ChatCraftFile in the files table and add to the chat
+      const chatCraftFile = await ChatCraftFile.findOrCreate(file, { text });
       await chat.addFile(chatCraftFile);
+
+      // Add the file's contents to the chat as a message
+      if (isImage) {
+        onImageImport(text);
+      } else if (isPDF || isWordDoc || !isImage) {
+        // Add the content as a human message for non-image files
+        await chat.addMessage(new ChatCraftHumanMessage({ text: `${text}\n` }));
+      }
     },
     [chat, onImageImport]
   );
--- a/src/hooks/use-file-import.tsx
+++ b/src/hooks/use-file-import.tsx
@@ -5,6 +5,7 @@ import { ChatCraftHumanMessage } from "../lib/ChatCraftMessage";
 import { compressImageToBase64, formatAsCodeBlock } from "../lib/utils";
 import { getSettings } from "../lib/settings";
 import { JinaAIProvider, type JinaAiReaderResponse } from "../lib/providers/JinaAIProvider";
+import { ChatCraftFile } from "../lib/ChatCraftFile";
 
 function readTextFile(file: File) {
   return new Promise<string>((resolve, reject) => {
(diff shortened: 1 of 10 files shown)


tarasglek:
> Think about file deduping (e.g., we store by hash id) and how that affects file names. I could add the same file more than once, and the original filename might differ from the new one. Do we care?

Everything is solvable with more layers of indirection :)

This is solved by adding another 'file-content' table

src/lib/ChatCraftFile.ts
```diff
@@ -0,0 +1,317 @@
...
+
+export class ChatCraftFile {
+  readonly id: string;
+  readonly name: string;
+  readonly type: string;
+  readonly size: number;
+  readonly content: Blob;
+  text?: string;
+  readonly created: Date;
+  expires?: Date;
```
tarasglek, line 27: lets not add expires for now.. We can stuff expires into metadata

src/lib/ChatCraftFile.ts
```diff
@@ -0,0 +1,317 @@
...
+   * @param input File or ChatCraftOptions
+   * @returns ChatCraftFile instance
+   */
+  static async findOrCreate(
+    input: File | ChatCraftFileOptions,
+    additionalOptions: Partial<ChatCraftFileOptions>
+  ): Promise<ChatCraftFile> {
+    let options: ChatCraftFileOptions;
+
+    if (input instanceof File) {
```
tarasglek, line 83: this only makes sense to do via sha

src/lib/ChatCraftFile.ts
```diff
@@ -0,0 +1,317 @@
...
+ * Examples:
+ * - *.csv           -> matches: test.csv, data.csv
+ * - data/*.json     -> matches: data/config.json, data/1.json
+ * - logs/????-*.log -> matches: logs/2023-01.log, logs/2024-02.log
+ *
+ * @param pattern Glob pattern (defaults to '*' for all files)
+ * @param options Listing options for sorting and filtering
+ * @returns Array of matching ChatCraftFile instances
+ */
+export async function ls(
```
tarasglek, line 282: Why not use duckdb glob here?

src/lib/commands/DuckCommand.ts
```diff
@@ -1,33 +1,49 @@
...
 
     if (!args?.length) {
+      // Show all the tables, as well as the chatcraft.* virtual tables
       sql = "SHOW TABLES;";
       markdown = await getTables();
+      messageParts.push(...this.formatSqlQueryAndResult(sql, markdown));
+
+      // Include files if there are any, including files in the chat
+      const files = await chat.files();
+      if (files.length) {
```
tarasglek, line 35: love it

tarasglek:
Overall, nice stuff. What's left to land this? Just my review?

src/hooks/use-file-import.tsx
```diff
@@ -180,22 +181,30 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
   const settings = getSettings();
 
   const importFile = useCallback(
-    (file: File, contents: string | JinaAiReaderResponse) => {
+    async (file: File, contents: string | JinaAiReaderResponse) => {
+      let chatCraftFile: ChatCraftFile;
+
       if (file.type.startsWith("image/")) {
         const base64 = contents as string;
+        chatCraftFile = await ChatCraftFile.findOrCreate(file, { text: base64 });
```
tarasglek, line 189: I'm 99% sure
```ts
const importFile = useCallback(
  async (file: File, contents: string | JinaAiReaderResponse) => {
    const isImage = file.type.startsWith("image/");
    const isPDF = file.type === "application/pdf";

    // Extract text based on file type
    const text = isPDF
      ? (contents as JinaAiReaderResponse).data.content
      : (contents as string);

    // Handle image import and non-image logic separately
    if (isImage) {
      onImageImport(text);
    } else {
      // Add the content as a human message for non-image files
      await chat.addMessage(new ChatCraftHumanMessage({ text: `${text}\n` }));
    }

    // Create or find the ChatCraftFile and add it to the chat
    const chatCraftFile = await ChatCraftFile.findOrCreate(file, { text });
    await chat.addFile(chatCraftFile);
  },
  [chat, onImageImport]
);
```
is the correct way to rewrite this

src/lib/ChatCraftFile.ts
humphd: That's exactly what it will do.  Below this, I calculate the sha based on the file.

This lets us receive a `File` object in the app and use it to do a search, handling the hashing internally.

src/lib/ChatCraftFile.ts
humphd: This is not part of duckdb, it's the files in Dexie.  We also need it for duckdb, but not in this file.

src/lib/duckdb-chatcraft.ts
```diff
@@ -115,3 +163,61 @@ export async function getTables() {
...
   return jsonToMarkdownTable(json);
 }
+
+/**
+ * Get a list of all files available in DuckDB's virtual
+ * filesystem, as well as any files that we could inject
+ * from the chat's files.
+ * @param chat the ChatCraftChat, potentially with files
+ */
+export async function getFiles(chat: ChatCraftChat) {
```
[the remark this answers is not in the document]
humphd: Let's enhance this in a follow up.  I wonder if we should be returning an object like this:

```ts
const result = query('select ...');

// result can now be used in various ways:
const markdown = result.toMarkdown();
const json = result.toJSON();
...
```

We could add some way to do further sub-queries on it too, by passing the Arrow Table result back into DuckDB and running another sql query.

humphd:
@tarasglek I think this is good to go.  If you're ok with it, let's merge and keep improving in follow-ups.

humphd:
Made one more fix: creating files in duckdb now means they also show-up when you do `/duck`:

Screenshot 2025-01-22 at 8 37 20 PM

We can do file download, using as attachments in the chat, etc. in a follow-up.

src/lib/duckdb-chatcraft.ts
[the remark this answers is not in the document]
tarasglek: I like your proposal of an object with multiple "projections"...ok with followup

tarasglek:
Fantastic work here

Issue #786, opened 2025-01-16
Hover over effects with saved chats and shared chats.


aldrin312:
Adding a hover over effect on the saved and shared chats that would reveal the options that user can do with them.
This would streamline the UX since it can reduce the actions needed to either delete a chat or renaming it. 
Right now the user needs to select the chat first before they have an option to do anything about it.

May I work on this issue?

humphd:
Yes, that makes sense to me, and matches what we do with hover-icons on chats.

PR #791, opened 2025-01-20
Hover over effects on ChatSidebarItem and FunctionSidebareItem

aldrin312:
Issue: #786 

Added a hover over effect on the sidebar items to improve UX. Making it easy to identify which item is about to be selected.
Also show the edit and delete icons when hovering over to reduce the number of actions needed to either delete or edit the items. 

Demo:
Recording 2025-01-20 at 14 32 23


aldrin312:
@humphd I don't think I have the permission to push directly to the repo.

PR #793, opened 2025-01-21
Sidebar's accordion state saved after reloading the page

### LINKED ISSUE #782: Page refreshes after deleting a shared chat.
When I was trying to delete a shared chart it refreshes the page and resets the accordions. Making me need to open the accordion again to delete another one. This can get annoying if i want to delete multiple shared chats consecutively. 

As can be seen here:

Recording 2025-01-15 at 14 26 32

A possible solution being able to delete them without needing to select them first. This also reduce the number of actions needed to delete a shared chat.

mulla028:
Closes #782

## Description 

Accordion saves item's index in the `localStorage`, so every time user deletes chat or refreshes the page he doesn't need to re-open appropriate tab. However, default value of index is still `0` which is `Saved Chats` Accordion Item. Therefore, if user opens sidebar first time for the session, the accordion will open `Saved Chats` item.

## Preview
ScreenRecording2025-01-21at4 58 18PM-ezgif com-video-to-gif-converter

### DIFF
--- a/src/components/Sidebar/SidebarContent.tsx
+++ b/src/components/Sidebar/SidebarContent.tsx
@@ -302,15 +302,7 @@ function SidebarContent({ selectedChat, selectedFunction }: SidebarContentProps)
 
   const handleAccordionChange = (expandedIndex: number | number[] | undefined) => {
     // Check if expandedIndex isn't undefined and set index to this value, otherwise set index to 0
-    try {
-      if (expandedIndex !== undefined) {
-        setOpenIndex(expandedIndex);
-      } else {
-        setOpenIndex(0);
-      }
-    } catch (err) {
-      console.warn("Unable to save accordion state", err);
-    }
+    setOpenIndex(expandedIndex ?? 0);
   };
 
   const chatsTotal = useLiveQueryTraced<number, number>(
--- a/src/components/Sidebar/SidebarContent.tsx
+++ b/src/components/Sidebar/SidebarContent.tsx
@@ -301,14 +301,11 @@ function SidebarContent({ selectedChat, selectedFunction }: SidebarContentProps)
   );
 
   const handleAccordionChange = (expandedIndex: number | number[] | undefined) => {
-    // Use 0 if index undefined (all accordions closed, but first by default opened)
-    setOpenIndex(expandedIndex ?? 0);
-
+    // Check if expandedIndex isn't undefined and set index to this value, otherwise set index to 0
     try {
       if (expandedIndex !== undefined) {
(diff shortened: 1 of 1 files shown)


src/components/Sidebar/SidebarContent.tsx
```diff
@@ -294,6 +294,38 @@ function SidebarContent({ selectedChat, selectedFunction }: SidebarContentProps)
   const navigate = useNavigate();
   const [recentCount, setRecentCount] = useState(10);
 
+  const ACCORDION_STATE_KEY = "sidebar-accordion-index";
+
+  const getSavedAccordionIndex = () => {
```
humphd, line 299: We already have a hook we use for this:

```ts
import { useLocalStorage } from "react-use";
```

Search for other uses of it in the code and copy what they do

humphd (changes requested):

src/components/Sidebar/SidebarContent.tsx
mulla028: Thank you for the feedback! I will pay more attention

mulla028:
@humphd Do we really need to reload page after every deletion? I could remove it tho

src/components/Sidebar/SidebarContent.tsx
```diff
@@ -294,6 +295,27 @@ function SidebarContent({ selectedChat, selectedFunction }: SidebarContentProps)
   const navigate = useNavigate();
   const [recentCount, setRecentCount] = useState(10);
 
+  const [openIndex, setOpenIndex] = useLocalStorage<number | number[] | undefined>(
+    "sidebar-accordion-index",
+    0
+  );
+
+  const handleAccordionChange = (expandedIndex: number | number[] | undefined) => {
```
humphd, line 303: Simplify this:

```ts
const handleAccordionChange = (expandedIndex: number | number[] | undefined) => {
  // Use 0 if index undefined (all accordions closed, with 0 to open the first)
  setOpenIndex(expandedIndex ?? 0);
};
```

humphd (changes requested):
Better, let's simplify more...

humphd (changes requested):
Why didn't you use the version I gave you?  What you have is still too complicated.

What are you expecting to throw here?  You don't need try/catch.

You don't need if/else, you can do it with what I showed you in a single line.

Less code, fewer bugs.

mulla028:
@humphd oh, my apologies. I accidentally deleted wrong part, considering it as quick fix...

humphd (approved):
Works well, looks good.  Nice work.

Issue #792, opened 2025-01-21
Icon for settings and logout in the header menu.


aldrin312:
I was having issues trying to find the settings initially, couldn't find it until I took the time to thoroughly look for it. When I found it didn't have an icon beside it indicating that its the settings, which I was originally searching for at first.

Adding a settings icon might make it easier to find for a new user. It will also might the text align with the other menu items.

![Image](https://github.com/user-attachments/assets/ede10649-3daf-42b4-b3bc-0eaf3f1b9364)

The logout one might also need an icon and possibly make it red so its easy to find for a first time user.

May I work on this?

humphd:
Go for it @aldrin312

PR #797, opened 2025-01-24
Icons for settings and logout 

aldrin312:
Issue: #792 

Added react-icons for settings and logout in the header menu.

image


humphd (changes requested):
Screenshot 2025-01-26 at 12 40 34 PM

Let's be consistent with icons we use in the UI--we already have a "gear" icon being used in the Options menu, let's reuse that.

humphd (approved):

humphd:
`Merge branch 'main' into menuIcons` please don't do this.  If you need to update your branch, rebase it.  We only want to merge once when we land changes.

This looks good otherwise.

PR #799, opened 2025-01-25
Add virtual fs module

humphd:
We need a way to work with our "filesystem," which is able to work across Dexie and DuckDB.  The files in Dexie are `ChatCraftFile` objects, and DuckDB uses `WebFile` instances.

This PR adds an `fs` module that creates a virtual filesystem that links the two file types and provides a common interface.  Using it, we can:

- `ls()` to get all files
- `readFile(path, chat)` to get a `Blob` for a given file
- `exists(path, chat)` to see if a file exists
- `rm(path, chat)` to delete a file
- `getFileUrl(path, chat)` to get a URL we can use for the file
- `downloadFile(path, chat)` to trigger a download of the file from the browser

The UI can use these methods in order to create UX for managing files for the work that @mulla028 and @aldrin312 want to start working on next week.


src/lib/duckdb.ts
```diff
@@ -30,12 +31,18 @@ async function init(logToConsole = true) {
...
+ */
+export const isUsingDuckDB = (): boolean => !!_duckdb;
+
 /**
  * Gets or initializes the DuckDB instance
  * @param logToConsole Whether to log DuckDB operations to console
  * @returns Promise resolving to the DuckDB instance
  */
-export const getDuckdb = async (logToConsole?: boolean) => {
+export const requireDuckDB = async (logToConsole?: boolean) => {
```
tarasglek, line 45: love this naming

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  async read(): Promise<Blob> {
+    return this.chatCraftFile.content;
+  }
+
+  async delete(): Promise<void> {
+    return await ChatCraftFile.delete(this.chatCraftFile.id);
+  }
+
+  async getUrl(): Promise<string> {
```
tarasglek, line 67: call it toURL(), and add comment that this converts file to data url.

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  constructor(chatCraftFile: ChatCraftFile) {
+    this.chatCraftFile = chatCraftFile;
+    this.id = chatCraftFile.id;
+    this.name = chatCraftFile.name;
+    this.size = chatCraftFile.size;
+    this.type = chatCraftFile.type;
+  }
+
+  async read(): Promise<Blob> {
```
tarasglek, line 59: toBlob

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+    return this.chatCraftFile.content;
+  }
+
+  async delete(): Promise<void> {
```
tarasglek, line 63: removeFile() for clarity and consistency with the Web File API.

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+    return true;
+  } catch (error) {
+    if (error instanceof FileNotFoundError) {
+      return false;
+    }
+    throw error;
+  }
+}
+
+export async function rm(path: string, chat: ChatCraftChat): Promise<void> {
```
tarasglek, line 205: removeFile

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+  const file = await getFile(path, chat);
+  await file.delete();
+}
+
+export async function getFileUrl(path: string, chat: ChatCraftChat): Promise<string> {
```
tarasglek, line 210: fileToURL

src/lib/fs.ts
```diff
@@ -0,0 +1,218 @@
...
+    const blob = await this.read();
+    download(blob, this.name, this.type);
+  }
+}
+
+/**
+ * List all ChatCraftFiles and DuckDB WebFiles. When there are duplicates,
+ * between the two filesystems, use the ChatCraftFile.
+ */
+export async function ls(chat: ChatCraftChat): Promise<VirtualFile[]> {
```
tarasglek, line 158: I would prefer opfs style Directory object and .values() on it to iterate through stuff instead of unix-style toplevel functions

https://shaneosullivan.wordpress.com/2023/11/28/how-to-list-all-files-in-a-browsers-origin-private-file-system/

tarasglek:
The mechanics of doing file stuff seem excellent here. Please try to align more with web file and/or OPFS apis. However, I'm terrible at naming, so I am not insisting on any of my name changes.

tarasglek (approved):

src/lib/fs.ts
humphd: I've made the rest of the changes, but this one feels wrong to me.  We aren't going to have so many files in a chat, or such a deeply nested directory structure that we can't simply return them all.  Working with an iterator will make calling code more complicated, since you'll end up needing to use it in order to produce the array you actually want.

Let's not do this until my assumptions above are proven wrong, then refactor to it when it solves a problem.

PR #800, opened 2025-01-25
Migration for files to extract fileRefs (id, name)

humphd:
This PR migrates the `files` table such that you can have the same file referenced in multiple chats, but with different filenames:

## Mermaid Diagram of DB

```mermaid
erDiagram
    chats ||--o{ messages : contains
    chats ||--o{ files : references

    chats {
        string id PK
        date date
        string summary
        string[] messageIds
        FileRef[] fileRefs
    }

    messages {
        string id PK
        date date
        string chatId FK
        string type
        string model
        object user
        string text
        object func
        object[] versions
    }

    shared {
        string id PK
        string url
        string summary
        date date
        object chat
    }

    functions {
        string id PK
        date date
        string name
        string description
        object parameters
        string code
    }

    starred {
        string text PK
        date date
        number usage
    }

    files {
        string id PK
        string name
        string type
        number size
        blob content
        string text
        date created
        object metadata
    }
```

## Major Changes
1. **New File Naming System**
   - Introduced `FileRef` type to allow files to have different names in different chat contexts. A `FileRef` is `{ id, name }`
   - Files can now be renamed within a chat while maintaining the original file content
   - Added `originalName` property to `ChatCraftFile` to track the initial filename

2. **Improved File Storage**
   - Replaced array-based file storage with Map-based system
   - Files are now stored with both content (`ChatCraftFile`) and reference (`FileRef`) information

3. **Database Updates**
   - Added migration (v12) to convert old `fileIds` to new `fileRefs` system
   - Updated chat table schema to use `fileRefs` instead of `fileIds`

4. **New File Operations on ChatCraftChat**
   - Added `renameFile` method to change file names within a chat context
   - Enhanced `addFile` to support custom file names
   - Updated `removeFile` to handle both file and ID inputs

5. **Improved Data Consistency**
   - Enhanced file cleanup during chat deletion to prevent orphaned files
   - Added transaction-based operations for safer database updates

## API Changes
- `addFile` now accepts an optional name parameter
- `removeFile` now accepts either a `ChatCraftFile` or string ID
- Added `renameFile` method
- File names can now differ between chats while maintaining content identity

I'd appreciate some help testing this (i.e., export your db, import it to this preview branch and report back).  Should all be fine, but migrations are always scary.

### DIFF
--- a/src/lib/ChatCraftChat.ts
+++ b/src/lib/ChatCraftChat.ts
@@ -9,7 +9,13 @@ import {
   ChatCraftSystemMessage,
   type SerializedChatCraftMessage,
 } from "./ChatCraftMessage";
-import db, { ChatCraftFileTable, type ChatCraftChatTable, type ChatCraftMessageTable } from "./db";
+import db, {
+  ChatCraftFileTable,
+  FileRef,
+  isFileRef,
+  type ChatCraftChatTable,
+  type ChatCraftMessageTable,
+} from "./db";
 import summarize from "./summarize";
 import { createSystemMessage } from "./system-prompt";
 import { createDataShareUrl, createShare } from "./share";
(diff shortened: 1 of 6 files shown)


humphd:
OK, I think this is ready.  Until we have UI that uses all of this, I can't really test for bugs.  I'm sure @mulla028 will find them for me :)

tarasglek:
Sorry, I'm confused about file refs. Are these basically like Linux symlinks? Eg you can upload a file..and every time you point to it you use a FileRef(eg symlink)?

Or is this a full content-addresable storage where there can't be 2 files with same content sha?

humphd:
A `FileRef` is (currently):

- `id`: hash of record in `files` table
- `name`: the filename being used for this file

We'll only ever store the same file contents once in `files`, indexed by the hash.  Chats can include those files via `FileRef`s, which allow them to use different names, but refer to the same data.

mulla028:
> OK, I think this is ready.  Until we have UI that uses all of this, I can't really test for bugs.  I'm sure @mulla028 will find them for me :)

No doubts 🫡

PR #804, opened 2025-01-27
[UI] File Attachment Added 

### LINKED ISSUE #794: Add file attachment UI
## Description

As a user, I want to access `Attach Files` option without pressing `Options` button. Moreover, all modern AI chats provide this option next to `Text Input Field`.  It would improve UX.  I could move it next to `Start Recording` button.

## How It Looks Now

Image

## Examples How It Looks In Other Platforms

**Claude** 

Image

**ChatGPT**

Image

mulla028:
Closes #794 

## Description (UPDATED)

This PR improves UX in terms of file management. Previously user was only able to see the file in the PromptForm, but now he is able to use `paperclip` icon to see all files attached in modal window, and manage them:
- Attach files... (Same as in OptionsButton)
- Delete (using `removeFile` from `src/lib/fs.ts`)
- Download (using `downloadFile` from `src/lib/fs.ts`)

If user never attached the file(s) `paperclip` will trigger file attachment process without opening modal window. 

## Preview (UPDATED)

ScreenRecording2025-01-27at7 36 33PM-ezgif com-video-to-gif-converter

### DIFF
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -1,4 +1,14 @@
-import { Box, Flex, IconButton, Text, Tooltip, useColorModeValue } from "@chakra-ui/react";
+import {
+  Box,
+  Editable,
+  EditablePreview,
+  EditableInput,
+  Flex,
+  IconButton,
+  Text,
+  Tooltip,
+  useColorModeValue,
+} from "@chakra-ui/react";
 import { IconType } from "react-icons";
 import { IoClose } from "react-icons/io5";
 import {
@@ -15,7 +25,7 @@ import {
   BsFileEarmark,
 } from "react-icons/bs";
 import { ChatCraftChat } from "../lib/ChatCraftChat";
-import { downloadFile, removeFile } from "../lib/fs";
+import { downloadFile, removeFile, renameFile } from "../lib/fs";
 import { formatFileSize } from "../lib/utils";
 import { IoMdDownload } from "react-icons/io";
 
@@ -133,21 +143,35 @@ const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
         </Tooltip>
       </Flex>
       <Box position="absolute" bottom={6} width="full" textAlign="center">
-        <Tooltip label="Download File">
-          <Text
-            color="blue.300"
-            fontSize="sm"
-            mb={1}
-            noOfLines={1}
-            px={2}
-            onClick={async (e) => {
-              e.preventDefault();
-              await downloadFile(file.name, chat);
+        <Tooltip label="Edit Name">
+          <Editable
+            defaultValue={file.name}
+            mx="auto"
+            w="90%"
+            onSubmit={async (newVal) => {
+              await renameFile(file.name, chat, newVal);
+              onRefresh();
             }}
-            cursor="pointer"
           >
-            {file.name}
-          </Text>
+            <EditablePreview
+              fontSize="sm"
+              color="blue.300"
+              noOfLines={1}
+              px={2}
+              cursor="pointer"
+              textOverflow="ellipsis"
+              overflow="hidden"
+              whiteSpace="nowrap"
+            />
+            <EditableInput
+              fontSize="sm"
+              color="blue.300"
+              px={2}
+              textOverflow="ellipsis"
+              overflow="hidden"
+              whiteSpace="nowrap"
+            />
+          </Editable>
         </Tooltip>
         <Text fontSize="xs" color="gray.400">
           {formatFileSize(file.size)}
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -101,7 +101,6 @@ const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
             aria-label="Remove file"
             icon={<IoClose />}
             size="sm"
-            colorScheme="red"
             variant="ghost"
             onClick={async () => {
               await removeFile(file.name, chat);
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -137,7 +137,6 @@ const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
             variant="ghost"
             onClick={async () => {
               await downloadFile(file.name, chat);
-              onRefresh();
             }}
           />
         </Tooltip>
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, IconButton, Text, Tooltip } from "@chakra-ui/react";
+import { Box, Flex, IconButton, Text, Tooltip, useColorModeValue } from "@chakra-ui/react";
 import { IconType } from "react-icons";
 import { IoClose } from "react-icons/io5";
 import {
@@ -50,6 +50,8 @@ type FileIconProps = {
 };
 
 const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
+  const hoverBg = useColorModeValue("gray.300", "gray.600");
+
   // Moved the fileIcon function into the component
   const fileIcon = (name: string) => {
     const extension = name.split(".").pop()?.toLowerCase() || "";
@@ -78,6 +80,7 @@ const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
           opacity: 1,
           transform: "translateY(0)",
         },
+        bg: hoverBg,
       }}
       transition="all 0.2s ease-in-out"
     >
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -1,0 +1,136 @@
+import { Box, Flex, IconButton, Text, Tooltip } from "@chakra-ui/react";
+import { IconType } from "react-icons";
+import { IoClose } from "react-icons/io5";
+import {
+  BsFiletypeJpg,
+  BsFiletypeCsv,
+  BsFiletypeDoc,
+  BsFiletypePdf,
+  BsFiletypePng,
+  BsFiletypeXls,
+  BsFiletypeTxt,
+  BsFiletypeJson,
+  BsFiletypeMd,
+  BsFiletypeSvg,
+  BsFileEarmark,
+} from "react-icons/bs";
+import { ChatCraftChat } from "../lib/ChatCraftChat";
+import { downloadFile, removeFile } from "../lib/fs";
+import { formatFileSize } from "../lib/utils";
+
+// File icon map moved to the new component
+const FILE_ICON_MAP: Record<string, IconType> = {
+  // Images
+  jpg: BsFiletypeJpg,
+  jpeg: BsFiletypeJpg,
+  png: BsFiletypePng,
+  svg: BsFiletypeSvg,
+  // Documents
+  doc: BsFiletypeDoc,
+  docx: BsFiletypeDoc,
+  pdf: BsFiletypePdf,
+  txt: BsFiletypeTxt,
+  // Spreadsheets
+  csv: BsFiletypeCsv,
+  xls: BsFiletypeXls,
+  xlsx: BsFiletypeXls,
+  // Code
+  json: BsFiletypeJson,
+  md: BsFiletypeMd,
+};
+
+type FileIconProps = {
+  file: {
+    id: string;
+    name: string;
+    size: number;
+  };
+  chat: ChatCraftChat;
+  onRefresh: () => void;
+};
+
+const FileIcon = ({ file, chat, onRefresh }: FileIconProps) => {
+  // Moved the fileIcon function into the component
+  const fileIcon = (name: string) => {
+    const extension = name.split(".").pop()?.toLowerCase() || "";
+    const IconComponent = FILE_ICON_MAP[extension] || BsFileEarmark;
+    return <IconComponent size="80px" />;
+  };
+
+  return (
+    <Box
+      key={file.id}
+      p={6}
+      borderWidth="1px"
+      borderRadius="md"
+      minW="200px"
+      maxW="200px"
+      h="200px"
+      display="flex"
+      flexDirection="column"
+      alignItems="center"
+      justifyContent="space-between"
+      aspectRatio="1"
+      position="relative"
+      _hover={{
+        borderColor: "blue.500",
+        "& .hover-buttons": {
+          opacity: 1,
+          transform: "translateY(0)",
+        },
+      }}
+      transition="all 0.2s ease-in-out"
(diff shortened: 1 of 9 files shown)


mulla028:
@humphd @tarasglek opened a draft PR to hear about UI adjustments, also I'm having trouble with `+` button functionality while adding files. Should I copy the code from `OptionsButton` for `+`?

humphd (changes requested):
Some initial feedback:

- thanks for doing something so quickly!
- the paperclip icon should automatically open the file chooser if there are no existing attachments vs being disabled, and do the same flow as the "Attach Files..." menu item
- don't do the animated + button only the paperclip
- In the model, have the first icon in the top-left be an empty file icon with a + sign, and "Add Files..." so the user can add more files.

mulla028:
@humphd, made all changes based on review/feedback

mulla028:
@humphd , I found a bug in my code, but hitting my head off the wall, I don't understand how to fix it, tried different ways.
How to reproduce it:
- If you delete all the files using modal, and then close modal.
- Try to click on `paperclip`, it doesn't open file selector. However, debugging shows that it enters `handleAttachFiles`.

src/components/PromptForm/PaperclipIcon.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  // Generates file icons based on file extension
+  const fileIcon = (name: string) => {
+    const extension = name.split(".").pop()?.toLowerCase() || "";
+    const IconComponent = FILE_ICON_MAP[extension] || BsFileEarmark;
+    return <IconComponent size="80px" />;
+  };
+
+  // File size formator, generates file format
+  const formatFileSize = (bytes: number) => {
```
humphd, line 101: Move this into lib/utils.ts

src/components/PromptForm/PaperclipIcon.tsx
```diff
@@ -0,0 +1,276 @@
...
+    <>
+      <Tooltip label={isAttached ? "View Attached Files" : "Please Attach Files. Nothing To View"}>
+        {!isAttached && (
+          <Input
+            multiple
+            type="file"
+            ref={fileInputRef}
+            hidden
+            onChange={handleFileChange}
+            accept="image/*,text/*,.pdf,application/pdf,*.docx,application/vnd.openxmlformats-officedocument.wordprocessingml.document,.json,application/json,application/markdown"
```
humphd, line 118: I wonder if we should export this string of file formats from the import files code, so we can use it in multiple places and not worry about forgetting to update it.

src/components/PromptForm/PaperclipIcon.tsx
```diff
@@ -0,0 +1,276 @@
...
+import { useAlert } from "../../hooks/use-alert";
+import { useCallback, useRef } from "react";
+
+type PaperClipProps = {
+  chat: ChatCraftChat;
+  onAttachFiles?: (files: File[]) => Promise<void>;
+};
+
+function PaperclipIcon({ chat, onAttachFiles }: PaperClipProps) {
+  const { files, loading, error } = useFiles(chat);
```
humphd, line 46: You should get `refresh` from this hook, and call it to update `files` in various places.  The list isn't live.

src/components/PromptForm/PaperclipIcon.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    <BsFileEarmarkPlus size="80px" />
+                  </Box>
+                  <Box position="absolute" bottom={6} width="full" textAlign="center">
+                    <Text fontSize="sm" mb={1} noOfLines={1} px={2}>
+                      Attach Files...
+                    </Text>
+                  </Box>
+                </Box>
+                {files.map((file) => (
+                  <Box
```
humphd, line 192: Move this to a `FileIcon` component or something to clean this up

humphd (changes requested):
This is getting better and better. Some things I notice while testing this:

- When the app loads, there is weird flash when the paperclip seems to change state and get repositioned, or hidden/revealed?  It should always be visible.  Is it due to the mic icon changing state and causing reflow?  Maybe we should put the paperclip to the right of the mic?
- When you attach a lot of files, the modal grows in height forever.  The body of the modal should be scrollable
- the contrast on the blue and red icons for download and delete isn't great.  Use high contrast (e.g., white on black, or dark gray, etc).
- let's get rid of the "download" icon and make the filename a link that does download
- let's change the trash can to an 'X' icon
- change the tooltip for the paperclip to "Attach Files..."

mulla028:
@humphd I decided to get rid of hover background while border colour still makes user feel interaction with the file!

uday-rana:
> @humphd I decided to get rid of hover background, since I couldn't find colour that fits both themes, while border colour still makes user feel interaction with the file! By the way, I think that I didn't utilize `refreshFiles` properly

You can set different colours based on the current theme with `useColorMode`: https://v2.chakra-ui.com/docs/styled-system/color-mode#usecolormode

mulla028:
> > @humphd I decided to get rid of hover background, since I couldn't find colour that fits both themes, while border colour still makes user feel interaction with the file! By the way, I think that I didn't utilize `refreshFiles` properly
> 
> You can set different colours based on the current theme with `useColorMode`: https://v2.chakra-ui.com/docs/styled-system/color-mode#usecolormode

Thank you! Will try it out!

humphd (changes requested):
There is some odd state issue with the text pop-up on the mic and paperclip icons:

Screenshot 2025-01-28 at 3 29 38 PM

It doesn't always vanish when I move my mouse.  Could the text popups go above vs. below these so they don't cover the Ask button?

Let's not animate the `x` icon in the file icons in the modal, and let's not use a colour

Let's put the file count at the top of the modal:

```
8 Attached Files
```

Should we get rid of the `Add Files` icon, and instead make it a button in the footer of the modal?  If so we could add a "Remove All" option too.  Not sure if this is good or not, just thinking...

Another idea for a follow-up, should hovering over the icon show you the contents instead of the icon?  Something to play with later, since we have the URL for each item or its text content, so we could display an image or first few lines of text, etc.

mulla028:
@humphd may I also change mic's tooltip label placement, since it closes the ask button?

humphd:
Do it in a follow-up PR. The more we add to this, the hard it is to land.

At this point you're very close to a workable first version we can land.  I'd get the outstanding issues done here, and file follow-up issues then tackle them next in subsequent PRs.

Small PRs are the way you build big software projects.

mulla028:
Last one question before I begin. By saying:

> Let's not use colour and `x` animation

Do you mean that I may get rid of hover effect for FileIcons at this point?

humphd:
> Do you mean that I may get rid of hover effect for FileIcons at this point?

I like the `x` icon showing up when you hover over the icon.  I don't like how it animates and slides down.  Just show/hide it.

humphd (commented):
- Don't use red for the `x` icon when deleting
- Make the buttons in the footer look more like this with a single button as the "Call to Action" (e.g., "Add files...") and the secondary button being a "ghost" variant with red for "Remove All". Get rid of the icons on the buttons.  This way it's easier not to accidentally click the wrong thing:

Screenshot 2025-01-28 at 4 41 36 PM

tarasglek:
This is cool, we can download stuff out of duckdb now =D

1.
type the following command:
```
/duck copy (select 1) to 'foo.json'
```

2. click attachments and click download!

3. Notice UI bug that the file is seemingly of 0 bytes, but it downloads correctly.

image

### Thoughts

I find the download UI to be unclear. I would prefer some permanent https://react-icons.github.io/react-icons/search/#q=download icon on every attachment

In general I would expect adding an attachment to popup a sidebar with list of attachments, not to have hidden modal...then we could per-file options like [include in chat, include as RAG, exclude]

I think we can address these in follow up, this is good to land, awesome feature

tarasglek:
Again, this is amazing seeing all this infrastructure code culminate in user-visible UI!

tarasglek (approved):

mulla028:
@tarasglek would you like to have kinda same sidebar as claude.ai has? 
If you prefer to have a sidebar, I could file an issue and replace modal window in follow-up.
By the way, I don't think that `download`, `remove` and `Add Files` is enough. What other file manipulations would you like to see?

humphd:
Claude style sidebar is likely the better UI long term after we land this, and can do more research.  It probably should get integrated some how, or replace, our other sidebar, which needs improvements too.

We need Rename to get added for files too.

NOTE: the "size 0" bug with DuckDB files is due to the fact that we'd have to read the file's contents to know it's size.  If that's important, we can do it; but it seems wasteful.  Maybe we should hide the size?

@mulla028 let's discuss this a bit tomorrow and then plan to land and iterate.

mulla028:
Talking about download icon, I had it in previous commits. David suggested to replace it with clickable text. What I think is  we may have download button, but if user clicks on the file name, he may rename it.

humphd:
> Talking about download icon, I had it in previous commits. David suggested to replace it with clickable text. What I think is we may have download button, but if user clicks on the file name, he may rename it.

That works.  I didn't like how jammed-up the icon was getting with all the icons.  Maybe the delete should be in one corner, the download in the opposite?

I like the "click to rename" idea.  Chakra has UI for this.

mulla028:
May I do it in this PR?

> > Talking about download icon, I had it in previous commits. David suggested to replace it with clickable text. What I think is we may have download button, but if user clicks on the file name, he may rename it.
> 
> That works. I didn't like how jammed-up the icon was getting with all the icons. Maybe the delete should be in one corner, the download in the opposite?
> 
> I like the "click to rename" idea. Chakra has UI for this.

humphd:
Sure, but then let's end it :)

mulla028:
Will work on it in the follow-up PR 🫡

mulla028:
@humphd I implemented new function in `src/lib/fs.ts` called `renameFile`, tried to make it as good as possible. I think everything for this PR implemented.

src/lib/fs.ts
```diff
@@ -216,3 +217,46 @@ export async function downloadFile(path: string, chat: ChatCraftChat): Promise<v
   const file = await getFile(path, chat);
   await file.download();
 }
+
+export async function renameFile(
```
humphd, line 221: Let's not do this, since I have it implemented in https://github.com/tarasglek/chatcraft.org/pull/800.  I'll merge that and you can rebase and use it.

humphd (changes requested):

src/lib/fs.ts
mulla028: Riight! I was confused since you had it in #800, and decided to implement myself just for preview =)

humphd:
Do you want to fix or file an issue on the bug we saw in class today with the paperclip not opening the modal sometimes?

src/components/FileIcon.tsx
```diff
@@ -0,0 +1,184 @@
...
+      >
+        <Tooltip label="Download File">
+          <IconButton
+            aria-label="Download File"
+            icon={<IoMdDownload />}
+            size="sm"
+            variant="ghost"
+            onClick={async () => {
+              await downloadFile(file.name, chat);
+              onRefresh();
```
humphd, line 140: Do we need to refresh after you download?

mulla028:
It consistently doesn't open if you deleted all the files unless user refreshes the page. The only way to fix it , to refresh the page automatically, once user leaves the modal.

mulla028:
Let me file an issue, and I will work on it during the rest of the week. 
Let's land it 😄

src/components/PromptForm/PaperclipIcon.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }, [fileInputRef]);
+
+  const handlePaperClipToggle = () => {
+    onOpen();
+  };
+
+  const handleDeleteAll = () => {
+    files.map((file) => {
+      removeFile(file.name, chat);
+    });
```
humphd, line 61: Should add a .catch() here and wait on the promises:

```ts
await Promise.all(files.map(file => removeFile(file.name, chat))).catch(...)
```

And show an error UI

src/components/Sidebar/SidebarContent.tsx
```diff
@@ -515,7 +515,7 @@ function SidebarContent({ selectedChat, selectedFunction }: SidebarContentProps)
               <VStack px={2} align="left">
                 <Text>You don&apos;t have any functions yet.</Text>
                 <Text>Functions can be called by some models to perform tasks.</Text>
-                <Link to="/f/new" target="_blank" style={{ textDecoration: "underline" }}>
+                <Link to="/f/new" style={{ textDecoration: "underline" }}>
```
humphd, line 518: Why is this change in this PR?

src/lib/ChatCraftChat.ts
```diff
@@ -40,37 +46,77 @@ function createSummary(chat: ChatCraftChat, maxLength = 200) {
   return summary.length > maxLength ? summary.slice(0, maxLength) + "..." : summary;
 }
 
+// We store both the fileRef and file object in files
+type FileRefWithFile = { ref: FileRef; file: ChatCraftFile };
```
humphd, line 50: Same question here.  It looks like you pulled in other changes vs. rebasing.

humphd (changes requested):

src/lib/ChatCraftChat.ts
mulla028: Right, I tried to rebase, and there was a message that I must pull it for some reasons...

mulla028:
I'm sorry for doing that, hope now everything looks correct. The only thing left is to catch an error and implement UI for it

tarasglek:
Can we land this? I love this for uploading json into duckdb

src/lib/utils.ts
```diff
@@ -3,6 +3,9 @@ export const isWindows = () => !isMac();
 
 export const formatNumber = (n: number) => (n ? n.toLocaleString() : "0");
 
+export const acceptableFileFormats =
```
humphd, line 6: Move this into the use-file-imports hook file instead.

humphd (approved):
One small fix and this is good to land.

Great work @mulla028

tarasglek:
I noticed we can't upload .jsonl files is there a reason we don't allow all files?

mulla028:
Input element accepts what is defined in 'hooks/use-file-import' acceptableFileFormats, may be the cause is there

mulla028:
@tarasglek , I made a research, and added `.jsonl` files upload support.

mulla028:
It didn't work tho, this problem isn't related to current pr. We have to provide a support of `.jsonl` files, I will file an issue.

humphd:
Land this and do follow-ups, please.

tarasglek:
awesomeness!

PR #813, opened 2025-02-03
[BUG] Mic's Flickering Problem Fixed

### LINKED ISSUE #811: Don't reveal/hide Mic icon
The mic icon gets shown/hidden on startup, and it causes some visual jank in the UI, as things update and shift.  I think this is due to it loading the model and checking it's capability.

Let's enable/disable it instead, and have it always be visible.

This is more obvious now that the attach-files-icon is always present.

mulla028:
Closes #811 

## Description

Currently, once page opened or reloaded, it is checking for speech support before initialization and returns empty if unsupported, it causes the flickering since the component remounts the mic icon once supported. 

The new solution uses `isSpeechToTextSupported` in `isDisabled` prop instead, that is eliminating the flickering.

### DIFF
--- a/src/components/PromptForm/MicIcon.tsx
+++ b/src/components/PromptForm/MicIcon.tsx
@@ -33,10 +33,6 @@ export default function MicIcon({
 
   const { getSpeechToTextClient, isSpeechToTextSupported, allProvidersWithModels } = useModels();
 
-  if (!isSpeechToTextSupported) {
-    return <></>;
-  }
-
   const onRecordingStart = async () => {
     clearAudioQueue();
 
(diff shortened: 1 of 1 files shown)


humphd (approved):
Looks good to me.  @Amnish04, you cool with this approach?

Amnish04:
> Looks good to me. @Amnish04, you cool with this approach?

It's always a good idea to hide controls vs disabling, but since this is not an RBAC feature, it should be fine.

Other than that, I think it makes sense to specify the reason in a tooltip when we disable the mic icon because of provider limitations.

```ts
const tooltipLabel = useMemo(
  () =>
    !isSpeechToTextSupported
      ? "No configured provider supports speech to text conversion!"
      : isRecording
        ? "Finish Recording"
        : "Start Recording",
  [isRecording, isSpeechToTextSupported]
);
```

mulla028:
@Amnish04 is it good enough now?

Amnish04 (approved):
@mulla028 Looks good to me!

Issue #806, opened 2025-01-29
Size of model selection menu list changes on search input.


aldrin312:
The window size of the model selection menu is based on the longest model name depending on the search input. The size should remain static to stop it from moving all over the place when typing.

![Image](https://github.com/user-attachments/assets/ed0f3c96-5a8c-455e-b7f9-40403af7cdc5)

PR #815, opened 2025-02-03
Static window size for model selection list menu window

aldrin312:
Made the window for the model selection stop dynamically changing on search input for model.

Issue #806 

demo:

Recording 2025-02-03 at 13 35 14


humphd (changes requested):
The default width is a bit much when you have a large screen:

Screenshot 2025-02-03 at 7 07 07 PM

I'd say we also need a max width so it doesn't grow so large.

aldrin312:
Oh yeah I didn't think of large screens.

humphd (approved):

PR #817, opened 2025-02-04
Add scroll back to top of the message button

aldrin312:
Added a button at the end of the messages to scroll back to the top of the message if text length greater than 1500.

Demo:

Recording 2025-02-04 at 17 10 22


humphd:
I wasn't able to get this to show, maybe the 1500 length limit is too short?  What if we have it happen if the message exceeds the viewport height?

aldrin312:
Scroll back to top button now only appear if message box height is greater then 1.1 times of window size.

mulla028:
What if we make an arrow up instead of button somewhere outside of the response container? 
Imo there are already enough buttons. As a user, I would want to have an arrow that appears if the top edge of response out of user's view.

humphd:
Agreed, an icon button with an up-arrow or something might be better here, so the text doesn't distract from the content of the message.

aldrin312:
changed the button to an icon button.
image

humphd (commented):
Screenshot 2025-02-18 at 7 56 20 AM

Now we have two different arrows right next to each other.  Any reason not to have it be consistent?

I would also make it a round like the other ones we do and maybe smaller?

Works well, I like the idea.

aldrin312:
I'm not sure if they should have the same icon since they perform different functions. One opens a menu and the other scrolls the chat.
I want the user to differentiate the icons from each other, considering how they function.

aldrin312:
Looking it up online, scrolling back to top are usually indicated by these icons:

image
image

I'm not sure if these are more preferred for this.

humphd:
I like the circle and up chevron.

aldrin312:
Changed to the circle chevron icon.

image

src/components/Message/MessageBase.tsx
```diff
@@ -133,6 +134,9 @@ function MessageBase({
   const isSystemMessage = message instanceof ChatCraftSystemMessage;
   const isLongMessage =
     text.length > 5000 || (isSystemMessage && summaryText && text.length > summaryText.length);
+  const isOverflowing =
+    (messageContent.current?.offsetHeight ?? 0) > window.innerHeight * 1.1 ||
+    (isSystemMessage && summaryText && text.length > summaryText.length);
```
humphd, line 139: This is at odds with what you do on line 718 below, where you ignore isOverflowing if !isSystemMessage (i.e., why bother doing this here if you won't include system messages)?

aldrin312: Oh yeah I forgot that I added that there. I was wondering why the logic I did below was weird.

humphd (approved):

PR #822, opened 2025-02-09
Add useChat() hook and ChatProvider context

### LINKED ISSUE #818: Add useChat() hook
Many places in our code, we need access to the current chat object, which may or may not exist.  Since adding the filesystem code, we need to be able to access the chat in any scenario where we might want to work with files or the duckdb database.

For example in https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/run-code.ts#L204 we don't currently have access to the `chat` because it's not available in the component where it's called https://github.com/tarasglek/chatcraft.org/blob/main/src/components/CodeHeader.tsx#L65.

The solution to this deep "prop drilling" is to create a React context and hook.  A context is a component you wrap around your main app, and then any component deep in the tree can use the hook it provides to access data without using props.

The `useUser()` hook is a good example, see https://github.com/tarasglek/chatcraft.org/blob/main/src/hooks/use-user.tsx

The `<UserProvider>` is the context, used here to wrap our app https://github.com/tarasglek/chatcraft.org/blob/main/src/main.tsx#L23

Then in the code, whenever we need the `user` object, we can use the `useUser()` hook, e.g., https://github.com/tarasglek/chatcraft.org/blob/main/src/components/ShareModal.tsx#L199

I want this same thing for `<ChatProvider>` and `useChat()`, so that we can access the `chat` object anywhere in the app without using props.

Once that's in place, fix https://github.com/tarasglek/chatcraft.org/blob/main/src/lib/run-code.ts#L204 so we have access.

Then update all of our components that pass `chat` on props to switch to using the `useChat()` hook.

humphd:
Fixes #818, adding a new `<ChatProvider />` context and associated `useChat()` hook.  We can use this to access the current `chat` anywhere in the React tree without props.

After this lands, it would be nice to refactor our code to not bother passing the chat via props all the time.

You can test this by trying to have the LLM write SQL that references one of the files or tables in DuckDB (e.g., `select * from chatcraft.chats`) and then clicking the run button to execute it.

### DIFF
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -111,7 +111,7 @@ function CodeHeader({
     } finally {
       setIsRunning(false);
     }
-  }, [onPrompt, code, language]);
+  }, [onPrompt, code, language, chat]);
 
   const toUrl = (code: string) =>
     URL.createObjectURL(new Blob([code], { type: "text/plain;charset=utf-8" }));
--- a/src/components/CodeHeader.tsx
+++ b/src/components/CodeHeader.tsx
@@ -15,6 +15,7 @@ import { TbCopy, TbDownload, TbExternalLink, TbRun } from "react-icons/tb";
 import { download, formatAsCodeBlock } from "../lib/utils";
 import { useAlert } from "../hooks/use-alert";
 import { isRunnableInBrowser, runCode } from "../lib/run-code";
+import { useChat } from "../hooks/use-chat";
 
 type PreHeaderProps = {
   language: string;
@@ -38,6 +39,7 @@ function CodeHeader({
   const [isRunning, setIsRunning] = useState(false);
   // Only show the "Run" button for JS code blocks, and only when we aren't already loading
   const shouldShowRunButton = isRunnableInBrowser(language) && onPrompt;
+  const { chat } = useChat();
 
   const handleCopy = useCallback(() => {
     onCopy();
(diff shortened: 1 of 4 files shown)


src/components/CodeHeader.tsx
```diff
@@ -62,7 +64,7 @@ function CodeHeader({
     setIsRunning(true);
 
     try {
-      let { logs, ret } = await runCode(code, language);
+      let { logs, ret } = await runCode(code, language, chat);
```
mulla028, line 67: Since you pass `code`, `language`, and `chat` to the runCode.
We want to re-render handleRunBrowser with the latest value of `chat`. It re-renders on change of dependencies.
Therefore, you have a missing dependency: `chat` on **line 114**.
It should look like: 
```
}, [onPrompt, code, language, chat]);
```
vs. 
```
}, [onPrompt, code, language]);
```

mulla028 (changes requested):
Small fix required. Overall, LGTM!

src/components/CodeHeader.tsx
mulla028: Currently, it doesn't work as expected, but it works with the oldest value of `chat`. As you can see `messageids` aren't updating for the new chats, since `chat` value isn't updating:
image

Fixed locally, now the same query returns current data table, since **runCode** receives updated `chat` value:
image

humphd:
Thanks for catching that, @mulla028, fixed.

PR #823, opened 2025-02-09
Update deps Feb 2025

humphd:
Nothing major here, should all be safe.  Would appreciate people testing anything they care about and letting me know if it breaks.


package.json
```diff
@@ -17,45 +17,45 @@
...
+    "@eslint/compat": "^1.2.6",
+    "@szhsin/react-menu": "^4.3.0",
+    "@uiw/react-codemirror": "^4.23.8",
     "apache-arrow": "^17.0.0",
     "browser-image-compression": "^2.0.2",
-    "compromise": "^14.14.3",
+    "compromise": "^14.14.4",
     "cookie": "^1.0.2",
-    "dexie": "^4.0.10",
+    "dexie": "^4.0.11",
```
mulla028, line 33: As per [release notes](https://github.com/dexie/Dexie.js/releases/tag/v4.0.11) of `dexie`, the latest version of `dexie-cloud-addon` **required**, but I couldn't find it anywhere.

mulla028 (changes requested):
Never reviewed something like this before. Interesting experience. Hope that I am doing it correctly. Ran it locally, and it didn't break anywhere. Just some tiny changes to improve deps performance. Overall, LGTM!

aldrin312 (changes requested):
Other than openai dropping a new version today. Everything to looks fine. Nothing seems to be broken from my testing, at least on the UI side of things.

tarasglek (approved):

PR #832, opened 2025-02-18
Refactor chatbase to use the new chat completion method in ChatCraftChat class.

aldrin312:
Added a new method for chat completion the ChatCraftChat class and refactored ChatCraftBase to use the new method.
The new method doesn't implement UI elements like forceScrolling. May need to refactor again to include it.

Currently all of the hooks are being passed in from ChatCraftBase. Wasn't sure how to use them directly in ChatCraftChat class or if they are possible to use.


aldrin312:
Create a new hook for chat completion. I find it doing it like this is much more easier and cleaner.
Using it now would be just ```await chatCompletion(prompt ?? "", chat);``` and would be easily accessible anywhere in the code.

I'm not sure if the completion method in the ChatCraftChat class should be removed now.

PR #839, opened 2025-02-23
Add `CTRL+L` shortcut to clear the chat

### LINKED ISSUE #826: Add "CTRL+L" shortcut to clear chat
I have recently developed a habit to clear my terminal with the `CTRL+L` shortcut, and I have found myself accidentally trying the same with ChatCraft as well.

This makes me wonder if it's a good idea to have such a shortcut for **ChatCraft** that would do exactly the same thing as `/clear` command.

If we think this feature is worth adding, we might want to keep the following in mind:
1. We might need some form of an onboarding tooltip that makes the user aware of this shortcut when they visit for the first time.
2. Display a confirmation modal when this shortcut is used for the first time, as unaware users might accidentally lose some important chats.

Amnish04:
I've done some work to implement `CTRL+L` shortcut and close #826 

This PR:
1. Implements the keyboard shortcut for users to clear the chat using `CTRL+L`, which is commonly used to clear the screen in Unix like terminals.
2. Adds a new column to the **commands table** displaying the keyboard shortcut help text for relevant commands, falling back to `N/A` for the rest.
3. Updated the placeholder text of the Desktop Prompt to include "('CTRL+L' to clear chat)" at the end, so its easier for users to discover.

**Note:** One problem that I see with adding `keyboardShortcutDescription` to `ChatCraftCommand` is that the developers need to remember to keep the **description** and **behaviour** in sync, as both things reside at different places.

Here's what the new table looks like:
message

### DIFF
--- a/src/components/PromptForm/DesktopPromptForm.tsx
+++ b/src/components/PromptForm/DesktopPromptForm.tsx
@@ -1,4 +1,12 @@
-import { FormEvent, KeyboardEvent, type RefObject, useEffect, useMemo, useState } from "react";
+import {
+  FormEvent,
+  KeyboardEvent,
+  type RefObject,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import {
   Box,
   Card,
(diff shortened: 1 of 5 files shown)


humphd:
This needs a rebase

Amnish04:
> This needs a rebase

Just rebased

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -406,7 +442,7 @@ function DesktopPromptForm({
                         _dark={{ bg: "gray.700" }}
                         placeholder={
                           !isLoading && !isRecording && !isTranscribing
-                            ? "Ask a question or use /help to learn more"
+                            ? "Ask a question or use /help to learn more ('CTRL+L' to clear chat)"
```
humphd, line 445: CTRL+l (lowercase)

src/lib/commands/ClearCommand.ts
```diff
@@ -3,7 +3,7 @@ import { ChatCraftChat } from "../ChatCraftChat";
 
 export class ClearCommand extends ChatCraftCommand {
   constructor() {
-    super("clear", "/clear", "Erases all messages in the current chat.");
+    super("clear", "/clear", "Erases all messages in the current chat.", "CTRL+L");
```
humphd, line 6: Same thing here, lowercase l

humphd (approved):
Looks good, two small things and good to land.

src/components/PromptForm/DesktopPromptForm.tsx
Amnish04: Nice catch! Just fixed

PR #841, opened 2025-02-25
Add react scan

tarasglek:
Added it in a super-manual mode for now. @humphd if you had any better ideas lmk.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "react-icons": "^5.4.0",
     "react-markdown": "^9.0.3",
     "react-router-dom": "^6.29.0",
+    "react-scan": "^0.1.3",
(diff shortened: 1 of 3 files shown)


tarasglek:
I would love help in making our ai completions only update the AiMessage, cos perf is terrible otherwise...

menghif:
> Once you flip on react-scan, it's easier to see problems without relying on react devtools.
> 
> 
> 
> I would love help in making our ai completions only update the AiMessage instead of whole window as it happens now. Chatcraft is basically unusable for long chats on slower hw.
> 
> 
> 
> @menghif do you maybe have some cycles to look into this? I'm less talented/efficient at react-slinging. Even hack-level improvement for this would be of huge help.

I am not very familiar with this but I can try to see what I can do.

PR #843, opened 2025-02-27
Draft for system providers

tarasglek:
Idea is that we will move the free provider to be a system one.

In the wip, i simply swap the free one with another one.

### DIFF
--- a/functions/api/user-info.ts
+++ b/functions/api/user-info.ts
@@ -6,7 +6,6 @@ interface Env {
 }
 
 interface Provider {
-  name: string;
   apiUrl: string;
   defaultModel: string;
   apiKey: string;
@@ -32,22 +31,20 @@ export const onRequestGet: PagesFunction<Env> = async ({ request, env }) => {
   }
 
   const freeAINonLoggedIn: Provider = {
-    name: "Free AI",
     apiUrl: "https://free-chatcraft-ai.deno.dev/api/v1",
     defaultModel: "auto",
     apiKey: "",
   };
 
-  let systemProviders = { [freeAINonLoggedIn.name]: freeAINonLoggedIn };
+  let systemProviders = { "Free AI": freeAINonLoggedIn };
   if (username) {
     const freeAILoggedIn: Provider = {
-      name: "Custom AI Providers",
       apiUrl: "https://free-chatcraft-ai.coolness.fyi/api/v1",
       defaultModel: "auto",
       apiKey: username,
     };
 
-    systemProviders = { [freeAILoggedIn.name]: freeAILoggedIn };
+    systemProviders = { "Custom AI Providers": freeAILoggedIn };
   }
 
   const responseBody = { systemProviders };
--- a/functions/api/user-info.ts
+++ b/functions/api/user-info.ts
@@ -38,7 +38,7 @@ export const onRequestGet: PagesFunction<Env> = async ({ request, env }) => {
     apiKey: "",
   };
 
-  let systemProviders = [freeAINonLoggedIn];
+  let systemProviders = { [freeAINonLoggedIn.name]: freeAINonLoggedIn };
   if (username) {
     const freeAILoggedIn: Provider = {
       name: "Free AI",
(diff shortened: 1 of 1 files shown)


tarasglek:
Note, the idea here is that I can make a provider that gets added automatically for everyone that logs in with a particular email 
 pattern..Eg everyone at my old university logged in at @uvic.ca could get org-wide ai provider, and some unique access token to track each user's ai usage.

functions/api/user-info.ts
```diff
@@ -0,0 +1,60 @@
...
+    name: "Free AI",
+    apiUrl: "https://free-chatcraft-ai.deno.dev/api/v1",
+    defaultModel: "auto",
+    apiKey: "",
+  };
+
+  let systemProviders = [freeAINonLoggedIn];
+  if (username) {
+    const freeAILoggedIn: Provider = {
+      name: "Free AI",
```
humphd, line 44: Can we use a different name here, since they use different URLs?  Maybe `Default Provider` vs. `Free Provider` or something

functions/api/user-info.ts
```diff
@@ -0,0 +1,60 @@
...
+      apiKey: username,
+    };
+
+    systemProviders = [freeAILoggedIn];
+  }
+
+  const jsonHeaders = { "Content-Type": "application/json" };
+  const responseBody = { systemProviders };
+
+  return new Response(JSON.stringify(responseBody), {
```
humphd, line 56: return `Response.json(...)`

functions/api/user-info.ts
```diff
@@ -0,0 +1,60 @@
...
+interface Provider {
+  name: string;
+  apiUrl: string;
+  defaultModel: string;
+  apiKey: string;
+}
+
+/**
+ * Calling /api/user-info will return user info. Part of that user-info is systemProviders. System providers should be overwritable from server. Thus no need to conflict resolution if local provider settings different from user ones
+ * a) so if user already has a provider of same name as 'system' one..then we log error message and leaver user's provider as is
```
humphd, line 17: What if you need to change the URL for everyone's free provider?  We'd have no way to do this.

functions/api/user-info.ts
```diff
@@ -0,0 +1,60 @@
...
+ * b) if user already has a provider of same name marked 'system', then we are free to replace it
```
humphd, line 18: What does "marked 'system'" mean here?

functions/api/user-info.ts
```diff
@@ -0,0 +1,60 @@
...
+ * c) if user does not have a provider as returned by systemProviders, we add it
+ * d) if user has no other providers, the provider that just got added becomes the default
+ * e) if the current provider is a system one..newly added system provider becomes the default
```
humphd, line 21: This seems like it will be confusing

humphd (changes requested):
This seems like `/api/provider-info` vs. `user-info` to me (I don't see any user info).

I think you should always return the full list of "system providers" including the free/default one even if you're logged in.  Then indicate the name of the one to use in a field:

```js
{
  providers: [...],
  defaultProvider: 'provider-name-to-use-from-providers'
}
```

That way the user can select between all the various providers available to them.  For example: I might want to use the free provider even though I'm also able to use an authenticated one.

tarasglek:
> This seems like /api/provider-info vs. user-info to me (I don't see any user info).

Idea is that we can store and return stuff like system prompts here, etc in future

functions/api/user-info.ts
tarasglek: i mean when chatcraft receives these from server they should have extra metadata associated with them or stored differently than current manually-entered providers. We should not have a situation where we overwrite manual providers

functions/api/user-info.ts
tarasglek: open to other ideas

functions/api/user-info.ts
tarasglek: that means returning provider with same name, but different url

tarasglek:
added the rest of user stuff to endpoint. I think it's better to expose this via this endpoint than via cookie. Reusing cooking for now, but in future can move this info into http-only cookie

humphd:
Doing #844, I don't think we should change how we send the user info (e.g., leave it in the cookie), which is working great, and instead have this just be about sending the extra "system settings/providers" you want to use with the authenticated user.

humphd:
Merged into #844.

PR #844, opened 2025-03-02
Add system providers handling to UI

humphd:
This is the web UI pieces of #843, adding the concept of "system providers."  The list of providers available to a user now includes two types of providers:

- system providers: these are defined by ChatCraft on the server, and sent when authenticated users should have extra provider(s) added to their account.  These are not managed in settings or stored in `localStorage`
- user providers: these are user defined providers that are managed in settings and stored in `localStorage`

Users will now see both "System Providers" and "User Providers" in the list of available providers to choose from:

Screenshot 2025-03-02 at 9 23 26 AM

I've added a new `useProviders()` hook to return the list of `systemProviders`, `userProviders` and the merged `providers` with all available providers.

One thing I need help thinking through here: when you select a provider, we update `settings.currentProvider` to use it.  When you select a system provider, then log out, what should we do with the value of `settings.currentProvider`, which will point to a provider that doesn't exist anymore?  Currently I'm not handling this case.

I've made a change to the `onClick` handler for the providers menu, such that clicking on a provider doesn't close the menu.  I'm not sure if this is the right behaviour or not, and if we keep it, we might want to add a `loading` state/UI while we fetch the models for the new provider.

Until we finish #843, I've hard-coded a single system provider for testing, which needs to get fixed.


tarasglek:
I think we should store a last-use timestamp when switching providers. Then always show current most recently used provider. This will fix logout

tarasglek:
I'm also not sure we need to plumb this all the way through ui. Might be sufficient to just differentiate them in settings.

tarasglek:
Overall this is pretty cool, will need a few days to test this probably

tarasglek:
Yeah after integrating this with my pull request. We need to not have duality between system/user providers... system:boolean should just be an attribute on the provider

humphd:
- Have a single set of providers, but include a `.systemProvider` property in the provider itself when it's a system provider
- Always get the list of providers, whether logged in or not, and then sync what we get back into the settings.  To do this sync, we would: a) get the local providers from localStorage; b) remove all `.systemProviders`; c) use the downloaded `.systemProviders` to merge in; d) update localStorage.  This should handle the cases of: a) new system provider being added; b) existing system provider being modified; c) existing system provider being removed.
- If there is a conflict between your local providers (i.e., non system providers) and what is in the system providers (e.g., same name), we error and ignore the system provider.  Users can delete their local provider if they want to use the system provider.
- In the UI, don't show these system providers differently
- Move all "default" providers (e.g., the Free Provider) to the server, and we'll get it on load.  No more hard-coded providers in the UI

humphd:
Did some more work on this:

- merged in the backend function for `/api/providers` and reworked it to only send the providers.
- modified the logged-in providers to return two providers (easier to see): "Amazing Models" and "More Models".  When not logged in, you only get "Free AI Models".
- made it so that we pull the list of providers on startup and when you log in/log out.  Simplified the fetching logic to reduce re-renders
- made it so we merge the system providers with what's in local storage to show you the list of available ones.
- made it so that any errors talking to the `/api/providers` end-point will automatically fallback to the free provider like we do now, so nothing breaks.
- decided against trying to write the system providers into the settings--I think that's just asking for trouble.

Here's me not logged in, then logged in via Google:

Screenshot 2025-03-15 at 4 24 40 PM
Screenshot 2025-03-15 at 4 25 24 PM

In both cases my local OpenAI provider is there, but the list of system models differs.

I tried to start reworking the settings modal, but I honestly didn't have the patience to fuss over what is required right now.  I think we should tackle that separate to this.

This is closer than before.

You can run it locally with `pnpm dev` and localhost:5173 or `pnpm dev-functions` and localhost:9339 to include the API backend functions.

tarasglek:
Finally had time to take a look. Code looks perfect

humphd:
So should we land, or iterate more?

@Amnish04 can you please review/test this, too?

src/hooks/use-user.tsx
```diff
@@ -40,50 +32,55 @@ export const UserProvider: FC<{ children: ReactNode }> = ({ children }) => {
...
     } catch (err) {
       console.error("Unable to decode id token", { err, idToken });
     }
-  }, [idToken, setUser]);
+  }, [idToken]);
 
-  const logout = useCallback(
-    async (chatId?: string) => {
-      const logoutUrl = chatId ? `/api/logout?chat_id=${chatId}` : `/api/login`;
+  const logout = async (chatId?: string) => {
```
Amnish04, line 69: Curious why we don't wrap this in `useCallback` anymore? I see that it currently does not provide any benefit, but if we introduce any state here in the future, wouldn't the linter warning be useful?

Amnish04 (commented):
This is looking great! I've left a few comments for minor potential issues.

tarasglek:
> So should we land, or iterate more?

I wanna make an end-to-end test here, but I didn't have time for coding this weekend. I think we can land and iterate.

humphd:
Updated to fix @Amnish04's issues.  @Amnish04 can you take a quick look and then I'll land?

Amnish04:
> Updated to fix @Amnish04's issues.  @Amnish04 can you take a quick look and then I'll land?

Lgtm!

PR #846, opened 2025-03-02
Fix #833 - don't run live queries in hidden tabs

### LINKED ISSUE #833: chatcraft burns cpu in background tabs
Repro:
1. Load any old chat in one tab
2. Run /stats
3. Click new window or open chatcraft in new tab manually, chat there a bunch
4. Switch back to old chat and run /stats again

observe that we spent a bunch of time running queries that were:
a) slowing down the chat window we were actually in
b) burning cpu on background tabs we weren't using
c) likely some database contention too

Image

humphd:
Fixes #833.  This updates our traced `useLiveQuery` hook to respect page visibility, ignoring queries that try to run when the page isn't visible.

@tarasglek can you test?

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -59,7 +59,7 @@ import ImageModal from "../ImageModal";
 import Markdown from "../Markdown";
 
 // Styles for the message text are defined in CSS vs. Chakra-UI
-import { useLiveQuery } from "dexie-react-hooks";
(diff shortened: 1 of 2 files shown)


tarasglek:
Nice hack. This is a little hard to follow. The risk with this change is that you will stack up a lot of queries that will now run when tab becomes visible. Please add a measurement for query time on tab switch.

tarasglek:
I'm actually not sure if my understand is correct, can you please explain the corner cases here as you understand them?

PR #847, opened 2025-03-02
Update deps for March 2025

humphd:
Updating deps for March, most things are easy, but I had to leave `apache-arrow` at 17 vs. 19 for duckdb compat, and the new TypeScript isn't compatible with eslint yet (PR hasn't landed).  The rest is probably fine to leave. 

Please test and make sure this works for you.


mulla028 (approved):
LGTM!

Issue #838, opened 2025-02-20
Bug: Deleting an attached file and adding new file breaks the "Add File" button in the attached file window.


aldrin312:
The add file button in the attached file windows breaks after deleting all of the files and adding a new one after. Problem is fixed after closing and reopening the window.

Maybe related to issue #837.

![Image](https://github.com/user-attachments/assets/c8846208-1096-489d-a24f-e281a08a9ebf)

mulla028:
Will fix this problem, I also noticed it while had been working on File Attachments UI

PR #849, opened 2025-03-05
[FIX] File Input bug fixed

### LINKED ISSUE #837: Bug: File attachment button breaks after clicking "Removing All" button from the attached file window.
The button breaks after removing all of the files in the attached file window. Refreshing the page fixes the problem.

Image

mulla028:
Closes #837 & #838

## Description 

During the implementation of #804 I couldn't fix the problem after the deletion of all files, I couldn't interact with paperclip button unless refreshed the page. This time I realized that problem was hidden inside of the `<Input>` elements that were duplicated, and also we didn't need a condition to render it, we had to render it in any case. Good lesson for me tho :)

### DIFF
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -59,9 +59,12 @@ type FileIconProps = {
   };
   chat: ChatCraftChat;
   onRefresh: () => void;
+  fileInputRef?: React.RefObject<HTMLInputElement>;
+  totalFiles: number;
+  onClose: () => void;
 };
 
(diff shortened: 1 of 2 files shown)


aldrin312:
Removing the file via using the "x" in the attached file window, won't let it be added again.
Refreshing the page fixes it or adding a different file first and adding the first deleted file again fixes it too.

Recording 2025-03-05 at 21 53 03

mulla028:
@aldrin312 

Thank you for pointing that out, sounds like a great idea! I will file a new issue now. Landing this PR!

PR #850, opened 2025-03-07
Fix popup menu from moving when hovering over

aldrin312:
Fixed the message popup menu from moving when hovering over the menu or the message box.

Before:
Recording 2025-03-06 at 11 42 46

After:
Recording 2025-03-07 at 14 03 33


mulla028 (approved):
LGTM!

PR #854, opened 2025-03-12
Use unique names for screenshots pasted into chat

humphd:
MacOS uses `image.png` for all screenshots you paste.


humphd:
I hate this bug.  I've tried twice to fix it, and this still isn't 100% doing it.

humphd:
I think that the issue here is related to the `useChat()` hook, which was caching the chat and losing the old messages.  I've switched how `useChat()` does its internal handling of the `chat` state, so that it's based on what's in the database.

cc @aldrin312, @uday-rana, since you're both working on stuff related to `useChat()` too.

humphd:
This isn't perfect, but it's better.  I'm going to merge.

PR #856, opened 2025-03-13
Autocomplete feature for commands

aldrin312:
Added a new feature where the input box shall now open a suggestion window based on the key inputs.
Features:
- Clicking the command from the suggestion list will put the command in the input box.
- Clicking outside of the autocomplete suggestion will close the window.

Demo:
Recording 2025-03-13 at 12 34 14


humphd:
Very cool, @Amnish04 will be interested.

Amnish04 (changes requested):
This is a great start!

I haven't gone through the code yet, but found a few things while testing:
1.  The popup is hard coded to dark mode theme. We need to change based on what mode is being used.
See [useColorModeValue](https://v2.chakra-ui.com/docs/styled-system/color-mode#usecolormodevalue)
image

2. I think the correct behaviour for this should be to execute the selected command vs auto-completing the prompt in case the command needs no arguments (my bad for not mentioning in the issue). For cases where commands need arguments, we could autocomplete with template arguments like so:
```
/image <prompt>
``` 
3. The list is not accessible with keyboard. The correct behaviour should allow users to use **arrow keys** to navigate the list and **enter** key to make a selection.
4. Also, it would be good to use correct aria properties if we're rolling generic divs. This case looks like a `combobox` to me, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role. Although this could probably be done in a follow up, not sure? What do you think @humphd

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -103,7 +105,38 @@ function DesktopPromptForm({
...
+    { command: "/help", description: "Shows this help message." },
+    { command: "/commands", description: "Shows a list of supported commands in ChatCraft" },
+    { command: "/import", description: "Loads the provided URL and imports the text." },
+    {
+      command: "/image ",
+      description: "Creates an image using the provided prompt.",
+      placeholder: "[layout-option]",
+    },
+    { command: "/stats", description: "Shows performance statistics for all operations." },
+    { command: "/duck", description: "Do some SQL queries." },
```
Amnish04, line 136: Its not a good idea to hard code this. We should probably derive this dynamically from [ChatCraftCommandRegistry](https://github.com/tarasglek/chatcraft.org/blob/d3b14b1689589833ce88fb25481fd3f554d66edf/src/lib/ChatCraftCommandRegistry.ts#L6)

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                                    : undefined
+                                }
+                                overflowY="auto"
+                                flex={1}
+                              />
+                            </Box>
+                          </PopoverTrigger>
+                          <PopoverContent
+                            width={`${inputWidth}px`}
+                            borderRadius="10px"
```
Amnish04, line 581:
```suggestion
                            borderRadius={"sm"}
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                            boxShadow="lg"
+                            bg={bgColor}
+                            zIndex="1000"
+                            left={popupPosition.left}
+                          >
+                            <PopoverBody maxHeight="250px" overflowY="auto" p={2} width="100%">
```
Amnish04, line 587:
```suggestion
                            <PopoverBody
                              maxHeight="250px"
                              overflowY="auto"
                              py={1}
                              px={0}
                              width="100%"
                            >
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                              {suggestions.map((suggestion, index) => (
+                                <Box
+                                  key={index}
+                                  ref={(el) => (suggestionRefs.current[index] = el)}
+                                  p={3}
```
Amnish04, line 592:
```suggestion
                                  p={2}
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                                  bg={selectedIndex === index ? hoverBg : bgColor} // Dynamic background
+                                  _hover={{ bg: hoverBg }}
+                                  cursor="pointer"
+                                  borderRadius="8px"
```
Amnish04, line 596:
```suggestion
                                  borderRadius={"sm"}
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                                  onClick={() => {
+                                    if (inputPromptRef.current) {
+                                      inputPromptRef.current.value = "/" + suggestion.command;
+                                    }
+                                    setSuggestions([]);
+                                    setIsPopoverOpen(false);
+                                    inputPromptRef.current?.focus();
+                                  }}
+                                >
+                                  <strong>{suggestion.helpTitle}</strong>
```
Amnish04, line 607:
```suggestion
                                  <Text fontWeight={"semi-bold"}>{suggestion.helpTitle}</Text>
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -419,45 +526,104 @@ function DesktopPromptForm({
...
+                        </Box>
+                      ) : (
+                        <Popover
+                          isOpen={isPopoverOpen}
+                          onClose={() => setIsPopoverOpen(false)}
+                          autoFocus={false}
+                          placement="top"
+                        >
+                          <PopoverTrigger>
+                            <Box ref={inputBoxRef} style={{ width: "90%" }}>
```
Amnish04, line 547: Let's wrap this Popover stuff into an independent component like `AutoComplete` that can accept an `input` like element as its child.

Amnish04 (changes requested):
@aldrin312 This is great progress! I was playing around with different CSS properties and left some suggestions for a better fit.

**Before:**
image
**After:**
image

Also, I think we should wrap all this into a reusable component called something like `AutoCompleteInput`. Although, I am not sure if `Popover` is the right element to use here semantically and for accessibility purposes, but we can fix that in a follow up.

Amnish04 (changes requested):
Another issue I noticed is that we can't dismiss the suggestions popover on `Esc` click, and can't return to the input box using `Down` arrow.

aldrin312:
> Another issue I noticed is that we can't dismiss the suggestions popover on `Esc` click, and can't return to the input box using `Down` arrow.

I'm not sure what this means, like the suggestion box closes when pressing `Esc` and return focus to the input box?

Also returning to the input box when the current index is on the last one or with only one suggestion?

Amnish04:
> > Another issue I noticed is that we can't dismiss the suggestions popover on `Esc` click, and can't return to the input box using `Down` arrow.
> 
> I'm not sure what this means, like the suggestion box closes when pressing `Esc` and return focus to the input box?
> 
> Also returning to the input box when `Down` arrow is press, if the current index is on the last one or with only one suggestion?

@aldrin312 That's correct!

aldrin312:
Added the `Esc` and `Down` controls.

aldrin312:
I think everything needed is already implemented?

src/components/PromptForm/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,53 @@
+import { Popover, PopoverTrigger, PopoverContent, PopoverBody } from "@chakra-ui/react";
+import { Box, Text } from "@chakra-ui/react";
+
+const AutoCompleteProps = ({
```
Amnish04, line 4:
```suggestion
const AutoCompleteInput = ({
```

This should be named `AutoCompleteInput`.

src/components/PromptForm/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+  inputWidth,
+  popupPosition,
+  bgColor,
+  children,
+  suggestions,
+  suggestionRefs,
+  selectedIndex,
+  hoverBg,
+  onSelect,
+}: any) => {
```
Amnish04, line 16: You should define `AutoCompleteProps` vs using `any` here

src/components/PromptForm/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+  onClose,
+  inputWidth,
+  popupPosition,
+  bgColor,
+  children,
+  suggestions,
+  suggestionRefs,
+  selectedIndex,
+  hoverBg,
+  onSelect,
```
Amnish04, line 15: I think you're leaking too much implementation details here. The correct API would be to pass an `Input` component as a child to this, and a list of suggestions as a prop. The rest should be handled inside this component.

src/components/PromptForm/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+              </Box>
+            </Box>
+          ))}
+        </PopoverBody>
+      </PopoverContent>
+    </Popover>
+  );
+};
+
+export default AutoCompleteProps;
```
Amnish04, line 53: Also, let's move this to `/src/components` as this is supposed to be a generic reusable component to be used with any `Input` or `TextArea` in the application.

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -94,7 +104,21 @@ function DesktopPromptForm({
...
+  const [popupPosition, setPopupPosition] = useState<{ left: number }>({
+    left: 0,
+  });
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [inputWidth, setInputWidth] = useState<number>(0);
+  const [suggestions, setSuggestions] = useState<
+    { command: string; helpTitle: string; helpDescription: string }[]
+  >([]);
+  const [selectedIndex, setSelectedIndex] = useState<number>(-1);
+  const suggestionRefs = useRef<(HTMLDivElement | null)[]>([]);
```
Amnish04, line 119: This should all be handled by the autocomplete component.

Amnish04 (changes requested):
I think we're super close, the functionality is almost perfect!

A couple of problems I see:
1. If the suggestions list is still open and I try to submit whatever I have in `TextArea` with `Enter` key, nothing happens until I go and close the list manually.
2. The current implementation of `AutoCompleteInput` is too tightly coupled with the component that uses it. That is, a lot of what you are leaking in `DesktopPromptForm` needs to happen in the reusable component. I've written my thoughts of an ideal API in a comment below.

aldrin312:
Refactored component to be independent as possible.
Props that needs to be passed is the list of suggest and due to issues i have to pass in the reference for the text area still.

```typescript
    <AutoComplete
      triggerRef={inputPromptRef}
      parentRef={parentRef}
      availablePrompts={availablePrompts}
    >
```

@Amnish04

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  PopoverTrigger,
+  PopoverContent,
+  PopoverBody,
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { Box, Text } from "@chakra-ui/react";
+import React, { useCallback, useEffect } from "react";
+import { useRef, useState } from "react";
+
+type AutoCompleteProps = {
```
Amnish04, line 12:
```suggestion
type AutoCompleteInputProps = {
```

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  children: React.ReactNode;
+  triggerRef: React.RefObject<HTMLTextAreaElement>;
+  parentRef: React.RefObject<HTMLDivElement>;
+  availablePrompts: {
+    command: string;
+    helpTitle: string;
+    helpDescription: string;
+  }[];
```
Amnish04, line 20: Since this is a generic component, we should not couple the props with a specific use case like commands. For example, see how React MUI does this https://mui.com/material-ui/react-autocomplete/#options-structure

```suggestion
  options: {
    label: string;
  }[]
```

Its better to define an `AutoCompleteOption` type exclusively
```ts
type AutoCompleteOption = {
    label: string;
}
...
...
options: AutoCompleteOption[]
```

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  PopoverBody,
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { Box, Text } from "@chakra-ui/react";
+import React, { useCallback, useEffect } from "react";
+import { useRef, useState } from "react";
+
+type AutoCompleteProps = {
+  children: React.ReactNode;
+  triggerRef: React.RefObject<HTMLTextAreaElement>;
```
Amnish04, line 14: We should probably allow Input elements as well
```suggestion
  triggerRef: React.RefObject<HTMLTextAreaElement> | React.RefObject<HTMLInputElement>;
```

src/components/AutoCompleteInput.tsx
Amnish04: I see we need to pass in all this complex option information and have custom styling, but the right way to do this is by specifying an option renderer function like so:
```ts
type AutoCompleteInputProps<OptionType> = {
  ...
  ...
  options: OptionType[];
  renderOption: (option: OptionType) => ReactNode;
};

function AutoCompleteInput<OptionType>(props: AutoCompleteInputProps<OptionType>) {
  return (
```

See https://mui.com/material-ui/react-autocomplete/#githubs-picker for reference.

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -383,45 +424,54 @@ function DesktopPromptForm({
...
-                          isTranscribing={isTranscribing}
-                          recordingSeconds={recordingSeconds}
-                        />
-                      </Box>
-                    ) : (
-                      <AutoResizingTextarea
-                        ref={inputPromptRef}
-                        variant="unstyled"
-                        onKeyDown={handleKeyDown}
+                  <Box style={{ position: "relative", width: "100%" }}>
```
Amnish04, line 427: Chakra has these props in-built, you don't have to use `style` here

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -30,6 +39,8 @@ import ImageModal from "../ImageModal";
 import { ChatCraftChat } from "../../lib/ChatCraftChat";
 import { useFileImport } from "../../hooks/use-file-import";
 import PaperclipIcon from "./PaperclipIcon";
+import { ChatCraftCommandRegistry } from "../../lib/ChatCraftCommandRegistry";
+import AutoComplete from "../AutoCompleteInput";
```
Amnish04, line 43:
```suggestion
import AutoCompleteInput from "../AutoCompleteInput";
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -383,45 +424,54 @@ function DesktopPromptForm({
...
+                      {inputType === "audio" ? (
+                        <Box py={2} px={1} flex={1}>
+                          <AudioStatus
+                            isRecording={isRecording}
+                            isTranscribing={isTranscribing}
+                            recordingSeconds={recordingSeconds}
+                          />
+                        </Box>
+                      ) : (
+                        <AutoComplete
```
Amnish04, line 438:
```suggestion
                        <AutoCompleteInput
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -383,45 +424,54 @@ function DesktopPromptForm({
...
+                            _dark={{ bg: "gray.700" }}
+                            placeholder={
+                              !isLoading && !isRecording && !isTranscribing
+                                ? "Ask a question or use /help to learn more ('CTRL+l' to clear chat)"
+                                : undefined
+                            }
+                            overflowY="auto"
+                            flex={1}
+                          />
+                        </AutoComplete>
```
Amnish04, line 463:
```suggestion
                        </AutoCompleteInput>
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -383,45 +424,54 @@ function DesktopPromptForm({
...
+                            isRecording={isRecording}
+                            isTranscribing={isTranscribing}
+                            recordingSeconds={recordingSeconds}
+                          />
+                        </Box>
+                      ) : (
+                        <AutoComplete
+                          triggerRef={inputPromptRef}
+                          parentRef={parentRef}
+                          availablePrompts={availablePrompts}
```
Amnish04, line 441: Lets rename this to `options`

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -95,6 +106,9 @@ function DesktopPromptForm({
     onImageImport: (base64) => updateImageUrls(base64, setInputImageUrls),
   });
 
+  const availablePrompts = ChatCraftCommandRegistry.getCommands();
```
Amnish04, line 109:
```suggestion
  const availableCommands = ChatCraftCommandRegistry.getCommands();
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -95,6 +106,9 @@ function DesktopPromptForm({
...
+  const parentRef = useRef<HTMLDivElement | null>(null);
```
Amnish04, line 110:
```suggestion
  const autocompleteParentRef = useRef<HTMLDivElement | null>(null);
```

src/components/PromptForm/DesktopPromptForm.tsx
```diff
@@ -309,9 +351,8 @@ function DesktopPromptForm({
   const closeModal = () => setImageModalOpen(false);
 
   const dragDropBorderColor = useColorModeValue("blue.200", "blue.600");
-
   return (
-    <Flex dir="column" w="100%" h="100%">
+    <Flex ref={parentRef} id="parent" dir="column" w="100%" h="100%">
```
Amnish04, line 355: Why are we setting an `id` here?

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+                }
+                setIsPopoverOpen(false);
+                triggerRef.current?.focus();
+              }}
+            >
+              <Text fontWeight={"semibold"}>{suggestion.helpTitle}</Text>
+              <Box fontSize="sm" color="gray.400">
+                {suggestion.helpDescription.split(".")[0]}.
+              </Box>
+            </Box>
```
Amnish04, line 200: This should be a part of `renderOption` function

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  useColorModeValue,
+} from "@chakra-ui/react";
+import { Box, Text } from "@chakra-ui/react";
+import React, { useCallback, useEffect } from "react";
+import { useRef, useState } from "react";
+
+type AutoCompleteProps = {
+  children: React.ReactNode;
+  triggerRef: React.RefObject<HTMLTextAreaElement>;
+  parentRef: React.RefObject<HTMLDivElement>;
```
Amnish04, line 15: I see you need to pass `parentRef` to properly position the listbox relative to the prompt form container. But based on the [combobox design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/), the `listbox` is associated with the `input` element and hence, should be positioned relative to the input element itself.
Could we rather allow passing an `offset` property to position it properly similar to how Chakra does with Menus?
https://v2.chakra-ui.com/docs/components/menu/props
image

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  // const triggerRef = useRef<HTMLInputElement>(null);
+
+  const bgColor = useColorModeValue("white", "gray.700");
+  const hoverBg = useColorModeValue("gray.200", "gray.600");
+
+  const [suggestions, setSuggestions] = useState<
+    { command: string; helpTitle: string; helpDescription: string }[]
+  >([]);
+
+  // Set possition and width
```
Amnish04, line 43:
```suggestion
  // Set position and width
```

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  return (
+    <Popover
+      isOpen={isPopoverOpen}
+      onClose={() => setIsPopoverOpen(false)}
+      autoFocus={false}
+      placement="top"
+    >
+      <PopoverTrigger>
+        <PopoverTrigger>
```
Amnish04, line 151: We have 2 `PopoverTrigger`s nested

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+const AutoCompleteInput = ({
+  children,
+  triggerRef,
+  parentRef,
+  availablePrompts,
+}: AutoCompleteProps) => {
+  const [popupPosition, setPopupPosition] = useState<number>(0);
+  const [inputWidth, setInputWidth] = useState<number>(0);
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [selectedIndex, setSelectedIndex] = useState<number>(-1);
```
Amnish04, line 32:
```suggestion
  const [focusedIndex, setFocusedIndex] = useState<number>(-1);
```

`focusedIndex` would be a better name since this represents the focused item on the list.

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const suggestionRefs = useRef<(HTMLDivElement | null)[]>([]);
+  // const triggerRef = useRef<HTMLInputElement>(null);
```
Amnish04, line 34: Delete this

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const bgColor = useColorModeValue("white", "gray.700");
+  const hoverBg = useColorModeValue("gray.200", "gray.600");
+
+  const [suggestions, setSuggestions] = useState<
+    { command: string; helpTitle: string; helpDescription: string }[]
```
Amnish04, line 40: This type should be dynamic

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  >([]);
+
+  // Set possition and width
+  useEffect(() => {
+    if (triggerRef.current && parentRef.current) {
+      const parentRect = parentRef.current.getBoundingClientRect();
+      setInputWidth(parentRect.width);
+      setPopupPosition((parentRect.left + window.scrollX) / 10);
+    }
+  }, [parentRef, suggestions.length, triggerRef]);
```
Amnish04, line 50:
```suggestion
  }, [parentRef, triggerRef]);
```

`suggestions` is not a dependency of this `useEffect`

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  useEffect(() => {
+    const updateWidth = () => {
+      if (triggerRef.current && parentRef.current) {
+        setInputWidth(parentRef.current.getBoundingClientRect().width);
+      }
+    };
+    window.addEventListener("resize", updateWidth);
+    return () => window.removeEventListener("resize", updateWidth);
+  }, [parentRef, triggerRef]);
+  // Onchange event for popup content
```
Amnish04, line 62:
```suggestion
  // When associated input's value is changed
```

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  handleChange(e);
+                },
+                onKeyDown: (e: React.KeyboardEvent<HTMLInputElement>) => {
+                  if (!isPopoverOpen) {
+                    children.props.onKeyDown?.(e);
+                  }
+                  handleKeyDown(e);
+                },
+              })
+            : children}
```
Amnish04, line 166: We should add a comment here to describe what's happening as its not obvious.

src/components/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+      }
+    },
+    [isPopoverOpen, selectedIndex, suggestionRefs, suggestions, triggerRef]
+  );
+
+  return (
+    <Popover
+      isOpen={isPopoverOpen}
+      onClose={() => setIsPopoverOpen(false)}
+      autoFocus={false}
```
Amnish04, line 147: Why don't we use this? Is it because this expects a natively focusable element and we don't have any as we use divs?

Amnish04 (changes requested):
@aldrin312 We're getting closer! I've left some comments to get your feature in a landable state. There are still some accessibility issues that I haven't pointed out, but we can take care of them in a follow up.

PR #857, opened 2025-03-22
Update API key field to not erase key when it does not validate

### LINKED ISSUE #851: Should not erase key when it does not validate
https://github.com/user-attachments/assets/ce066749-bff5-4671-8dd9-76ab247e3e8a

hpatel292-seneca:
fixed #851 

Update ModelsSettings.tsx to keep the API key value.
image


uday-rana (approved):
Nice fix!

PR #860, opened 2025-03-26
Use useChat in MessageBase.tsx

### LINKED ISSUE #835: Refactor Code To Use `useChat()` Hook
## Description

Since @humphd landed #822, introduced `ChatProvider` & `useChat()` hook. We need to refactor code to stop passing `chat` on `props` in components and utilize `useChat()` hook wherever it's possible .

uday-rana:
Fixes #835

The `useChat` hook can't be used in any components that are children of `ChatBase`. `ChatBase` receives the `chat` prop from it's parent components `LocalChat` and `RemoteChat`. These components retrieve `chat` using `useLoaderData()`, while the hook retrieves `chat` from the route parameter using `useParams()`. From what I understand, this means the hook can't replace the usage in `LocalChat` and `RemoteChat`.

This leaves `MessageBase` as the only component where `useChat` can be used in it's current implementation.

Using the hook in `LocalChat` and `RemoteChat` will require refactoring them and their usage in `router.tsx`. I'll try this and see if it makes sense to do this.

### DIFF
--- a/src/components/Message/MessageBase.tsx
+++ b/src/components/Message/MessageBase.tsx
@@ -59,15 +59,14 @@ import ImageModal from "../ImageModal";
 import Markdown from "../Markdown";
 
 // Styles for the message text are defined in CSS vs. Chakra-UI
-import { useLiveQuery } from "dexie-react-hooks";
 import useAudioPlayer from "../../hooks/use-audio-player";
 import useMobileBreakpoint from "../../hooks/use-mobile-breakpoint";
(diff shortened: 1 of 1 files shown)


uday-rana:
I think trying to use `useChat` in `LocalChat` and `RemoteChat` would make things more complicated than necessary so I'm going to hold off on it.

mulla028 (approved):
LGTM!

humphd (approved):

Issue #848, opened 2025-03-04
openrouter completely broken in settings


tarasglek:
For repro 
1. use chatcraft.org in private window.
2. open settings

## Click on checkbox next to openrouter

it selects openai

## Click on openrouter api key box

1. It now shows up the magic "get openrouter key" box...We should probably surface that better
2. Click it, login
3. Observe how openrouter key was not added

uday-rana:
I tried to fix this and found that when using `.map()` on `tableProviders`, in `onChange` in `<Checkbox />`, the value of `provider` is always the first entry in `tableProviders`. If you add a new provider, it passes the new provider.

uday-rana:
I think the wrong selection is a JSX bug. I stripped the component down to this and it worked, but re-adding the chakra components broke it again (See below):

```jsx
import {
  Button,
  ButtonGroup,
  Checkbox,
  Flex,
  FormControl,
  FormErrorMessage,
  IconButton,
  Input,
  InputGroup,
  Spinner,
  Table,
  Tbody,
  Td,
  Text,
  Th,
  Thead,
  Tooltip,
  Tr,
} from "@chakra-ui/react";
import { useEffect, useState } from "react";

import { FaCheck } from "react-icons/fa";
import { useAlert } from "../../hooks/use-alert";
import { useSettings } from "../../hooks/use-settings";
import { ChatCraftProvider, ProviderData } from "../../lib/ChatCraftProvider";

import { providerFromUrl, supportedProviders } from "../../lib/providers";
import { CustomProvider } from "../../lib/providers/CustomProvider";
import { FreeModelProvider } from "../../lib/providers/DefaultProvider/FreeModelProvider";
import { OpenRouterProvider } from "../../lib/providers/OpenRouterProvider";
import PasswordInput from "../PasswordInput";
import { MdCancel } from "react-icons/md";

interface ModelsSettingsProps {
  isOpen: boolean;
}

function ModelsSettings(isOpen: ModelsSettingsProps) {
  const { settings, setSettings } = useSettings();

  const { error, success, warning } = useAlert();
  const [isApiKeyInvalid, setIsApiKeyInvalid] = useState(false);
  const [selectedProvider, setSelectedProvider] = useState<ChatCraftProvider | null>(null);
  const [newCustomProvider, setNewCustomProvider] = useState<ChatCraftProvider | null>(null);
  const [apiKeySaved, setApiKeySaved] = useState(false);
[...]

PR #864, opened 2025-03-28
In User Settings > Models > LLM Providers, show "Get OpenRouter Key" button even when OpenRouter is not selected

uday-rana:
Related: #848

Button still doesn't actually get the key but I figured this might be worth landing separately


humphd (approved):
Nice

mulla028 (approved):

PR #866, opened 2025-03-29
Revert `CTRL+l` shortcut to clear chats

### LINKED ISSUE #858: Ctrl+L shortcut causing dataloss
I usually pretty ctrl+L to focus the url bar in browser. However in chatcraft this clears my chat, which causes me to lose value-able chats. Could we use a different shortcut for clearing chats?

Amnish04:
This disables the `CTRL+l` shortcut for the `clear` command on windows so users don't lose their chats accidentally, as it is originally used for focusing on url bar.

I think a proper fix would be the ability for users to toggle shortcuts as per their convenience, and disable them by default. Will file a follow up for further discussion.

OSSpecificShortcut

Fixes #858

### DIFF
--- a/src/lib/commands/ClearCommand.ts
+++ b/src/lib/commands/ClearCommand.ts
@@ -7,6 +7,8 @@ export class ClearCommand extends ChatCraftCommand {
   }
 
   async execute(chat: ChatCraftChat) {
-    return chat.resetMessages();
+    if (ClearCommand.isShortcutEnabled()) {
+      return chat.resetMessages();
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


Amnish04:
Filed #867 as a follow up

src/lib/commands/ClearCommand.ts
```diff
@@ -1,12 +1,23 @@
...
+      keyboardShortcutDescription
+    );
   }
 
   async execute(chat: ChatCraftChat) {
     return chat.resetMessages();
   }
+
+  static isShortcutEnabled() {
+    return !isWindows();
```
tarasglek, line 21: it should be if (isMac()), cos i'm actually on Linux :)

tarasglek (changes requested):

src/lib/commands/ClearCommand.ts
humphd: @Amnish04 I think we should back this out, period.  If CTRL+L is used anywhere to focus the URL bar, it's not a good idea to include it.

Amnish04: @humphd @tarasglek Ok, I've reverted the shortcut changes completely. Please take a look

Issue #868, opened 2025-03-29
RAG on DuckDB Implementation Based on Prototype


mulla028:
## Description

Recently, we have implemented a [prototype](https://github.com/mulla028/duckdb-rag-prototype) of `RAG on DuckDB`, and it proves that implementation is doable for the `ChatCraft` it's time to start working on it!

The implementation will take several steps, lets call them **stages**. Since we already have the set up of DuckDB using `duckdb-wasm`, the file loader, and format to text extractors, we are skipping some of the steps(**stages**). Therefore here are the steps we need to take in order successfully implement it:   

## Proposed Implementation Stages

1. **Stage 1: Documents Retrieving From IndexDB**
   - Implement Logic For the Documents Retrieval from Browser's Local `IndexDB` to the `DuckDB` ( **If such doesn't exist** )
1. **Stage 2: Create Two New Tables in DuckDB** 
   - Embeddings Table
   - Chunk Table
2. **Stage 3: Implement Chunking Logic**
   - Proper Chunking
   - Proper Chunking Storage in DuckDB
3. **Stage 4: Implement Embeddings Generation**
   - Add the OpenAI Model ( `text-embedding-3-small` fits our needs, but **requires API KEY** )
   - Properly Store Generated Embeddings in DuckDB Table.
 4. **Stage 5: Vector Search**
   - Use DuckDB's extension Called **VSS**
   - Apply **HNSW** Indexing to Increase Speed of the Search ( `HNSW` Indexing Provided by VSS extension)
 5. **Stage 6: LLM Integration** 
   - Modify Prompt Construction to Include Retrieved Context
   - Implement Source Attribution in Responses
   - Adjust Token Management to Account For Context
6. **Stage 7: Query Processing**
   - Implement Embedding Generation for **User Queries**
   - Use the Same Embedding Model as Documents for Consistency(`text-embedding-3-small`)

@humphd, @tarasglek please take a look at the proposed implementation stages, and approve them. Let me know if I am missing something :)

humphd:
I made some edits.  Basically, we should plan to store the chunks and embeddings beside the files in IndexedDB, so you can re-use them across chats (if they only live in DuckDB, they will get erased each time, which is wasteful).

We'll also need some way to indicate that we should do RAG vs. loading the entire file into the chat context.  For example, when you attach files, we should probably decide based on the size of the file, and if it's over some threshold, we chunk and do embeddings as part of the import.

mulla028:
[...]

PR #873, opened 2025-04-09
[RAG] Stage - 3: Embeddings Generation

mulla028:
## Description

This is **stage 3** of #868. We are adding the capability of generation and storage vector embeddings for document chunks. It introduces modular embedding provider archtecture, and supports:
* OpenAI's **text-embedding-3-small** API
* Local **tensorflow.js** alternative

[ChatCraftFile](https://github.com/tarasglek/chatcraft.org/blob/mulla028/868-3/src/lib/ChatCraftFile.ts) has been extended with methods to:

- generate embeddings
- store embeddings
- manage embeddings

Integrated embedding generation with the [use-file-import](https://github.com/tarasglek/chatcraft.org/blob/mulla028/868-3/src/hooks/use-file-import.tsx) to **automatically create embeddings after chunking**. 

**New settings added** to control:
- the embedding provider
- batch size
- automatic generation preference

> [!IMPORTANT]
> UI for the embedding preferences required! Therefore, we must land this PR with the updated Settings UI...

## Test It

>[!TIP]
> You may want to test it. In order to do so follow the steps below!

1. Open `CloudFlare` deployment below
2. Upload File **>=300KB**
3. Go to **DevTools**
4. `Application` → `Storage` → `IndexedDB` → `ChatCraftDatabase` → `files`
5. Wuolah! You can see the generated embeddings! (I HOPE :D) 

## Screenshot

image

#### Question 

Maybe we could reduce the size of the minimum chunking size from 300KB → 100 KB. Therefore, the minimum character per chunk is from 1000 → 300.

### DIFF
--- a/src/components/Preferences/RagSettings.tsx
+++ b/src/components/Preferences/RagSettings.tsx
@@ -2,7 +2,6 @@ import {
   Checkbox,
   FormControl,
   FormErrorMessage,
-  FormHelperText,
   FormLabel,
   Radio,
   RadioGroup,
--- a/src/components/Preferences/RagSettings.tsx
+++ b/src/components/Preferences/RagSettings.tsx
@@ -1,0 +1,69 @@
+import {
+  Checkbox,
+  FormControl,
+  FormErrorMessage,
+  FormHelperText,
+  FormLabel,
+  Radio,
+  RadioGroup,
+  Slider,
+  SliderFilledTrack,
+  SliderThumb,
+  SliderTrack,
+  Stack,
+  VStack,
+} from "@chakra-ui/react";
+import { useSettings } from "../../hooks/use-settings";
+import { EmbeddingProviderType } from "../../lib/embeddings";
+
+function RagSettings() {
+  const { settings, setSettings } = useSettings();
+
+  return (
+    <VStack gap={6} my={3}>
+      <FormControl>
+        <FormLabel>RAG - Autogeneration Preference</FormLabel>
+        <Checkbox
+          isChecked={settings.autogenerateEmbeddings}
+          onChange={(e) => setSettings({ ...settings, autogenerateEmbeddings: e.target.checked })}
+        >
+          Autogenerate Chunks and Embeddings
+        </Checkbox>
+      </FormControl>
+      <FormControl>
+        <FormLabel>Embedding Strategy:</FormLabel>
+        <RadioGroup
+          value={settings.embeddingProvider}
+          onChange={(nextValue) =>
+            setSettings({ ...settings, embeddingProvider: nextValue as EmbeddingProviderType })
+          }
+          isDisabled={!settings.autogenerateEmbeddings}
+        >
+          <Stack>
+            <Radio value="openai">Open-AI (Requires API KEY!)</Radio>
+            <Radio value="tensorflow">Tensorflow</Radio>
+          </Stack>
+        </RadioGroup>
+      </FormControl>
+      <FormControl>
+        <FormLabel>Embedding Batch Size {settings.embeddingBatchSize} Chunks</FormLabel>
+        <Slider
+          value={settings.embeddingBatchSize}
+          onChange={(value) => setSettings({ ...settings, embeddingBatchSize: value })}
+          min={20}
+          max={80}
+          step={20}
+          isDisabled={!settings.autogenerateEmbeddings}
+        >
+          <SliderTrack>
+            <SliderFilledTrack />
+          </SliderTrack>
+          <SliderThumb />
+        </Slider>
+        <FormErrorMessage>Maximum batch size must be between 20 and 80 chunks</FormErrorMessage>
+      </FormControl>
+    </VStack>
+  );
+}
+
+export default RagSettings;
--- a/src/components/Preferences/RagSettings.tsx
+++ b/src/components/Preferences/RagSettings.tsx
@@ -33,8 +33,12 @@ function RagSettings() {
         <FormLabel>Embedding Strategy:</FormLabel>
         <RadioGroup
           value={settings.embeddingProvider}
-          onChange={(nextValue) =>
-            setSettings({ ...settings, embeddingProvider: nextValue as EmbeddingProviderType })
+          onChange={(value) =>
+            setSettings({
+              ...settings,
+              embeddingProvider: value as EmbeddingProviderType,
+              embeddingMaxBatchSize: value === "openai" ? 4000 : 256,
+            })
           }
           isDisabled={!settings.autogenerateEmbeddings}
         >
@@ -45,13 +49,13 @@ function RagSettings() {
         </RadioGroup>
       </FormControl>
       <FormControl>
-        <FormLabel>Embedding Batch Size {settings.embeddingBatchSize} Chunks</FormLabel>
+        <FormLabel>Embedding Batch Size {settings.embeddingMaxBatchSize} Chunks</FormLabel>
         <Slider
-          value={settings.embeddingBatchSize}
-          onChange={(value) => setSettings({ ...settings, embeddingBatchSize: value })}
-          min={20}
-          max={80}
-          step={20}
+          value={settings.embeddingMaxBatchSize}
+          onChange={(value) => setSettings({ ...settings, embeddingMaxBatchSize: value })}
+          min={settings.embeddingProvider === "openai" ? 500 : 16}
+          max={settings.embeddingProvider === "openai" ? 4000 : 256}
+          step={settings.embeddingMaxBatchSize > 256 ? 500 : 16}
           isDisabled={!settings.autogenerateEmbeddings}
         >
           <SliderTrack>
--- a/src/components/Preferences/RagSettings.tsx
+++ b/src/components/Preferences/RagSettings.tsx
@@ -13,10 +13,23 @@ import {
   VStack,
 } from "@chakra-ui/react";
 import { useSettings } from "../../hooks/use-settings";
-import { EmbeddingProviderType } from "../../lib/embeddings";
+import { EmbeddingsProviderType, EMBEDDINGS_PROVIDER_CONFIG } from "../../lib/embeddings";
 
 function RagSettings() {
   const { settings, setSettings } = useSettings();
+
+  const currentProviderConfig = EMBEDDINGS_PROVIDER_CONFIG[settings.embeddingsProvider];
+
+  const handleProviderChange = (value: string) => {
+    const providerType = value as EmbeddingsProviderType;
+    const providerConfig = EMBEDDINGS_PROVIDER_CONFIG[providerType];
+
+    setSettings({
+      ...settings,
+      embeddingsProvider: providerType,
+      embeddingsBatchSize: providerConfig.defaultBatchSize,
+    });
+  };
 
   return (
     <VStack gap={6} my={3}>
@@ -32,30 +45,24 @@ function RagSettings() {
       <FormControl>
         <FormLabel>Embedding Strategy:</FormLabel>
         <RadioGroup
-          value={settings.embeddingProvider}
-          onChange={(value) =>
-            setSettings({
-              ...settings,
-              embeddingProvider: value as EmbeddingProviderType,
-              embeddingMaxBatchSize: value === "openai" ? 4000 : 256,
-            })
-          }
+          value={settings.embeddingsProvider}
+          onChange={handleProviderChange}
           isDisabled={!settings.autogenerateEmbeddings}
         >
           <Stack>
-            <Radio value="openai">Open-AI (Requires API KEY!)</Radio>
-            <Radio value="tensorflow">Tensorflow</Radio>
+            <Radio value="openai">Use OpenAI (Requires API KEY!)</Radio>
+            <Radio value="tensorflow">Generate In Browser (tensorflow.js)</Radio>
           </Stack>
         </RadioGroup>
       </FormControl>
       <FormControl>
-        <FormLabel>Embedding Batch Size {settings.embeddingMaxBatchSize} Chunks</FormLabel>
+        <FormLabel>Embedding Batch Size {settings.embeddingsBatchSize} Chunks</FormLabel>
         <Slider
-          value={settings.embeddingMaxBatchSize}
-          onChange={(value) => setSettings({ ...settings, embeddingMaxBatchSize: value })}
-          min={settings.embeddingProvider === "openai" ? 500 : 16}
-          max={settings.embeddingProvider === "openai" ? 4000 : 256}
-          step={settings.embeddingMaxBatchSize > 256 ? 500 : 16}
+          value={settings.embeddingsBatchSize}
+          onChange={(value) => setSettings({ ...settings, embeddingsBatchSize: value })}
+          min={currentProviderConfig.minBatchSize}
+          max={currentProviderConfig.maxBatchSize}
+          step={currentProviderConfig.minBatchSize}
           isDisabled={!settings.autogenerateEmbeddings}
         >
           <SliderTrack>
@@ -63,7 +70,10 @@ function RagSettings() {
           </SliderTrack>
           <SliderThumb />
         </Slider>
-        <FormErrorMessage>Maximum batch size must be between 20 and 80 chunks</FormErrorMessage>
+        <FormErrorMessage>
+          Maximum batch size must be between {currentProviderConfig.minBatchSize} and{" "}
+          {currentProviderConfig.maxBatchSize} chunks
+        </FormErrorMessage>
       </FormControl>
     </VStack>
   );
--- a/src/hooks/use-file-import.tsx
+++ b/src/hooks/use-file-import.tsx
@@ -217,14 +217,14 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
       await chat.addFile(chatCraftFile);
 
       // Generate chunks for the file if its size > 300
-      if (!chatCraftFile.hasChunks()) {
+      if (!chatCraftFile.hasChunks() && settings.autogenerateEmbeddings) {
         try {
           await chatCraftFile.generateChunks();
           console.log(
             `Generated ${chatCraftFile.chunks?.length || 0} chunks for file: ${chatCraftFile.name}`
           );
 
-          if (settings.autogenerateEmbeddings && chatCraftFile.hasChunks()) {
+          if (!chatCraftFile.hasEmbeddings()) {
             try {
               await chatCraftFile.generateEmbeddings();
             } catch (err) {
--- a/src/hooks/use-file-import.tsx
+++ b/src/hooks/use-file-import.tsx
@@ -216,7 +216,8 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
       const chatCraftFile = await ChatCraftFile.findOrCreate(file, { text });
       await chat.addFile(chatCraftFile);
 
-      // Generate chunks for the file if its size > 300
+      // generates chunks and embeddings only when user decided to turn on
+      // experimental feature
       if (!chatCraftFile.hasChunks() && settings.autogenerateEmbeddings) {
         try {
           await chatCraftFile.generateChunks();
--- a/src/hooks/use-file-import.tsx
+++ b/src/hooks/use-file-import.tsx
@@ -223,6 +223,16 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
           console.log(
             `Generated ${chatCraftFile.chunks?.length || 0} chunks for file: ${chatCraftFile.name}`
           );
+
+          if (settings.autogenerateEmbeddings && chatCraftFile.hasChunks()) {
+            try {
+              await chatCraftFile.generateEmbeddings();
+            } catch (err) {
+              throw new Error(
+                `Error generating embeddings: ${err}, continue without embeddings...`
+              );
+            }
+          }
         } catch (err) {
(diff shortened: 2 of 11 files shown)


mulla028:
Right now I am implementing the UI for RAG settings, and I am trying to figure out what batch size limit for the `TensorFlow`. According to some research, it depends on some kind of browser **WebGL**, I don't really know what's this. It says `50-100` chunks is the potential maximum, but `20-30` is a safer option. Just wanna know which values add to the slider. @humphd, maybe you know the answer.

mulla028:
Since now, by default we do not generate embeddings

mulla028:
By the way, using the openai option generates embeddings within seconds for those files that we were trying to upload within the lecture

mulla028:
By increasing the tensorflow batch size from 20 to 40 chunks, it takes 5 seconds to generate embeddings for 11.3MB PDF file

src/lib/embeddings/OpenAIEmbedding.ts
```diff
@@ -31,13 +30,27 @@ export class OpenAIEmbeddingProvider implements EmbeddingProvider {
    */
   async generateBatchEmbeddings(texts: string[]): Promise<number[][]> {
     try {
-      const response = await this.openai.embeddings.create({
-        model: "text-embedding-3-small",
-        input: texts,
-        encoding_format: "float",
+      const settings = getSettings();
+      console.log(`Current API KEY is: ${settings.currentProvider.apiKey}`);
```
humphd: Don't do this

src/lib/embeddings/OpenAIEmbedding.ts
```diff
@@ -31,13 +30,27 @@ export class OpenAIEmbeddingProvider implements EmbeddingProvider {
...
+      const response = await fetch(`${this.url}`, {
```
humphd: Why can't use use the OpenAI client here?

mulla028:
I may confirm that now embeddings are properly generated for both strategies: `tensorflow` and `openai`. All settings work as expected, but I had to change the way I use the openai method since the client version wasn't allowing it for the reason that it has an ERROR: **Seams like you use browser-like etc... API KEY MAY BE REVEALED**, and I had to use fetch. The last thing is that I am going to try to use `Worker` for the parallel threading and make a progress bar update while generating both chunks and embeddings. After we finish landing the RAG I will have a follow-up for  UI improvement during **file import**.

mulla028:
Do I remember correctly that we need to make huge files, for example, >5MB to be generated using openai only, or to make an alert for the user that he could use Openai, so he isn't forced and still has a choice

humphd:
You have to set a flag to do it in a browser:

>Why is this dangerous?
>Enabling the dangerouslyAllowBrowser option can be dangerous because it exposes your secret API credentials in the client-side code. Web browsers are inherently less secure than server environments, any user with access to the browser can potentially inspect, extract, and misuse these credentials. This could lead to unauthorized access using your credentials and potentially compromise sensitive data or functionality.
>
>When might this not be dangerous?
>In certain scenarios where enabling browser support might not pose significant risks:
>
>Internal Tools: If the application is used solely within a controlled internal environment where the users are trusted, the risk of credential exposure can be mitigated.
Public APIs with Limited Scope: If your API has very limited scope and the exposed credentials do not grant access to sensitive data or critical operations, the potential impact of exposure is reduced.
>Development or debugging purpose: Enabling this feature temporarily might be acceptable, provided the credentials are short-lived, aren't also used in production environments, or are frequently rotated.

mulla028:
You wanna see the following changes: 
* revert to openai usage vs. fetch
* add that flag ?

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </Stack>
+        </RadioGroup>
+      </FormControl>
+      <FormControl>
+        <FormLabel>Embedding Batch Size {settings.embeddingBatchSize} Chunks</FormLabel>
+        <Slider
+          value={settings.embeddingBatchSize}
+          onChange={(value) => setSettings({ ...settings, embeddingBatchSize: value })}
+          min={20}
+          max={80}
```
humphd: Why 80?  OpenAI can do way more.  Is Tensorflow.js limited?

src/hooks/use-file-import.tsx
```diff
@@ -217,12 +217,22 @@ export function useFileImport({ chat, onImageImport }: UseFileImportOptions) {
       await chat.addFile(chatCraftFile);
 
       // Generate chunks for the file if its size > 300
```
humphd: This comment doesn't totally make sense with the code that follows.  You're not looking at the file size here, only whether there are already chunks and what the preferences say.

src/lib/ChatCraftFile.ts
```diff
@@ -180,6 +182,28 @@ export class ChatCraftFile {
...
+
+  /**
+   * Get embedding dimensions
+   */
+  getEmbeddingDimensions(): number | undefined {
+    if (!this.hasEmbeddings() || !this.chunks?.[0]?.metadata) {
+      return undefined;
+    }
+
+    return this.chunks[0].metadata.embeddingDimensions as number;
```
humphd: The fact that here and on line 193 you have to get the first chunk in order to get the metadata you need makes me think that we should lift this info out of the chunks and store it on its own object.

src/lib/ChatCraftFile.ts
```diff
@@ -205,6 +229,64 @@ export class ChatCraftFile {
     return !!this.chunks && this.chunks.length > 0;
   }
 
+  /**
+   * Check if all chunks have embeddings
+   */
+  hasEmbeddings(): boolean {
+    return (
+      this.hasChunks() &&
+      this.chunks!.every((chunk) => Array.isArray(chunk.embeddings) && chunk.embeddings.length > 0)
```
humphd: Don't use `!` in TypeScript like this, you're turning off the type checking.

src/lib/ChatCraftFile.ts
```diff
@@ -205,6 +229,64 @@ export class ChatCraftFile {
...
+
+    const settings = getSettings();
+    const provider = getEmbeddingProvider(providerType || settings.embeddingProvider);
+    const chunks = this.chunks!;
+
+    console.log(`Generating embeddings for ${chunks.length} chunks using ${provider.name}...`);
+
+    const BATCH_SIZE = settings.embeddingBatchSize || 20;
+
+    for (let i = 0; i < chunks.length; i += BATCH_SIZE) {
```
humphd, line 264: This code can move to a worker thread in a follow-up.

src/lib/embeddings/TensorflowEmbedding.ts
```diff
@@ -0,0 +1,78 @@
...
+  private isLoading: boolean = false;
+  private loadPromise: Promise<void> | null = null;
+
+  constructor() {}
+
+  /**
+   * Load the model if it hasn't been loaded yet
+   */
+  private async loadModelIfNeeded(): Promise<void> {
+    if (this.model) return;
```
humphd: Always brace your if

src/lib/embeddings/TensorflowEmbedding.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!this.loadPromise) {
+      this.isLoading = true;
+
+      this.loadPromise = (async () => {
+        try {
+          console.log("Loading Universal Sentence Encoder model...");
+
+          await import("@tensorflow/tfjs");
```
humphd, line 46: Question: will this load the best backend (e.g., GPU or WebGL versions) by default?

src/lib/embeddings/TensorflowEmbedding.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  async generateBatchEmbeddings(texts: string[]): Promise<number[][]> {
+    try {
+      await this.loadModelIfNeeded();
+
+      const embeddings = await this.model.embed(texts);
+
+      const arrays = await embeddings.array();
+
+      embeddings.dispose();
```
humphd: This needs to go in a `finally` probably, so we don't leak

humphd (changes requested):
This is looking really good.  Some issues to sort out now, and some more for follow-ups.

I love that you also added some UI!

src/hooks/use-file-import.tsx
mulla028: Forgot to delete it. I am cleaning it up

src/lib/embeddings/TensorflowEmbedding.ts
mulla028: I wouldn't say it loads the best by default, but loads the best **available** by default

mulla028: TensorFlow has a setBackend() method, where I could specify one. Do you have a preference?

humphd: We can't know what the user will have (e.g., I might have a GPU, another user doesn't).  The goal would be that we load the most capable backend at runtime.

This can happen in a follow-up, don't let it block things here.  I was just curious.

src/lib/ChatCraftFile.ts
mulla028: Can we store it inside `this.metadata` instead of `this.chunks[0].metadata`?

humphd: That makes sense to me.

src/lib/embeddings/OpenAIEmbedding.ts
mulla028: Is it safe to use fetch here? I used it instead of the OpenAI client because it helped me avoid this error:

> You have to set a flag to do it in a browser:
>
>Why is this dangerous?
Enabling the dangerouslyAllowBrowser option can be dangerous because it exposes your secret API credentials in the client-side code. Web browsers are inherently less secure than server environments, any user with access to the browser can potentially inspect, extract, and misuse these credentials. This could lead to unauthorized access using your credentials and potentially compromise sensitive data or functionality.
>
>When might this not be dangerous?
In certain scenarios where enabling browser support might not pose significant risks:
>
>Internal Tools: If the application is used solely within a controlled internal environment where the users are trusted, the risk of credential exposure can be mitigated.
Public APIs with Limited Scope: If your API has very limited scope and the exposed credentials do not grant access to sensitive data or critical operations, the potential impact of exposure is reduced.
Development or debugging purpose: Enabling this feature temporarily might be acceptable, provided the credentials are short-lived, aren't also used in production environments, or are frequently rotated.

humphd: Using the OpenAI client is better--it's a wrapper around `fetch()` with extras implemented for you.

mulla028:
The `setMetadata()` method wasn't working properly tho. It wasn't updating the db for some reason.

src/lib/embeddings/OpenAIEmbedding.ts
mulla028: Oh, that makes sense. I will revert it to the OpenAI client usage but gotta find a way to avoid that error.

src/components/Preferences/RagSettings.tsx
mulla028: Tbh, I don't really understand their limits, but now I am testing different sizes for both providers. If you know, could you explain to me how I learn about the TensorFlow limits since I don't use the tokens and every user will have their own backend used(if there is a way to learn about it without the testing)?

mulla028: Even 1000 chunks work and it's rapid 😆

src/lib/embeddings/OpenAIEmbedding.ts
mulla028: Alright, I reverted it and added a flag to the OpenAI client `dangerouslyAllowBrowser: true`, hope this is a way, and I hope chatcraft is safe enough to use that flag.

src/components/Preferences/RagSettings.tsx
humphd: OpenAI can use thousands, and tensorflow.js will be limited by available RAM.  So I think the range we provide here needs to be built-into the embedding provider, since the two differ so much.

src/lib/embeddings/OpenAIEmbedding.ts
humphd: It's not that it's dangerous as such.  It's more that they want you to understand that using an API key in browsers is obviously dangerous in general.  They want you to think through what you're doing and opt into using it.

But it's the same as you passing it to the headers of the `fetch()` call.

humphd:
It looks like tensorflow.js will automatically try to load the best backend for your browser/hardware.

mulla028:
`await import("@tensorflow/tfjs");` it automatically loads the best backend available for your browser/hardware

src/lib/embeddings/OpenAIEmbedding.ts
mulla028: Thank you for the explanation! It makes sense now

src/components/Preferences/RagSettings.tsx
mulla028: Do we still want to provide the choice of batch size in the settings? Or just try to set the default value and max value based on the provider?

humphd: I think we should, but when using tensorflow.js, we'd use 16, 32, 64, etc. and for OpenAI we can do a lot more.

mulla028: Let's say that the limit for tensorflow.js is 256, I tried a huge pdf that generated 250 chunks and it worked well with a batch size of 220 ish, but the default value is going to be 128. For OpenAI I am going to set the default value as 2500  and the max as 4000.

humphd: I think 4000 is low, but let's try and adjust in follow-ups.

mulla028:
Tbh, I don't like my UI solution for the batch size adjustment logic it looks ugly, and I believe there should be a smarter way.

mulla028:
Despite the UI solution, I think this draft PR is ready to become at least an actual PR. Do we need to switch providers to OpenAI if huge files are uploaded? Or we could focus on the main core of landing RAG, and then implement it as a follow-up leaving it as an experimental feature

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <Radio value="tensorflow">Tensorflow</Radio>
+          </Stack>
+        </RadioGroup>
+      </FormControl>
+      <FormControl>
+        <FormLabel>Embedding Batch Size {settings.embeddingMaxBatchSize} Chunks</FormLabel>
+        <Slider
+          value={settings.embeddingMaxBatchSize}
+          onChange={(value) => setSettings({ ...settings, embeddingMaxBatchSize: value })}
+          min={settings.embeddingProvider === "openai" ? 500 : 16}
```
humphd: Move these values into the providers as `minChunkBatchSize` and `maxChunkBatchSize` properties

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </FormControl>
+      <FormControl>
+        <FormLabel>Embedding Strategy:</FormLabel>
+        <RadioGroup
+          value={settings.embeddingProvider}
+          onChange={(value) =>
+            setSettings({
+              ...settings,
+              embeddingProvider: value as EmbeddingProviderType,
+              embeddingMaxBatchSize: value === "openai" ? 4000 : 256,
```
humphd: Move this into the provider as `defaultChunkBatchSize`

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+            })
+          }
+          isDisabled={!settings.autogenerateEmbeddings}
+        >
+          <Stack>
+            <Radio value="openai">Open-AI (Requires API KEY!)</Radio>
+            <Radio value="tensorflow">Tensorflow</Radio>
```
humphd: "Generate in browser (tensorflow.js)"

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+            setSettings({
+              ...settings,
+              embeddingProvider: value as EmbeddingProviderType,
+              embeddingMaxBatchSize: value === "openai" ? 4000 : 256,
+            })
+          }
+          isDisabled={!settings.autogenerateEmbeddings}
+        >
+          <Stack>
+            <Radio value="openai">Open-AI (Requires API KEY!)</Radio>
```
humphd: "Use OpenAI (requires API Key)"

src/components/Preferences/RagSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+          max={settings.embeddingProvider === "openai" ? 4000 : 256}
+          step={settings.embeddingMaxBatchSize > 256 ? 500 : 16}
+          isDisabled={!settings.autogenerateEmbeddings}
+        >
+          <SliderTrack>
+            <SliderFilledTrack />
+          </SliderTrack>
+          <SliderThumb />
+        </Slider>
+        <FormErrorMessage>Maximum batch size must be between 20 and 80 chunks</FormErrorMessage>
```
humphd: This needs to match what the provider supports

src/lib/ChatCraftFile.ts
```diff
@@ -93,6 +95,7 @@ export class ChatCraftFile {
     input: File | ChatCraftFileOptions,
     additionalOptions: Partial<ChatCraftFileOptions>
   ): Promise<ChatCraftFile> {
+    console.log("findOrCreate invoked");
```
humphd: Remove this.

src/lib/ChatCraftFile.ts
```diff
@@ -105,6 +108,8 @@ export class ChatCraftFile {
         },
       };
 
+      console.log("Metadata that is getting used", options.metadata);
+      console.log("Additional Metadata: ", additionalOptions);
```
humphd: Remove these

src/lib/ChatCraftFile.ts
```diff
@@ -180,14 +185,37 @@ export class ChatCraftFile {
     return this.chunks?.[index];
   }
 
+  /**
+   * Get the embedding provider used for this file's chunks
+   */
+  getEmbeddingProviderId(): string | undefined {
```
humphd: Why not `getEmbeddingsProvider()` and return an object that has all the info for the provider?

src/lib/ChatCraftFile.ts
```diff
@@ -180,14 +185,37 @@ export class ChatCraftFile {
...
+      return undefined;
+    }
+
+    return this.metadata.embeddingProvider as string;
+  }
+
+  /**
+   * Get embedding dimensions
+   */
+  getEmbeddingDimensions(): number | undefined {
```
humphd: Use "Embeddings" (plural) everywhere

src/lib/ChatCraftFile.ts
```diff
@@ -205,6 +233,60 @@ export class ChatCraftFile {
...
 
+  /**
+   * Check if all chunks have embeddings
+   */
+  hasEmbeddings(): boolean | undefined {
+    if (!this.hasChunks()) {
+      return false;
+    }
+
+    return this.chunks?.some((chunk) => chunk.embeddings.length > 0);
```
humphd: Use `.every()`

src/lib/embeddings/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        break;
+
+      default:
+        throw new Error(`Unknown embedding provider type: ${type}`);
+    }
+
+    settings.embeddingMaxBatchSize = providers[type].maxBatchSize;
+  }
+
+  return providers[type]!;
```
humphd: Don't use `!`

humphd (changes requested):

src/lib/ChatCraftFile.ts
mulla028: I agree

mulla028:
Alright, now we have a single embeddings metadata getter in `ChatCraftFile.ts`

`src/lib/embeddings`:
* Added `minBatchSize`, `maxBatchSize` and `defaultBatchSize`
* Created configs
 
 `src/lib/settings.ts`:
 * renamed properties
 
 Based on the change request made a clean up.
 
 I don't know if you like this config approach, I just couldn't come up with the better one. Let me know if I am missing something or if you think could be implemented in another way

mulla028:
I know that we are all taking a break after this hard term, and I hope that we are still implementing this feature. However, I am inactive till the May, 3. Still waiting for the feedback, let me know when I have to expect it (without pushing anyone, probably you are also taking a break now.) Just let me know when to expect the feedback, since I haven't received anything yet, according to the latest updates. Have a good break :)

mulla028:
@humphd Could you review this pull-request, or do we have to figure [it](https://github.com/tarasglek/chatcraft.org/issues/874) out first?

humphd:
@mulla028 it's on my list, and getting this stuff fully integrated is something I want to do.  But I'm currently tied up with other projects.  TODO...

tarasglek:
I'm concerned about UX after this change. Now it's completely not clear if file is gonna get RAGed or wholly passed to LLM.

mulla028:
> I'm concerned about UX after this change. Now it's completely not clear if file is gonna get RAGed or wholly passed to LLM.

We have already discussed it with @humphd. We came up with the idea of experimental feature until we land all the stages. Now in the settings you may pick whether you want embedding to be generated(for the further stages) or not. By default its off

mulla028:
image

